PR #29, opened 2019-12-08
Generator voltage remote control support 

geofjamg:

No

Feature

There is no generator voltage remote control support.

Generator voltage remote control  is now supported. CoordinatedReactiveControl extension is used for reactive power calculation.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
@@ -27,6 +27,8 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
 
     private boolean dc = false;
 
+    private boolean generatorVoltageRemoteControl = false;
+
     private final List<AcLoadFlowObserver> additionalObservers = new ArrayList<>();
 
     @Override
@@ -58,6 +60,15 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
 
     public OpenLoadFlowParameters setDc(boolean dc) {
         this.dc = dc;
+        return this;
+    }
+
+    public boolean hasGeneratorVoltageRemoteControl() {
+        return generatorVoltageRemoteControl;
+    }
+
+    public OpenLoadFlowParameters setGeneratorVoltageRemoteControl(boolean generatorVoltageRemoteControl) {
+        this.generatorVoltageRemoteControl = generatorVoltageRemoteControl;
         return this;
     }
 
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
@@ -69,7 +69,7 @@ public final class AcEquationSystem {
         }
 
         // create voltage equation at remote control target bus
-        equationSystem.createEquation(bus.getNum(), EquationType.BUS_REMOTE_V).addTerm(new BusVoltageEquationTerm(bus, variableSet));
+        equationSystem.createEquation(bus.getNum(), EquationType.BUS_V).addTerm(new BusVoltageEquationTerm(bus, variableSet));
 
         List<Double> qKeys = createReactiveKeys(sourceBuses);
 
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
@@ -9,9 +9,10 @@ package com.powsybl.openloadflow.ac.equations;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.openloadflow.equations.*;
 import com.powsybl.openloadflow.network.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
@@ -20,6 +21,8 @@ import java.util.stream.Collectors;
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 public final class AcEquationSystem {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AcEquationSystem.class);
 
     private AcEquationSystem() {
     }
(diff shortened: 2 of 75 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfNetworks.java
```diff
@@ -73,11 +74,16 @@ public void visitThreeWindingsTransformer(ThreeWindingsTransformer transformer,
...
                 public void visitGenerator(Generator generator) {
-                    if (!Objects.equals(generator.getRegulatingTerminal().getBusView().getBus().getId(),
-                            generator.getTerminal().getBusView().getBus().getId())) {
-                        throw new PowsyblException("Generator remote voltage control is not yet supported");
-                    }
                     lfBus.addGenerator(generator);
+
+                    String controlBusId = generator.getRegulatingTerminal().getBusView().getBus().getId();
+                    String connectedBusId = generator.getTerminal().getBusView().getBus().getId();
+                    if (!Objects.equals(controlBusId, connectedBusId)) {
```
annetill: Are we sure that we want to do that even if the voltage regulation is disabled ? In my use case, if I remove the remote voltage control from generators, there is no calculation.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -47,6 +57,67 @@ private static void createBusEquations(LfNetwork network, VariableSet variableSe
...
+        List<Double> qKeys = createReactiveKeys(sourceBuses);
+
+        for (LfBus sourceBus : sourceBuses) {
+            // deactivate reactive equation for all remote voltage source bus (where the generator is connected)
+            equationSystem.createEquation(sourceBus.getNum(), EquationType.BUS_Q).setActive(false);
+        }
+
+        // create (number of source bus) - 1 equation to link the reactive equations together
+        LfBus firstSourceBus = sourceBuses.get(0);
+        for (int i = 1; i < sourceBuses.size(); i++) {
```
annetill: I don't really understand this loop.

src/main/java/com/powsybl/openloadflow/equations/Equation.java
```diff
@@ -100,10 +101,23 @@ void initTarget(LfNetwork network, double[] targets) {
                 targets[row] = network.getBus(num).getTargetV();
                 break;
 
+            case BUS_REMOTE_V:
+                targets[row] = network.getBus(num).getRemoteControlSourceBuses()
+                        .stream()
+                        .filter(LfBus::hasVoltageControl)
+                        .findFirst()
+                        .orElseThrow(() -> new IllegalStateException("None of the remote control source buses have voltage control on"))
```
annetill: Is it a real issue that all the generators with remote control are disabled ? I have this case on Litgrid network.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -61,6 +63,15 @@ public OpenLoadFlowParameters setDc(boolean dc) {
         return this;
     }
 
+    public boolean hasGeneratorVoltageRemoteControl() {
+        return generatorVoltageRemoteControl;
+    }
+
+    public OpenLoadFlowParameters setGeneratorVoltageRemoteControl(boolean generatorVoltageRemoteControl) {
```
annetill: Can we changed this parameter from the config.yml file ? It is required for me if I want to use the same tool (build with ittols packager) to validate LF results from different LF engines. Can we create a section ````openloadflow-default-parameters````. Or maybe this parameter should be located in the common loadflow API.

geofjamg: Yes, this parameter should be in common loadflow parameters as many other one (distributed slack, etc).
We should open an issue and list parameters (from open load flow and hades2) that could be shared  between all implementations.
Persistence of default values already exists for common loadflow parameters so we could do it here once the generatorVoltageRemoteControl have been moved.

src/main/java/com/powsybl/openloadflow/equations/Equation.java
geofjamg: This (normal) case has already been handled in network read part so at this stage we should not go into this part of the code.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -47,6 +57,67 @@ private static void createBusEquations(LfNetwork network, VariableSet variableSe
         }
     }
 
+    private static void createLocalVoltageEquation(VariableSet variableSet, EquationSystem equationSystem, LfBus bus) {
+        equationSystem.createEquation(bus.getNum(), EquationType.BUS_V).addTerm(new BusVoltageEquationTerm(bus, variableSet));
+        equationSystem.createEquation(bus.getNum(), EquationType.BUS_Q).setActive(false);
+    }
+
+    private static void createRemoteVoltageEquations(LfBus bus, List<LfBus> sourceBuses, EquationSystem equationSystem, VariableSet variableSet) {
```
annetill: This equation system for remote control does not work on my use cases, from the first NR iteration. It meens that the issue is in the equation system, not in the reactive limits outer loop. I have tried to make something simpler: in my code test, I do as if the generator with remote control is directly connected at the remote bus and the source buses just have the reactive linked equations. 
````
equationSystem.createEquation(bus.getNum(), EquationType.BUS_REMOTE_V).addTerm(new BusVoltageEquationTerm(bus, variableSet)); // New PV bus.
equationSystem.createEquation(bus.getNum(), EquationType.BUS_Q).setActive(false); // Not a PQ bus anymore.

….

for (LfBus sourceBus : sourceBuses) {
            // deactivate reactive equation for all remote voltage source bus (where the generator is connected)
            // equationSystem.createEquation(sourceBus.getNum(), EquationType.BUS_Q).setActive(false);
           // Source buses are in fact PQ buses., but the sum of reactive power of source buses should be linked.  
        }
````

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
annetill: I don't have found any real use case to test this part…

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
annetill: But maybe it is not the equation system you have in mind…

annetill:
I have tested this feature on networks without voltage remote control, on networks with shared or not shared voltage remote control. The voltage plan is close to those we can calculate with Hades2 (https://github.com/rte-france/hades2-distribution). We can merge this feature or wait a little bit for the treatement of reactive limits in case of a shared control.

geofjamg:
> I have tested this feature on networks without voltage remote control, on networks with shared or not shared voltage remote control. The voltage plan is close to those we can calculate with Hades2 (https://github.com/rte-france/hades2-distribution). We can merge this feature or wait a little bit for the treatement of reactive limits in case of a shared control.

We should merge it as there is already a lot of code and uncovered case is rare and cleanly handled by throwing an exception.

PR #36, opened 2019-12-30
SVC reactive limits

geofjamg:

No

Bug fix

Static var compensator reactive limits are not taken into account.

We use bmin and bmax to compute reactive limits at nominal voltage.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/AbstractLfGenerator.java
+++ b/src/main/java/com/powsybl/openloadflow/network/AbstractLfGenerator.java
@@ -66,7 +66,6 @@ public abstract class AbstractLfGenerator implements LfGenerator {
                             maxRangeQ = Math.max(maxRangeQ, point.getMaxQ() - point.getMinQ());
                         }
                     }
-                    maxRangeQ = maxRangeQ / PerUnit.SB;
                     break;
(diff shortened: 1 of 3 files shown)


src/main/java/com/powsybl/openloadflow/network/AbstractLfGenerator.java
```diff
@@ -65,6 +65,7 @@ public double getMaxRangeQ() {
                         } else {
                             maxRangeQ = Math.max(maxRangeQ, point.getMaxQ() - point.getMinQ());
                         }
+                        maxRangeQ = maxRangeQ / PerUnit.SB;
```
annetill: In fact, it seems that reactive capability curves are also already in per unit. What do you think ?

geofjamg:
> 
> 
> In `AbstractLfGenerator`, the reactive range could be computed from a reactive capability curve or from the min and the max values of a generator (and that could be a static var compensator). It seems that the reactive capability curve is not in per unit whereas the min and max values are. Maybe the good solution is to per unit the reactive capability curve directly.

I changed a little bit your fix to process same way per uniting of reactive limits in both min/max and curve cases.

PR #38, opened 2020-01-04
Implement shared voltage remote control with reactive limit case

geofjamg:

- [] Tests for the changes have been added (for bug fixes / features)

No

Bug fix

Reactive limit PV/PQ switching is not implemented in case of shared voltage remote control.

It is now implemented.

**Other information**:


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -69,26 +67,46 @@ private static void createLocalVoltageEquation(VariableSet variableSet, Equation
         equationSystem.createEquation(bus.getNum(), EquationType.BUS_Q).setActive(false);
     }
 
-    private static void createRemoteVoltageEquations(LfBus bus, List<LfBus> sourceBuses, EquationSystem equationSystem, VariableSet variableSet) {
-        // check voltage target consistency
-        if (sourceBuses.stream().mapToDouble(LfBus::getTargetV).distinct().count() != 1) {
-            throw new PowsyblException("Inconsistent target voltage at bus " + bus.getId());
+    public static void createRemoteVoltageEquations(LfBus bus, EquationSystem equationSystem, VariableSet variableSet) {
+        for (LfBus sourceBus : bus.getRemoteControlSourceBuses()) {
+            // deactivate reactive equation for all remote voltage source bus (where the generator is connected) that do
```
annetill: I think the comment should be:
- In case of a voltage remote control active, the reactive equation is disabled at each source bus. 
- In case of a bus (with remote voltage control) switching from PV to PQ, the reactive equation is enabled.

geofjamg: Ok I replaced with your comment

Issue #42, opened 2020-01-23
Allow remote regulation for static var compensators


annetill:
* **Do you want to request a *feature* or report a *bug*?**

This new feature will be probably integrated in the next release of powsybl-core. See PR:
https://github.com/powsybl/powsybl-core/pull/1124

* **What is the current behavior?**

Static var compensators can control voltage locally.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

Static var compensators could have a voltage remote control as generators. An other option could be to rescale the remote voltage target value to a local one. For generators, we have both options.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #44, opened 2020-01-28
SVC voltage remote control

geofjamg:

#42 

Feature

Static var compensator voltage remote control is not supported.

Static var compensator voltage remote control is now supported, in a same way than generators.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -70,26 +70,30 @@ public void visitThreeWindingsTransformer(ThreeWindingsTransformer transformer,
...
-                            double previousTargetV = generator.getTargetV();
-                            double remoteNominalV = generator.getRegulatingTerminal().getVoltageLevel().getNominalV();
-                            double localNominalV = generator.getTerminal().getVoltageLevel().getNominalV();
+                            double remoteNominalV = regulatingTerminal.getVoltageLevel().getNominalV();
+                            double localNominalV = injection.getTerminal().getVoltageLevel().getNominalV();
                             scaleV = localNominalV / remoteNominalV;
-                            LOGGER.warn("Generator remote voltage control is not yet supported. The voltage target of " +
-                                            "generator {} with remote control is rescaled from {} to {}",
-                                    generator.getId(), previousTargetV, previousTargetV * scaleV);
+                            LOGGER.warn("Remote voltage control is not yet supported. The voltage target of " +
```
annetill: Maybe change the comment just to say that remote voltage control is not active, so targets will be rescaled.

geofjamg:
> I understand that we will use the parameter `voltageRemoteControl` to activate voltage remote control for generators and SVCs, in the same time. I don't know if it could be useful to have 2 parameters instead of just one.

Ok we can keep the 2 parameters.

geofjamg:
> An other question, about the name chosen : `voltageRemoteControl` is not so precise and we will have to change it when we will model voltage remote regulation for shunt and tap changers. So, we keep like that and we change later or it is better to decide now ?

As this  PR will only be merge when core 3.2.0 will be released, we can prepare in another PR all needed options and throws an exception if not yet supported. It would be better to add all this parameters to LoadFlowParameters and not to OLF specific parameter. That would be a good idea to do it for core 3.2.0.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
geofjamg: Yes this warning has  been added while not yet implemented but now it has to be fixed "yet supported" => "active"

PR #48, opened 2020-02-13
A generator with maxP equal to Infinity is discarded from active power control.

annetill:

No.

A bug fix. I have just added a test to discard from the active power control generators which Pmax are close to Infinity. This test seams necessary for computations that comes from a CGMES conversion.  

The compensation fails because factors are equal to zero when they are rescaled by the sum of factors (=Infinity).

The sum is done without any Infinity value.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -59,6 +59,11 @@ private LfGeneratorImpl(Generator generator, LfNetworkLoadingReport report) {
             report.generatorsDiscardedFromActivePowerControlBecauseTargetPGreaterThenMaxP++;
             participating = false;
         }
+        if (generator.getMaxP() > Double.MAX_VALUE) {
```
geofjamg: Double.isInfinite ?

PR #56, opened 2020-03-17
Slack distribution on loads

annetill:

**Other information**:


annetill:
Hi Geoffory, it is just a first import and tests are missing. But I really don't know if you will be happy with the design and these first implementation. So feal free to do a quick review to help me !

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -25,6 +25,14 @@
 
     private boolean distributedSlack = true;
 
+    private BalanceType balanceType = BalanceType.PROPORTIONAL_TO_GENERATION_P; // Default slack distribution on generators.
+
+    public enum BalanceType {
+        PROPORTIONAL_TO_GENERATION_P,
+        // PROPORTIONAL_TO_GENERATION_P_MAX, Not implemented yet.
```
geofjamg: This is the one that has been implemented. PROPORTIONAL_TO_GENERATION_P has not been done.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -119,12 +116,17 @@ private OpenLoadFlowParameters getParametersExt(LoadFlowParameters parameters) {
...
             LOGGER.info("Voltage remote control: {}", parametersExt.hasVoltageRemoteControl());
 
             List<OuterLoop> outerLoops = new ArrayList<>();
             if (parametersExt.isDistributedSlack()) {
-                outerLoops.add(new DistributedSlackOuterLoop());
+                if (parametersExt.getBalanceType() == OpenLoadFlowParameters.BalanceType.PROPORTIONAL_TO_GENERATION_P) {
+                    outerLoops.add(new DistributedGenerationSlackOuterLoop());
+                } else if (parametersExt.getBalanceType() == OpenLoadFlowParameters.BalanceType.PROPORTIONAL_TO_LOAD) {
+                    outerLoops.add(new DistributedLoadSlackOuterLoop());
+                }
```
geofjamg: Throws an exception in case of not yet supported mode.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+
+        ParticipatingLoad(LfBus load, double factor) {
+            this.load = load;
+            this.factor = factor;
+        }
+    }
+
+    @Override
+    public String getName() {
+        return "Distributed slack";
```
geofjamg: You should change the name to not collide with the generator one

src/main/java/com/powsybl/openloadflow/network/LfBus.java
```diff
@@ -40,6 +40,12 @@
 
     double getLoadTargetP();
 
+    void setLoadTargetP(double loadTargetP);
+
+    boolean isParticipating();
```
geofjamg: I would not add isParticipating and getParticipationFactor to bus. Is is quite confusing with the one at generator level and in case of load distributed slack you only need to have  getLoadTargetP + setLoadTargetP + getLoadCount (to add) available to implement the algo.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+        if (factorSum == 0) {
+            throw new PowsyblException("No more load participating to slack distribution");
+        }
+        for (ParticipatingLoad participatingLoad : participatingLoads) {
+            participatingLoad.factor /= factorSum;
+        }
+    }
+
+    @Override
+    public OuterLoopStatus check(OuterLoopContext context) {
```
geofjamg: Note for later: this method could probably be shared with generator distrbuted slack.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public String getName() {
+        return "Distributed slack";
+    }
+
+    private static List<ParticipatingLoad> getParticipatingLoads(LfNetwork network) {
+        return network.getBuses()
+                .stream()
+                .filter(bus -> bus.getLoadTargetP() > 0 && bus.isParticipating())
```
geofjamg: I suspect this criteria to be incorrect: I would use bus.getLoadCount(). A bus with no load have getLoadTargetP() == 0 but should not be part of the slack dsitrib.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+                .map(bus -> new ParticipatingLoad(bus, bus.getParticipationFactor()))
```
geofjamg: You should use directly getLoadTargetP instead of  getParticipationFactor

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+        int loadsAtMin = 0;
+        Iterator<ParticipatingLoad> it = participatingLoads.iterator();
+        while (it.hasNext()) {
+            ParticipatingLoad participatingLoad = it.next();
+            LfBus load = participatingLoad.load;
+            double factor = participatingLoad.factor;
+
+            double targetP = load.getLoadTargetP();
+
+            double newTargetP = targetP - remainingMismatch * factor;
```
geofjamg: Be careful of the sign here I don't remenber the convention we are using.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+            }*/
+
+            if (newTargetP != targetP) {
+                if (LOGGER.isTraceEnabled()) {
+                    LOGGER.trace("Rescale '{}' active power target: {} -> {}",
+                            load.getId(), targetP * PerUnit.SB, newTargetP * PerUnit.SB);
+                }
+
+                load.setLoadTargetP(newTargetP);
+                done += targetP - newTargetP;
```
geofjamg: Here too, - newTargetP or + (+ newTargetP) ?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
annetill: Yes, you are right.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
annetill: Done.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
annetill: Done.

src/main/java/com/powsybl/openloadflow/network/LfBus.java
annetill: Done.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
annetill: Ok.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
annetill: I have decided to discarde from compensation the loads that have negative or null targetP. It is possible when we agregrate the distribution network for example. What do you think ?

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
annetill: Done.

src/main/java/com/powsybl/openloadflow/ac/DistributedLoadSlackOuterLoop.java
annetill: In fact, for a given mismatch, an increase in the production is equivalent to a decrease of loads active power and a decrease in the production is equivalent to an increase of loads. For sure, I am really not sure that my implementation makes it clear… what do you suggest ?

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -156,14 +160,18 @@ private double checkTargetV(double targetV) {
 
     void addLoad(Load load) {
         loads.add(load);
-        this.loadTargetP += load.getP0();
-        this.loadTargetQ += load.getQ0();
+        initialLoadTargetP = load.getP0();
```
sylvlecl: I think it should be an addition here: maybe we should have a unit test with several loads on one bus.

annetill: Yes, you are right. I am also going to add a unit test.

annetill:
About loads with negative active targets: it is not so clear for me that we have to discarde them from the slack distribution. A negative load could be the result of an aggregation of classical loads and generations on DSO network for example. The only test we have here is not to allow on a bus a positive target to become negative because we don't know if the load has the ability to produce. A solution could be to separate the constant part and the affine part of a load. But indeed, as a first approach, your are right it is more inuitive to discard them. What do you think @

PR #60, opened 2020-03-27
Support of LCC converter stations in the computation

annetill:
* LCC converter station are managed as loads.
* HVDC line does not appear as a branch.
* Add active losses modeling for VSC converter station.

Tests are missing. I want to use the powsybl-core conformity tests but I still have to work to reproduce the voltages of the SV values. 

LCC lines are not supported in the open loadflow.

It tries to support it. 

An exception is through when LCC converter station are found in a network.

As a first simple approach, LCC line are not converted as a branch. We only convert the LCC converter stations as loads. The loads have a reactive power computed from the active power and the converter station active power losses.

**Other information**:

I have noticed that VSC converter stations are not included in the slack distribution when the balance is done on generators. I think we have to add them.


annetill:
Hi Geoffroy, it is just a prototype. Some questions:
- I am still working on a small grid network with hvdc lines (from powsybl-core). Could I use the conformity test cases here for unit tests ?
- I wonder about my "update" fonction for LCC converter stations. It is in fact really similar as the setting function. 
- I put the reactive power to zero to start.

src/main/java/com/powsybl/openloadflow/network/impl/LfVscConverterStationImpl.java
```diff
@@ -33,10 +33,17 @@ public static LfVscConverterStationImpl create(VscConverterStation station) {
...
+        // If line.getConvertersMode() == HvdcLine.ConvertersMode.SIDE_1_RECTIFIER_SIDE_2_INVERTER then the active set point is positive.
+        // As the VSC converter station is at side 1, it means that it is rectifier: targetP should be negative.
+        // If line.getConvertersMode() == HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER then the active set point is negative.
+        // As the VSC converter station is at side 1, it means that it is inverter: targetP should be positive.
+        // If the VSC converter station is at side 2:
+        // If line.getConvertersMode() == HvdcLine.ConvertersMode.SIDE_1_RECTIFIER_SIDE_2_INVERTER then the active set point is positive.
+        // As the VSC converter station is at side 2, it means that it is inverter: targetP should be positive.
+        // If line.getConvertersMode() == HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER then the active set point is negative.
+        // As the VSC converter station is at side 2, it means that it is rectifier: targetP should be negative.
+        return line.getConverterStation1() == vscCs ? -line.getActivePowerSetpoint() : line.getActivePowerSetpoint();
```
annetill, line 46: Need to be fix: the active power set point is positive.

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -166,6 +168,26 @@ void addBattery(Battery battery) {
...
+        // If line.getConvertersMode() == HvdcLine.ConvertersMode.SIDE_1_RECTIFIER_SIDE_2_INVERTER then the active set point is positive.
+        // As the LCC converter station is at side 1, it means that it is rectifier: P should be positive.
+        // If line.getConvertersMode() == HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER then the active set point is negative.
+        // As the LCC converter station is at side 1, it means that it is inverter: P should be negative.
+        // If the LCC converter station is at side 2:
+        // If line.getConvertersMode() == HvdcLine.ConvertersMode.SIDE_1_RECTIFIER_SIDE_2_INVERTER then the active set point is positive.
+        // As the LCC converter station is at side 2, it means that it is inverter: P should be negative.
+        // If line.getConvertersMode() == HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER then the active set point is negative.
+        // As the LCC converter station is at side 2, it means that it is rectifier: P should be positive.
+        double p1 = line.getConverterStation1() == lccCs ? line.getActivePowerSetpoint() : -line.getActivePowerSetpoint();
```
annetill, line 184: Need to be fixed: the active power set point is positive.

PR #61, opened 2020-03-27
Avoid switching all buses PV to PQ

geofjamg:

No

Bug fix

In reactive limit outer loop, all PV buses could switch PQ (if reactive limit has been reach), leaving the network without any PV bus.

In reactive limit outer loop, we always keep at least one PV bus.
Among those who should switch PQ, we keep the strongest one, so the one at the highest nominal voltage and with the highest active power target.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
@@ -93,7 +93,9 @@ public class ReactiveLimitsOuterLoop implements OuterLoop {
         }
     }
 
-    private void switchPvPq(List<PvToPqBus> pvToPqBuses, EquationSystem equationSystem, VariableSet variableSet, int remainingPvBusCount) {
+    private OuterLoopStatus switchPvPq(List<PvToPqBus> pvToPqBuses, EquationSystem equationSystem, VariableSet variableSet, int remainingPvBusCount) {
+        OuterLoopStatus status = OuterLoopStatus.STABLE;
+
         int modifiedRemainingPvBusCount = remainingPvBusCount;
         if (modifiedRemainingPvBusCount == 0) {
             // keep one bus PV, the strongest one which is one at the highest nominal level and highest active power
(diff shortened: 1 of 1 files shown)


src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
```diff
@@ -63,35 +70,83 @@ private void switchPvPq(LfBus bus, EquationSystem equationSystem, VariableSet va
...
+            this.bus = bus;
+            this.q = q;
+            this.qLimit = qLimit;
+            this.limitDirection = limitDirection;
+        }
+    }
+
+    private void switchPvPq(List<PvToPqBus> pvToPqBuses, EquationSystem equationSystem, VariableSet variableSet, int remainingPvBusCount) {
+        int modifiedRemainingPvBusCount = remainingPvBusCount;
+        if (modifiedRemainingPvBusCount == 0) {
```
annetill: You need to add a test before the if statement ````if (pvToPqBuses.size() > 1) {...```` in order not to remove a PV bus when juste one PV bus is switching.

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
```diff
@@ -63,35 +70,83 @@ private void switchPvPq(LfBus bus, EquationSystem equationSystem, VariableSet va
...
+            // keep one bus PV, the strongest one which is one at the highest nominal level and highest active power
+            // target
+            PvToPqBus strongestPvToPqBus = pvToPqBuses.stream()
+                    .min(BY_NOMINAL_V_COMPARATOR
+                            .thenComparing(BY_TARGET_P_COMPARATOR)
```
annetill: I am not sure to understand all here: you first sort by nominal voltage the candidate PV buses (in fact it is a way to select the buses that regulate the highest voltage) and then by highest active power. What happens if we just have one PV bus at 400 kV but not with the highest active power ?

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
geofjamg: To void infinite loop I implemented another strategy: just consider outer loop stable if just one PV bus should have been switched PQ but not.

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
geofjamg: It is first sort on nominal V then target P, so yes if a generator is at higher nominalV but with a 1 MW targetP it will be chosen. Maybe I should remove the nominalV criteria and just focus on targetP?

annetill: No I think it is better to rely on nominalV criteria. It is the nominal voltage of PV bus, so it means that it is the target V of the generator if remote ?

geofjamg: No, in case of remove control, it is the nominal voltage of the generator not of the remote bus. I should probably fix this?

annetill: Yes, I think it will be better.

geofjamg: Done

annetill: I have performed some tests and I will approve this PR. Just a question before: we always remove the strongest PV bus from switching or it is just when all generators of the network are about to switch ?

Issue #49, opened 2020-02-27
Open Loadflow does not allow PQ buses to go back to PV bus


annetill:
* **Do you want to request a *feature* or report a *bug*?**

A feature.

* **What is the current behavior?**

In the code, a PV bus can switch to a PQ bus when reactive limits are reached. The bus is set to PQmin or PQmax and the flows are re-computed.    

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Some targets V are not followed by generators in regulation with the current algorithm. I am working on IOP dataset december and I can provide examples.

* **What is the expected behavior?**

A new feature could be to allow PQmin and PQmax buses to go back to PV buses after the re-computation of voltage V. Depending on that V, the PQ bus could go back to PV. 

* **What is the motivation / use case for changing the behavior?**

To have more PV buses in a network.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #66, opened 2020-04-07
Reactive limit outer loop, PQ -> PV switch

geofjamg:

#49 

feature

Generator reactive limits are taken into account in the loadflow in a Newton-Raphson outer loop by switch bus PV to PQ when reaching the limit.

Some buses that have been switched PQ are allowed to go back PV at next outer loop depending of calculated voltage.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -33,8 +33,12 @@
 
     private final double nominalV;
 
+    private boolean voltageControlCapacility = false;
+
     private boolean voltageControl = false;
 
+    private int voltageControlSwitchOffCount = 0;
```
annetill, line 40: I am not sure about the necessity of having "Off" in the variable name. Maybe ````voltageControlSwitchingCount````could be better ?

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
```diff
@@ -115,30 +123,95 @@ private OuterLoopStatus switchPvPq(List<PvToPqBus> pvToPqBuses, EquationSystem e
...
+        private final ReactiveLimitDirection limitDirection;
+
+        private PqToPvBus(LfBus bus, ReactiveLimitDirection limitDirection) {
+            this.bus = bus;
+            this.limitDirection = limitDirection;
+        }
+    }
+
+    private boolean switchPqPv(List<PqToPvBus> pqToPvBuses, EquationSystem equationSystem, VariableSet variableSet) {
+        int pqPvSwitchCount = 0;
```
annetill, line 169: Can we just add a small description here: "If Q is equal to Qmin and V is less than targetV, it means that the PQ bus can be unlocked in order to increase the reactive power and reach its targetV. If Q is equal to Qmax and V is more than targetV, it means that the PQ bus can be unlocked in order to decerase the reactive power and reach its target V." With your own words of course.

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -169,23 +186,24 @@ void addBattery(Battery battery) {
     private void add(LfGenerator generator, boolean voltageControl, double targetV, double targetQ,
                      LfNetworkLoadingReport report) {
         generators.add(generator);
-        boolean voltageControl2 = voltageControl;
+        boolean fixedVoltageControl = voltageControl;
```
annetill, line 189: What do you mean by "fixed" ?

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
geofjamg: Ok but in this case I have to increment when voltageControl is switched on and off and adapt the test for PV/PQ/PV max switch (we will have double value)

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
geofjamg: I mean repaired, because the input value is modified depending of additiona checks.
Maybe modifiedVoltageControl would be better?

annetill: Yes "modifiedVoltageControl" is better.

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
annetill: Ah okay, so no leave it like that, I understand know. Maybe just a small comment in the code ?

geofjamg: Comment added in the Javadoc

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
geofjamg: Added with your very clear wording!

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
geofjamg: Done

PR #68, opened 2020-04-11
Phase control support

geofjamg:

No

feature

We are able to support phase control in MW for 2 windings transformers in AC mode.
Will be done later:
 - DC mode
 - limiter mode in MW and A
 - control mode in A
 - phase tap changer attached to 3 windings transformers legs

**Other information**:


src/main/java/com/powsybl/openloadflow/equations/Variable.java
```diff
@@ -82,6 +86,10 @@ void updateState(LfNetwork network, double[] x) {
                 network.getBus(num).setAngle(Math.toDegrees(x[column]));
                 break;
 
+            case BRANCH_ALPHA1:
+                // TODO update tap position
```
annetill, line 90: Are you going to do that in this PR ?

src/main/java/com/powsybl/openloadflow/ac/equations/ClosedBranchSide2ReactiveFlowEquationTerm.java
```diff
@@ -29,8 +29,11 @@
 
     private double dq2dph2;
 
-    public ClosedBranchSide2ReactiveFlowEquationTerm(LfBranch branch, LfBus bus1, LfBus bus2, VariableSet variableSet) {
-        super(branch, bus1, bus2, variableSet);
+    private double dq2da1;
+
+    public ClosedBranchSide2ReactiveFlowEquationTerm(LfBranch branch, LfBus bus1, LfBus bus2, VariableSet variableSet,
+                                                     boolean deriveA) {
```
annetill, line 35: Just a question here: when we will add the aibility for ratio tap changer to control voltage, we will also introduce a other boolean to say that ratio ````r1```` can variate ? It will be maybe ````deriveR1```` (and ````deriveR```` is not possible), so maybe we have to rename ````deriveA```` in ````deriveA1```` ?

src/main/java/com/powsybl/openloadflow/equations/Equation.java
```diff
@@ -104,12 +102,22 @@ private static double getBusTargetV(LfBus bus) {
         }
     }
 
-    private static double getBranchPhi(LfBranch branch) {
+    private static double getBranchA(LfBranch branch) {
```
annetill, line 105: Here ````A```` is the same A that we found in ````deriveA```` ? In fact you choose ````deriveA```` because you compute the derivative of (A2 - A1) but A2 is fixed ?

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -64,10 +67,20 @@ public static LfBranchImpl create(Branch branch, LfBus bus1, LfBus bus2) {
                     .setB1(Transformers.getB1(twt) * zb)
                     .setR1(Transformers.getRatio(twt) / nominalV2 * nominalV1)
                     .setA1(Transformers.getAngle(twt));
+
+            PhaseTapChanger ptc = twt.getPhaseTapChanger();
+            if (ptc != null && ptc.isRegulating()) {
+                PhaseTapChanger.RegulationMode regulationMode = ptc.getRegulationMode();
+                if (regulationMode == PhaseTapChanger.RegulationMode.CURRENT_LIMITER) {
```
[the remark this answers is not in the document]
geofjamg: Yes, there is no difference in the equation system, between control and limiter mode. In both cases we target a flow (in MW or A).
  - in control mode, we activate the phase shifter at first outer loop
  - in limiter mode, we first de-activate phase shifter, after fist outer loop, we check if we are under the limit, for the one that are under the limit we activate them in the second outer loop for letting them going down to the limit.

This is why we often talk about phase shifter in "N" for control and phase shifter in "N-1" in limiter mode.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowPhaseShifterTest.java
```diff
@@ -106,4 +103,17 @@ public void tapPlusOneTest() {
         assertActivePowerEquals(-83.487, line2.getTerminal2());
         assertReactivePowerEquals(-22.169, line2.getTerminal2());
     }
+
+    @Test
+    public void flowControlTest() {
+        parameters.setPhaseShifterRegulationOn(true);
+        ps1.getPhaseTapChanger()
```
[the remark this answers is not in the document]
geofjamg: Yes, this is not implemented. We also count easily handle "remote flow". I remember there is use case for that in I don't known which country where we try to limit flow a line that is in parallel of the phase shifter. This has never been implemented in hades2 but here I do not see any reason to not do it. The way is modeled allows to not have the phase shift angle variable and the target flow on the same branch, right?

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowPhaseShifterTest.java
```diff
@@ -106,4 +103,17 @@ public void tapPlusOneTest() {
...
+                .setRegulationMode(PhaseTapChanger.RegulationMode.ACTIVE_POWER_CONTROL)
+                .setRegulating(true)
+                .setRegulationValue(83);
+
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        assertTrue(result.isOk());
+        assertActivePowerEquals(82.9, line2.getTerminal1());
```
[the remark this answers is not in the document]
geofjamg: Agree, this is a bug. I guess it is because we don't target the right side of the branch.

src/main/java/com/powsybl/openloadflow/equations/Variable.java
geofjamg: Yes, to do.

src/main/java/com/powsybl/openloadflow/ac/equations/ClosedBranchSide2ReactiveFlowEquationTerm.java
geofjamg: Agree it should be A1, in the code of this PR it is quite confusing.
The reason is that a2 is hardcoded to zero because never used when coming from IIDM data so a and a1 are the same everywhere (see PiModel code)
But I think we should support properly everywhere a1 and a2 variables so that if one connected to another model than IIDM where it is relevant to have a phase shift on side 2 to easily map it to OLF internal model.

geofjamg: And yes, this is here that we will add r1 and r2 derive option for ratio tap changer. Maybe we should have a DeriveConfig object to avoid many parameters when adding ratio tap changers.

src/main/java/com/powsybl/openloadflow/equations/Equation.java
geofjamg: I will change that allow both a1 and a2 phase shifting everywhere in equation system even if not used right now when connected to IIDM model. Here because of a2=0 hardcoded we confuse a1 - a2 and a1.

geofjamg:
> Just some questions to well understand the big picture before making choices for this active power flow regulation. The test is well working and I am going to find a real test case to validate this work in progress.

Thanks for the feedback, this PR is right now just a POC and there is still many things to implement (like the outer loop which is empty, network update, A ou MW mode, flow terminal to control or limit, fix a1 a2 in equation system, etc) before merging it.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -158,7 +162,69 @@ private static void createNonImpedantBranch(VariableSet variableSet, EquationSys
...
+            p2 = new OpenBranchSide1ActiveFlowEquationTerm(branch, bus2, variableSet);
+            q2 = new OpenBranchSide1ReactiveFlowEquationTerm(branch, bus2, variableSet);
+        }
+
+        if (p1 != null) {
+            equationSystem.createEquation(bus1.getNum(), EquationType.BUS_P).addTerm(p1);
+            branch.setP1(p1);
+
+            if (creationParameters.isPhaseControl()) {
+                createBranchActivePowerTargetEquation(branch, PhaseControl.ControlledSide.ONE, equationSystem, p1);
```
annetill, line 205: Maybe we have to put a comment here to say that only local regulation would be supported ?

annetill: Do we always have side 1 in IIDM = side 1 in the LFBranch ?

src/main/java/com/powsybl/openloadflow/ac/equations/AbstractClosedBranchAcFlowEquationTerm.java
```diff
@@ -30,24 +29,33 @@
...
-        PiModel piModel = branch.getPiModel();
-        a1 = piModel.getA1();
-        a2 = piModel.getA2();
+        ImmutableList.Builder<Variable> variablesBuilder = ImmutableList.<Variable>builder()
+                .add(v1Var, v2Var, ph1Var, ph2Var);
+        if (deriveA1) {
+            a1Var = variableSet.getVariable(branch.getNum(), VariableType.BRANCH_ALPHA1);
+            variablesBuilder.add(a1Var);
+        }
+        if (deriveA2) {
```
annetill, line 54: In a way our code is not so clear: dot we allow or not to have a PTC at side 2 ? The question will be raised for RTC too ? I think that if we Don't allow it from the Pi Model, we have to be consistent for the rest of the code. What do you think ?

src/main/java/com/powsybl/openloadflow/equations/EquationSystem.java
```diff
@@ -67,6 +59,26 @@ private void update() {
             invalide = false;
         }
 
+        private void reIndex() {
```
annetill, line 62: Just a question for later: all the AS will work around this function "reindex()" ? I mean we will update the equation system at this level and not at a lower level (in the admittance matrix) ?

src/main/java/com/powsybl/openloadflow/equations/Variable.java
```diff
@@ -82,6 +104,14 @@ void updateState(LfNetwork network, double[] x) {
                 network.getBus(num).setAngle(Math.toDegrees(x[column]));
                 break;
 
+            case BRANCH_ALPHA1:
+                network.getBranch(num).getPiModel().setA1(x[column]);
+                break;
+
+            case BRANCH_ALPHA2:
```
annetill, line 111: Same remark. Do we really allow an A2 ?

src/main/java/com/powsybl/openloadflow/network/PiModelArray.java
```diff
@@ -0,0 +1,126 @@
...
+    @Override
+    public void roundA1ToClosestTap() {
+        if (Double.isNaN(a1)) {
+            return; // nothing to do because a1 has not been modified
+        }
+
+        // find tap position with the closest a1 value
+        double smallestDistance = Math.abs(a1 - getModel().getA1());
+        for (int p = lowTapPosition; p < lowTapPosition + models.size(); p++) {
+            double distance = Math.abs(a1 - models.get(p).getA1());
```
annetill, line 118: Here we assume that we choose the closest tap from a distance criteria. In fact, we have 2 candidate taps: the lower one and the hupper one. I will keep that in mind for the deadband management (and I will put a comment).

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -35,39 +35,95 @@
...
+
+        return new LfBranchImpl(bus1, bus2, piModel, null, line);
+    }
+
+    private static LfBranchImpl createTransformer(TwoWindingsTransformer twt, LfBus bus1, LfBus bus2, double nominalV1,
+                                                  double nominalV2, double zb) {
+        PiModel piModel;
+        PhaseControl phaseControl = null;
+
+        PhaseTapChanger ptc = twt.getPhaseTapChanger();
```
annetill, line 61: [Memo] We have to put an issue to say that phase control on T3W will not be supported in the first version.

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -35,39 +35,95 @@
...
+        if (ptc != null
+                && ptc.isRegulating()
+                && ptc.getRegulationMode() != PhaseTapChanger.RegulationMode.FIXED_TAP) {
+            PhaseTapChanger.RegulationMode regulationMode = ptc.getRegulationMode();
+            PhaseControl.ControlledSide controlledSide;
+            if (ptc.getRegulationTerminal() == twt.getTerminal1()) {
+                controlledSide = PhaseControl.ControlledSide.ONE;
+            } else if (ptc.getRegulationTerminal() == twt.getTerminal2()) {
+                controlledSide = PhaseControl.ControlledSide.TWO;
```
annetill, line 70: I am not sure to understand : if PTC are always located on side 1 in IIDM, where do we put the PTC in the LFBranch ? If it is also always side 1 (remember for example that in the Pi Model, R2 is set to zero), do we really support a A1 variable and a side 2 for the regulating terminal ?

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public String getType() {
+        return "Phase control";
+    }
+
+    @Override
+    public OuterLoopStatus check(OuterLoopContext context) {
+        OuterLoopStatus status = OuterLoopStatus.STABLE;
+
+        if (context.getIteration() == 0) {
```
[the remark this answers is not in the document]
geofjamg: Yes, fhe first NR run is not part of an outerloop. 
Iteration here is iteration for this type of outerloop (PhaseControlOuterLoop), so iteration 0 is not necessarily the second NR run. It is guaranty to be the case only if you have one outer loop type configured.

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
```diff
@@ -0,0 +1,79 @@
...
+                    piModel.roundA1ToClosestTap();
+                    double roundedA1Value = piModel.getA1();
+                    LOGGER.info("Round phase shift of '{}': {} -> {}", branch.getId(), a1Value, roundedA1Value);
+
+                    // if at least one phase shifter has been switched off wee need to continue
+                    status = OuterLoopStatus.UNSTABLE;
+                }
+            }
+        } else if (context.getIteration() == 1) {
+            // at second iteration we switch on phase control for branches that are in limiter mode
```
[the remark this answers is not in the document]
geofjamg: Yes same outer loop but different iteration.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
geofjamg: We should indeed add a TODO to show that this part of the code will have to be updated for remote control.
Yes side 1 in IIDM is always side 1 in LF.

src/main/java/com/powsybl/openloadflow/ac/equations/AbstractClosedBranchAcFlowEquationTerm.java
geofjamg: There is no PTC on side 2 in IIDM. I kept most of the code for side2 in the very hypothetical case of OLF would be integrated with another data model. 
But maybe you are right we should clean a2 and r2 everywhere in the code.

src/main/java/com/powsybl/openloadflow/equations/EquationSystem.java
geofjamg: Exactly.

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
geofjamg: t3w is not handled here it is just for t2w. t3w code is in LfLegBranch

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
geofjamg: The side of the phase shift in the PI model is a different thing from the controlled side which will be later to allowed to not only be side 1 ou 2 of the phase shifter but another remote terminal.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
annetill: I understand that in the code, we can only have a variation on A1 but we can have 2 controller sides: side 1, that means a local control and side 2, that means like a remote control. Are we ok about that ?

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
annetill: Ok.

geofjamg:
I removed everything regarding a2 and r2 in the code except in the equations as a constant to allow if one day necessary re-add then again.
I think the code is now less confusing.
This PR is large and I think we should merge it as it is and improve it in further PR.

annetill:
I have performed some validations on real test cases and it is well working when the voltage init mode is set to ````voltageInitMode: UNIFORM_VALUES````. In case of ````voltageInitMode: DC_VALUES```` it is not working. Do you have any idea ?

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -142,7 +142,7 @@ private static void createNonImpedantBranch(VariableSet variableSet, EquationSys
             // create voltage magnitude coupling equation
             // 0 = v1 - v2 * rho
             PiModel piModel = branch.getPiModel();
-            double rho = piModel.getR2() / piModel.getR1();
+            double rho = 1 / piModel.getR1();
```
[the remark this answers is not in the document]
geofjamg: I revert R2 and A2 as a constant here and the 2 other points mentioned but it doesn't change anything as r2 and a2 when connected to IIDM won't ever be different from 1 and 0.
r2 != rated2
but 
r1 = rated2/rated1
r2 = 1

geofjamg:
> I have performed some validations on real test cases and it is well working when the voltage init mode is set to `voltageInitMode: UNIFORM_VALUES`. In case of `voltageInitMode: DC_VALUES` it is not working. Do you have any idea ?

Maybe we should also init the a1 in the voltage initilializer as when phase shifter is activated a1 is also a state variable.

PR #70, opened 2020-04-15
The lossFactor is a percentage (bug fix).

annetill:
Adapt unit tests.

LossFactor is considered without unit.

LossFactor is considered as a percentage.

**Other information**:

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowLccTest.java
+++ b/src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowLccTest.java
@@ -122,7 +122,7 @@ public class AcLoadFlowLccTest {
                 .setConnectableBus("b3")
                 .setBus("b3")
                 .setPowerFactor(0.8f)
-                .setLossFactor(0.1f)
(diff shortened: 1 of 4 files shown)


src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowLccTest.java
```diff
@@ -122,7 +122,7 @@ private Network createNetwork() {
                 .setConnectableBus("b3")
                 .setBus("b3")
                 .setPowerFactor(0.8f)
-                .setLossFactor(0.1f)
+                .setLossFactor(1.1f)
```
geofjamg, line 125: Not clear to me this change, in the previous code it was (1 + a) and now it is (1 + a/100). So it was already a deviation but not in % but with a [0, 1] value? Why was the loss factor 0.1.

annetill: In fact I just changed the value to have a more realistic value. 1.1% of losses during the AC/DC convertion. Before it was 10%...

PR #73, opened 2020-04-20
Fix distributed stack on load

geofjamg:

No

Buf fix

Negative loads are discarded from network. 

Negative loads are discarded from slack distribution

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
@@ -399,7 +399,7 @@ public class LfBusImpl extends AbstractLfBus {
         updateGeneratorsState(voltageControl ? calculatedQ + loadTargetQ : generationTargetQ, reactiveLimits);
 
         // update load power
-        double factor = loadTargetP / initialLoadTargetP;
+        double factor = initialLoadTargetP > 0 ? loadTargetP / initialLoadTargetP : 1;
         for (Load load : loads) {
             load.getTerminal()
                     .setP(load.getP0() >= 0 ? factor * load.getP0() : load.getP0())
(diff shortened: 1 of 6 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -400,20 +399,16 @@ public void updateState(boolean reactiveLimits) {
         updateGeneratorsState(voltageControl ? calculatedQ + loadTargetQ : generationTargetQ, reactiveLimits);
 
         // update load power
+        double dp = positiveLoadCount > 0 ? (loadTargetP - initialLoadTargetP) / positiveLoadCount : 0;
```
annetill, line 402: Hi Geoffroy, thank you for the fix ! But the algorithm here is quite different that the first one: if you have more than one load in a bus, the slack has an equal distribution between them. If you have a big one and a very small one, it could lead to problems. I am not sure that the ````positiveLoadCount```` is needed. Do you see ? You can a positive load in a bus in the unit test.

annetill:
Hi Geoffroy, can you review the small fixes I have done ? Thanks !

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -400,20 +399,15 @@ public void updateState(boolean reactiveLimits) {
         updateGeneratorsState(voltageControl ? calculatedQ + loadTargetQ : generationTargetQ, reactiveLimits);
 
         // update load power
+        double factor = initialLoadTargetP > 0 ? loadTargetP / initialLoadTargetP : 1;
```
geofjamg, line 402: > 0 or != 0 ?

annetill: !=0 ? I Don't want to divide by zero.

PR #82, opened 2020-05-10
Sensitivity analysis

geofjamg:

No

feature

This is a try for sensitivity analysis with OLF.

**Other information**:


src/main/java/com/powsybl/openloadflow/sensi/OpenSensitivityAnalysisProvider.java
```diff
@@ -0,0 +1,99 @@
...
+@AutoService(SensitivityAnalysisProvider.class)
+public class OpenSensitivityAnalysisProvider implements SensitivityAnalysisProvider {
+
+    private static final String NAME = "OpenSensitivityAnalysis";
+
+    private final DcSensitivityAnalysis dcSensitivityAnalysis;
+
+    private final AcSensitivityAnalysis acSensitivityAnalysis;
+
+    public OpenSensitivityAnalysisProvider(MatrixFactory matrixFactory) {
```
Djazouli, line 37: Don't we also need an empty constructor 
`public OpenSensitivityAnalysisProvider()` ? Otherwise the service loader (called by PlatformConfigNamedProvider) would throw 
` java.lang.NoSuchMethodException: com.powsybl.openloadflow.sensi.OpenSensitivityAnalysisProvider.<init>()` when trying to load the Service. 
Or is there a way to provide the matrixFactory to the service loader ?

floriand-e2r: indeed, we do need to add such a constructor, good catch! we should add an empty constructor which takes `SparseMatrixFactory` as default matrix factory.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -0,0 +1,230 @@
...
+        super(matrixFactory);
+    }
+
+    private SensitivityValue createBranchSensitivityValue(LfNetwork lfNetwork, EquationSystem equationSystem, String branchId,
+                                                          SensitivityFactor<?, ?> factor, SensitivityFactorGroup factorGroup,
+                                                          DenseMatrix states) {
+        LfBranch lfBranch = lfNetwork.getBranchById(branchId);
+        ClosedBranchSide1DcFlowEquationTerm p1 = equationSystem.getEquationTerm(SubjectType.BRANCH, lfBranch.getNum(), ClosedBranchSide1DcFlowEquationTerm.class);
+        double value = Math.abs(p1.calculate(states, factorGroup.getIndex()) * PerUnit.SB);
+        return new SensitivityValue(factor, value, 0, 0);
```
Djazouli, line 106: Do we not care about function reference and variable reference, or is this a todo ?

geofjamg: Yes to do

src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystem.java
```diff
@@ -71,7 +71,7 @@ private static void createImpedantBranch(VariableSet variableSet, EquationSystem
                                              DcEquationSystemCreationParameters creationParameters, LfBranch branch,
                                              LfBus bus1, LfBus bus2) {
         if (bus1 != null && bus2 != null) {
-            boolean deriveA1 = creationParameters.isForcePhaseControlOffAndAddAngle1Var() && branch.getDiscretePhaseControl() != null;
+            boolean deriveA1 = creationParameters.isForcePhaseControlOffAndAddAngle1Var() && branch.canControlPhase();
```
annetill, line 74: I prefer `hasPhaseControlCapability` for uniformity.

geofjamg: Done

annetill (commented):
Indeed I have merged the PR as we add a talk to manage last details. For me, the main issue is to remain clear enough with a1 variable in case of a security analysis and in case of a phase control outer loop (not available). Here, we do not rely on phase control objects and we have to be aware of that. Two other remarks:
- A lot of code smells comes from powsybl-core ;
- The test coverage is a bit low and has to be increased when improving this first version.

PR #91, opened 2020-05-16
Fix reactive power distribution equation

geofjamg:

No

Bug fix

3 fixes:
  - open branches are missing at reactive power distribution equation creation
  - controller bus with loads are not handled resulting to an incorrect generation reactive power distribution
  - controller bus with shunts are not handled resulting to an incorrect generation reactive power distribution 

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
@@ -69,7 +69,7 @@ public class ReactiveLimitsOuterLoop implements OuterLoop {
 
     private void updateControlledBus(LfBus controlledBus, EquationSystem equationSystem, VariableSet variableSet) {
         // clean reactive power distribution equations
-        controlledBus.getControllerBuses().forEach(b -> equationSystem.removeEquation(b.getNum(), EquationType.LOAD_Q_DIFF));
+        controlledBus.getControllerBuses().forEach(b -> equationSystem.removeEquation(b.getNum(), EquationType.ZERO_Q));
 
         // controlled bus has a voltage equation only if one of the controller bus has voltage control on
         List<LfBus> controllerBusesWithVoltageControlOn = controlledBus.getControllerBuses().stream()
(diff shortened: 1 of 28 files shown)


src/main/java/com/powsybl/openloadflow/equations/Equation.java
```diff
@@ -144,7 +161,10 @@ void initTarget(LfNetwork network, double[] targets) {
                 targets[row] = getBranchTarget(network.getBranch(num), PhaseControl.Unit.A);
                 break;
 
-            case ZERO_Q:
+            case LOAD_Q_DIFF:
+                targets[row] = getLoadQDiff(network, num, getData());
```
annetill, line 165: If you add a target different than zero, it will be impossible to keep the reactive distibution given by the qPercent. This distribution is satisfied because this target is equal to zero in the master branch. But the idea I like in your PR is quite an optimization of the reactive power used in the network. I have noticed with your small test that we don't managed this and sometimes less reactive power is needed because the network could be help by loads of shunts. I am not sure that what you want to do here is local Indeed.

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
```diff
@@ -69,7 +69,7 @@ private void switchPvPq(LfBus bus, EquationSystem equationSystem, VariableSet va
 
     private void updateControlledBus(LfBus controlledBus, EquationSystem equationSystem, VariableSet variableSet) {
         // clean reactive power distribution equations
-        controlledBus.getControllerBuses().forEach(b -> equationSystem.removeEquation(b.getNum(), EquationType.ZERO_Q));
```
annetill, line 72: I think that you should keep the type ````ZERO_Q````. Shunts and loads connected to generators are minor in real cases and we understand well the type ````ZERO_Q````.

geofjamg: Done

annetill:
Sorry Geoffroy, but I Don't know how to modify the unit test ````GeneratorTargetVoltageInconsistencyTest```` as an exception is not thrown but just an error is the log. Can you help me with that ? Thanks !

geofjamg:
> Sorry Geoffroy, but I Don't know how to modify the unit test `GeneratorTargetVoltageInconsistencyTest` as an exception is not thrown but just an error is the log. Can you help me with that ? Thanks !

We cannot assert on logs, so the only thing we can do is to check that calculated voltage on controlled bus has the right value

PR #102, opened 2020-05-29
Parallel non impedant branch support

geofjamg:

No

Bug fix

When we have 2 parallel non impedant branches, we end up with a "klu singular issue".

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AbstractBranchAcFlowEquationTerm.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AbstractBranchAcFlowEquationTerm.java
@@ -33,6 +33,9 @@ abstract class AbstractBranchAcFlowEquationTerm extends AbstractEquationTerm {
     protected AbstractBranchAcFlowEquationTerm(LfBranch branch) {
         this.branch = Objects.requireNonNull(branch);
         PiModel piModel = branch.getPiModel();
+        if (piModel.getR() == 0 && piModel.getX() == 0) {
+            throw new IllegalArgumentException("Non impedant branch not supported: " + branch.getId());
+        }
         r1 = piModel.getR1();
         b1 = piModel.getB1();
         b2 = piModel.getB2();
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AbstractBranchAcFlowEquationTerm.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AbstractBranchAcFlowEquationTerm.java
@@ -34,9 +34,6 @@ abstract class AbstractBranchAcFlowEquationTerm extends AbstractEquationTerm {
     protected AbstractBranchAcFlowEquationTerm(LfBranch branch) {
         this.branch = Objects.requireNonNull(branch);
         PiModel piModel = branch.getPiModel();
-        if (piModel.getR() == 0 && piModel.getX() == 0) {
-            throw new IllegalArgumentException("Non impedant branch not supported: " + branch.getId());
-        }
         r1 = piModel.getR1();
         b1 = piModel.getB1();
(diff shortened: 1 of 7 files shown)


src/main/java/com/powsybl/openloadflow/ac/equations/AbstractBranchAcFlowEquationTerm.java
```diff
@@ -34,9 +34,6 @@
     protected AbstractBranchAcFlowEquationTerm(LfBranch branch) {
         this.branch = Objects.requireNonNull(branch);
         PiModel piModel = branch.getPiModel();
-        if (piModel.getR() == 0 && piModel.getX() == 0) {
```
annetill, line 37: I falled in that exception during my tests so I removed it.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -265,13 +263,20 @@ private static void createImpedantBranch(LfBranch branch, LfBus bus1, LfBus bus2
...
+
         for (LfBranch branch : network.getBranches()) {
             LfBus bus1 = branch.getBus1();
             LfBus bus2 = branch.getBus2();
             PiModel piModel = branch.getPiModel();
             if (piModel.getZ() < DcEquationSystem.LOW_IMPEDANCE_THRESHOLD) {
                 if (bus1 != null && bus2 != null) {
-                    createNonImpedantBranch(variableSet, equationSystem, branch, bus1, bus2);
+                    String bus1Bus2Key = bus1.getNum() < bus2.getNum() ? bus1.getNum() + "_" + bus2.getNum() : bus2.getNum() + "_" + bus1.getNum();
+                    if (!nonImpedantBranchesAlreadyCreated.contains(bus1Bus2Key)) {
```
annetill, line 276: I am not sure to well understand what does it mean: 
- When you want to create a low impedance branch you check if a similar branch already exists between the same bus ;
- If yes, with the same rho and angle, what do you do ? Nothing ? We delete the branch ?

src/main/java/com/powsybl/openloadflow/ac/equations/AbstractBranchAcFlowEquationTerm.java
annetill: Maybe we can just for here to replace by a min low impedance line as it could be complex and write an issue about it.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
geofjamg: This is just at equation level, so we do not change the network for avoid to create redundant equation. I am going to investigate more deeply why we need to skip redundant coupling equation to better understand what the problem is.
Also could you commit your test case with non impedant branch loop?

geofjamg:
@annetill I fully rework this PR and now we are able to deal with any kind on non impedant branch topology (to test but it works with your loop unit test). I also add support for non impedant branch connected to controller bus.

annetill:
It seems that I am still falling in a KLU issue when initializing the state with parameter DC_VALUES. I Don't met any issue if I start with UNIFORM_VALUES.

geofjamg:
> It seems that I am still falling in a KLU issue when initializing the state with parameter DC_VALUES. I Don't met any issue if I start with UNIFORM_VALUES.

Oups I forgot to update the DC part, with same strategy...

geofjamg:
> > It seems that I am still falling in a KLU issue when initializing the state with parameter DC_VALUES. I Don't met any issue if I start with UNIFORM_VALUES.
> 
> Oups I forgot to update the DC part, with same strategy...

Done, is it ok now?

PR #105, opened 2020-06-02
Improve voltage init from IIDM values

geofjamg:

No

feature

This is an improvement of PREVIOUS_VALUES voltage initialization mode that allow to use it when there are dangling lines and 3 windings transformers. Voltage magnitude and angle of fictitious buses can be specified with IIDM properties.  

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
@@ -16,8 +16,12 @@ public class LfStarBus extends AbstractFictitiousLfBus {
 
     private final ThreeWindingsTransformer t3wt;
 
+    private final double nominalV;
+
     public LfStarBus(ThreeWindingsTransformer t3wt) {
+        super(Networks.getPropertyV(t3wt), Networks.getPropertyAngle(t3wt));
         this.t3wt = t3wt;
+        nominalV = t3wt.getLeg1().getTerminal().getVoltageLevel().getNominalV();
     }
 
     @Override
@@ -27,6 +31,6 @@ public class LfStarBus extends AbstractFictitiousLfBus {
 
     @Override
     public double getNominalV() {
-        return t3wt.getLeg1().getTerminal().getVoltageLevel().getNominalV();
+        return nominalV;
     }
 }
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
@@ -33,4 +33,10 @@ public class LfStarBus extends AbstractFictitiousLfBus {
     public double getNominalV() {
         return nominalV;
     }
+
+    @Override
+    public void updateState(boolean reactiveLimits) {
+        Networks.setPropertyV(t3wt, v);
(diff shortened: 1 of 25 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
```diff
@@ -58,11 +58,7 @@ private static double getDoubleProperty(Identifiable identifiable, String name)
 
     private static void setDoubleProperty(Identifiable identifiable, String name, double value) {
         Objects.requireNonNull(identifiable);
-        if (Double.isNaN(value)) {
-            identifiable.setProperty(name, null);
-        } else {
-            identifiable.setProperty(name, Double.toString(value));
-        }
+        identifiable.setProperty(name, Double.toString(value));
```
geofjamg, line 61: I think we should keep the previous behaviour, so remove the property in case of NaN. I does not make sense to keep a NaN in the property and is not sync with getDoubleProperty

src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
```diff
@@ -21,7 +21,7 @@
     public LfStarBus(ThreeWindingsTransformer t3wt) {
         super(Networks.getPropertyV(t3wt), Networks.getPropertyAngle(t3wt));
         this.t3wt = t3wt;
-        nominalV = t3wt.getLeg1().getTerminal().getVoltageLevel().getNominalV();
+        nominalV = t3wt.getRatedU0();
```
geofjamg, line 24: You are right we need this to be consistent with the other part of the code. But maybe we should rethink about this as it is not consistant with the way we per unit for the other node. Could we have a bu here with a missing fixed ratio (ratedU1/ratedU0)?

src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
annetill: In fact, I falled in a null pointer exception with the previous code because property cannot be null. So maybe, in case of Nan, we don't set the property ?

geofjamg: Oups, identifiable.setProperty(name, null) should have been identifiable.remove(name)

src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
annetill: Maybe. The default modeling of three windings transformer is ratedU0=ratedU1 but it is configurable.

src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
annetill: I have not this function remove from my identifiable :-)

annetill: identifiable.getPropertyNames().remove(name) ?

geofjamg:
@annetill The problem with that PR is that it does not only contains what is in the title as it only have a change in the way we per unit 3 w transfos. Do we let it as it is and merge everything or do we revert per unit change to be consistent with master way and handle per unit fix and validation in another PR?

PR #118, opened 2020-06-17
Split shunt admittance support

annetill:

No.

The shunt admittance can be split in two equals admittances, each on each side of the series impedance for transformers.

We only support G1 and B1 (side 1) of the transformer.

The option will be available in the open loadflow.

**Other information**:

A unit test is missing.


src/test/resources/n2.json
```diff
@@ -46,6 +46,7 @@
     "num2" : 2,
     "r" : 0.0013850415512465374,
     "x" : 0.06925207756232687,
+    "a1" : 0.3490658503988659,
```
annetill, line 49: @geofjamg I don't understand why this unit test has changed.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcLoadFlowParameters.java
```diff
@@ -38,10 +38,12 @@
 
     private final boolean minImpedance;
 
+    private final boolean specificCompatibility;
+
     public AcLoadFlowParameters(SlackBusSelector slackBusSelector, VoltageInitializer voltageInitializer,
```
annetill, line 43: @geofjamg It seems that I only support the specific compatibility for AC loadflow engine. Why ?

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -80,18 +80,22 @@ private static LfBranchImpl createTransformer(TwoWindingsTransformer twt, LfBus
...
 
                 for (int position = ptc.getLowTapPosition(); position <= ptc.getHighTapPosition(); position++) {
-                    PhaseTapChangerStep step = ptc.getStep(position);
-                    double r = twt.getR() * (1 + step.getR() / 100) / zb;
-                    double x = twt.getX() * (1 + step.getX() / 100) / zb;
-                    double g1 = twt.getG() * (1 + step.getG() / 100) * zb;
-                    double b1 = twt.getB() * (1 + step.getB() / 100) * zb;
-                    double r1 = twt.getRatedU2() / twt.getRatedU1() * ptc.getCurrentStep().getRho() / nominalV2 * nominalV1;
-                    double a1 = Math.toRadians(step.getAlpha());
+                    ptc.setTapPosition(position);
```
geofjamg, line 83: By setting the current tap position, you change the input network and you lost this information of the tap position.
We probably should use methods like Transformers.getG1(twt, position, specificCompatibility) so with an additional position parameter to not change the network.

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -77,6 +77,9 @@ private static LfBranchImpl createTransformer(TwoWindingsTransformer twt, LfBus
             }
 
             if (controlledSide != null) {
+
+                int currentPosition = ptc.getTapPosition();
```
geofjamg, line 81: It should not be done like this, network has not to be modified at this stage, even temporarily.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcLoadFlowParameters.java
```diff
@@ -38,10 +38,12 @@
 
     private final boolean minImpedance;
 
+    private final boolean specificCompatibility;
```
geofjamg, line 41: Now that we have switched to powsybl 3.4 we can rename it twtSplitShuntAdmittance?

annetill: Yes, you're right.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcLoadFlowParameters.java
geofjamg: I also added this option to DC mode

src/main/java/com/powsybl/openloadflow/network/impl/Transformers.java
```diff
@@ -27,26 +27,38 @@ private Transformers() {
...
         }
         if (twt.getPhaseTapChanger() != null) {
-            rho *= twt.getPhaseTapChanger().getCurrentStep().getRho();
+            rho *= twt.getPhaseTapChanger().getStep(ptcPosition).getRho();
         }
         return rho;
     }
 
+    private static int getCurrentPosition(RatioTapChanger rtc) {
+        return rtc != null ? rtc.getTapPosition() : -1;
```
geofjamg, line 45: The problem with this design is that -1 could be a valid tap position

annetill: So what else should I do ?

annetill: Indeed, this value "-1" is never used in the code because we have some checks after:
````
public static double getRatioLeg(ThreeWindingsTransformer twt, ThreeWindingsTransformer.Leg leg, int rtcPosition, int ptcPosition) {
        double rho = twt.getRatedU0() / leg.getRatedU();
        if (leg.getRatioTapChanger() != null) {
            rho *= leg.getRatioTapChanger().getStep(rtcPosition).getRho();
        }
        if (leg.getPhaseTapChanger() != null) {
            rho *= leg.getPhaseTapChanger().getStep(ptcPosition).getRho();
        }
        return rho;
    }
````
but I need to return an integer in my function, even if I don't have any RTC or PTC. Do you have an other suggestion ?

geofjamg:
@annetill I replace int by Integer and -1 by null. Is it ok for you?

Issue #106, opened 2020-06-05
Modeling of tap changer voltage control


annetill:
* **Do you want to request a *feature* or report a *bug*?**

A feature. 

* **What is the current behavior?**

In the current version, only the current tap is used to build the equation system. But ratio tap changer could have voltage regulating capabilities.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

Ratio tap changers can regulate voltage locally and remotely. We have planned to model this regulation in the first NR in seeing the ratio of these branches as a continuous variable and to discretize after the first resolution. 

What about shared control ? It is allowed ? Observed ?

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

The same approach have been followed for active power control of phase tap changer.

annetill:
I don't close this issue because I have still some factorization issues of some CGMES test cases. Then, if a bus is both controlled by a generator and a transformer, the transformer voltage control is disabled.

PR #127, opened 2020-07-06
Voltage control of transformer: remote and local

annetill:

#106 

**Other information**:


src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -0,0 +1,215 @@
...
+        parameters.setTransformerVoltageControlOn(true);
+        twt.getRatioTapChanger()
+                .setTargetDeadband(0)
+                .setRegulating(true)
+                .setTapPosition(0)
+                .setRegulationTerminal(twt.getTerminal2())
+                .setTargetV(34.0);
+
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        // assertTrue(result.isOk()); FIXME
```
annetill, line 212: @geofjamg it is just a first version of voltage control by transformer and it is not working. At this test, the norm of my cost function is ````NaN````. Do you have any idea of the problem ?

geofjamg: This is because the way tap changer voltage target equation is defined in AcEquationSystem.createBranchVoltageTargetEquation.
You should define a new equation type for tap changer target voltage (like BUS_TRANSFO_V)) because if you look at Equation.initTargets for type BUS_V, we look for a generator target voltage and not a tap changer target voltage so result is nan and this nan is added to targets vector before first NR.

src/main/java/com/powsybl/openloadflow/ac/outerloop/TransformerVoltageControlOuterLoop.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(TransformerVoltageControlOuterLoop.class);
+
+    @Override
+    public String getType() {
+        return "Transformer voltage control";
+    }
+
+    @Override
+    public OuterLoopStatus check(OuterLoopContext context) {
```
geofjamg, line 32: I don't known if it is clear for you what has to be done here but here is a raw extract (to translate) of what is done in H2:

Articulation des 3 load flows pour la simulations des regleurs en charges
 * Pour rappel:
 * -LF1 = gpesTHT PV, grpes HT PV, regleurs fixes
 * -LF2 = gpesTHT PV, grpes HT PQ, regleurs variables
 * -LF3 = gpesTHT PV, grpes HT PV, regleurs fixes (apres discretisation)

annetill: I am not sure that we have to follow these 3 steps here. I think that I am going to follow simply what we have done for phase tap changer. It will be easy after that to make a more business and complex outerloop. In the EMF specifications, it is just written:
When calculating the steady state solution, a full Newton Raphson power flow algorithm is to be used, using the following settings: 
•    Q limits shall be respected (also for slack node/swing bus) 
•    Transformer tap adjustment is set to enabled
•    Switched shunt adjustment is set to enabled
Transformer  tap  and  switch  shunt  control  is  used  during  power  flow  calculation  to  enable regulating control during contingency analysis (where the solved system state is used for the base case). In case of non-convergence, switched shunt and transformer tap regulation is disabled and power flow calculation is started again.

geofjamg: Even if not specified you need to follow these 3 steps in order to emulate a realistics tap changer and generation co ntrol behaviour.

annetill: Do we have a way to have a generic outerloop that can be customize easily to simulate real world ?

geofjamg: Yes just one outerloop and a different behaviour at iteration 0, 1 and 2

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -0,0 +1,215 @@
...
+        twt.getRatioTapChanger()
+                .setTargetDeadband(0)
+                .setRegulating(true)
+                .setTapPosition(0)
+                .setRegulationTerminal(twt.getTerminal2())
+                .setTargetV(34.0);
+
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        assertTrue(result.isOk());
+        assertVoltageEquals(34.433, bus3);
```
annetill, line 213: @geofjamg I have a small difference but I am not sure to understand why...

annetill:
@geofjamg I have finished my first version. Can you make a review in order to be technically good ? It could be a good starting point for a functional definition of the transformer voltage control outerloop with our colleagues.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
geofjamg: Because we round the r1?

geofjamg: There is something missing in the algo, maybe after rounding and fixing ratio it is to generation behind transformers du help reaching the target voltage?

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -0,0 +1,442 @@
...
+                .setTargetDeadband(0)
+                .setRegulating(true)
+                .setTapPosition(0)
+                .setRegulationTerminal(t3wt.getLeg3().getTerminal())
+                .setTargetV(10.0);
+
+        parameters.setTransformerVoltageControlOn(true);
+
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        assertTrue(!result.isOk()); //FIXME
```
annetill, line 179: @geofjamg something is not working at all because the solver failed in that case. The RTC is located on leg2 and I am trying to regulate voltage on leg3. Note that the bus1 is PV through a generator. I am not sure to get what is happening.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -0,0 +1,442 @@
...
+
+        parameters.setTransformerVoltageControlOn(true);
+        t2wt.getRatioTapChanger()
+                .setTargetDeadband(0)
+                .setRegulating(true)
+                .setTapPosition(0)
+                .setRegulationTerminal(t2wt.getTerminal1())
+                .setTargetV(134.0);
+
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
```
annetill, line 111: @geofjamg it is not working: here we have a RTC on a T2WT that is trying to regulate on side 1. But we also have a generator that maintains voltage at 135.0kV. What I don't understand is: why we are in step zero but voltages are not equals to base test case. I have maybe a problem of initialization after the firsyt iteration of the outerloop.

annetill:
@geofjamg (and maybe @floriand-e2r ) : indeed it does not seem to work well for some cases. I have put some FIXME if you want to analyze my problems. The remote mode is not working and for sure I need a bigger network in order to have a regulating ratio tap changer not close to generation. Any idea ?

annetill (commented):
I request some help!

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -98,13 +102,14 @@ private static void createVoltageControlledBusEquations(LfBus controlledBus, Equ
             } else {
                 boolean deriveA1 = creationParameters.isPhaseControl() && branch.isPhaseController()
                     && branch.getDiscretePhaseControl().getMode() == DiscretePhaseControl.Mode.CONTROLLER;
+                boolean deriveR1 = creationParameters.isTransformerVoltageControl() && branch.isVoltageController();
```
geofjamg, line 105: Somewhere in this file we should throw an exception if we have an open branch with voltage control on because
  - does not make sens?
  - not supported by equations.

src/main/java/com/powsybl/openloadflow/equations/Equation.java
```diff
@@ -111,6 +111,9 @@ public Equation addTerms(List<EquationTerm> terms) {
 
     private static double getBusTargetV(LfBus bus) {
         Objects.requireNonNull(bus);
+        if (bus.isDiscreteVoltageControlled()) {
```
geofjamg, line 114: Probably to be done on lf network side but we should throw en exception in case of inconsistent target with transformer voltage control and generation voltage control on same bus.

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -85,6 +87,32 @@ private static LfBranchImpl createTransformer(TwoWindingsTransformer twt, LfBus
             piModel = new PiModelArray(models, ptc.getLowTapPosition(), ptc.getTapPosition());
         }
 
+        RatioTapChanger rtc = twt.getRatioTapChanger();
+        if (rtc != null && rtc.isRegulating()) {
```
geofjamg, line 91: In this code rtc override ptc, so we don't support the case where we have both an active rtc and ptc. So we should either throw an exception or select one (which one?) and warn the user.

src/main/java/com/powsybl/openloadflow/network/impl/LfLegBranch.java
```diff
@@ -74,16 +78,43 @@ public static LfLegBranch create(LfBus bus1, LfBus bus0, ThreeWindingsTransforme
                         .setA1(a1));
             }
             piModel = new PiModelArray(models, ptc.getLowTapPosition(), ptc.getTapPosition());
+        }
 
-        } else {
+        RatioTapChanger rtc = leg.getRatioTapChanger();
```
geofjamg, line 83: Same here

src/test/resources/n.json
```diff
@@ -35,7 +35,7 @@
     "num2" : 1,
     "r" : 1.846153846153846E-4,
     "x" : 0.007690091988585015,
-    "r1" : 1.0526315789473686
+    "r1" : 1.0526315789473684
```
geofjamg, line 38: why is it different?

src/test/resources/logback-test.xml
```diff
@@ -16,7 +16,7 @@
     <root level="info">
         <appender-ref ref="STDOUT" />
     </root>
-    <logger name="com.powsybl.openloadflow" level="DEBUG" additivity="false">
+    <logger name="com.powsybl.openloadflow" level="TRACE" additivity="false">
```
geofjamg, line 19: To revert before merging

src/test/resources/n.json
```diff
@@ -61,6 +61,6 @@
     "num2" : 3,
     "r" : 2.0999999999999998E-4,
     "x" : 0.017998774958313132,
-    "r1" : 1.0013337777777775
+    "r1" : 1.0013337777777778
```
geofjamg, line 64: why is it different?

geofjamg (changes requested):

src/test/resources/n.json
annetill: Because I have introduced `baseRatio` as an intermediate.

src/test/resources/n.json
annetill: Because I have introduced `baseRatio` as an intermediate.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
annetill: For regulating shunt later maybe?

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
annetill: I have tried to make some improvements and I have created a test but... I fall in powsybl-core exceptions `com.powsybl.iidm.network.ValidationException: 2 windings transformer 'T2wT': Only one regulating control enabled is allowed`

src/main/java/com/powsybl/openloadflow/network/impl/LfLegBranch.java
annetill: I have tried to make some improvements and I have created a test but... I fall in powsybl-core exceptions `com.powsybl.iidm.network.ValidationException: 2 windings transformer 'T2wT': Only one regulating control enabled is allowed`

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -92,7 +92,7 @@ void voltageControlT2wtTest() {
         LoadFlowResult result = loadFlowRunner.run(network, parameters);
         assertTrue(result.isOk());
         assertVoltageEquals(134.281, bus2);
-        assertVoltageEquals(34.433, t2wt.getTerminal2().getBusView().getBus());
+        assertVoltageEquals(34.433, t2wt.getTerminal2().getBusView().getBus()); //FIXME: should be 34.427
```
annetill, line 96: @geofjamg here the FIXME: result is strange.

src/main/java/com/powsybl/openloadflow/equations/EquationType.java
```diff
@@ -18,7 +18,8 @@
     BRANCH_I("i", SubjectType.BRANCH),
     ZERO_Q("z_q", SubjectType.BUS),
     ZERO_V("z_v", SubjectType.BRANCH),
-    ZERO_PHI("z_\u03C6", SubjectType.BRANCH);
+    ZERO_PHI("z_\u03C6", SubjectType.BRANCH),
+    ZERO_RHO1("z_\u03B1", SubjectType.BRANCH);
```
floriand-e2r, line 22: \u03C1 for rho

src/main/java/com/powsybl/openloadflow/equations/VariableType.java
```diff
@@ -13,6 +13,7 @@
     BUS_V("v"),
     BUS_PHI("\u03C6"),
     BRANCH_ALPHA1("\u03B1" + "1"),
+    BRANCH_RHO1("rho" + "1"),
```
floriand-e2r, line 16: \u03C1 instead of rho

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -227,26 +232,51 @@ private static void createBranchActivePowerTargetEquation(LfBranch branch, Discr
         }
     }
 
+    private static void createDiscreteVoltageControlEquation(LfBus bus,  VariableSet variableSet, EquationSystem equationSystem) {
+        if (bus.isDiscreteVoltageControlled()) {
+            equationSystem.createEquation(bus.getNum(), EquationType.BUS_V).addTerm(new BusVoltageEquationTerm(bus, variableSet));
+        }
+        if (bus.isDiscreteVoltageControlled() && bus.getDiscreteVoltageControl().getControllers().size() > 1) {
```
floriand-e2r, line 239: you could insert that if in the if above. Or we could add a isDiscreteVoltageSharedControlled method?

floriand-e2r (changes requested):

floriand-e2r (approved):

src/main/java/com/powsybl/openloadflow/equations/Equation.java
```diff
@@ -147,14 +150,20 @@ private static double getBranchTarget(LfBranch branch, DiscretePhaseControl.Unit
...
+        LfBus firstControllerBus = network.getBus(data.getFirstControllerElementNum());
         double c = data.getC();
         return c * (controllerBus.getLoadTargetQ() - controllerBus.getGenerationTargetQ())
                 - firstControllerBus.getLoadTargetQ() - firstControllerBus.getGenerationTargetQ();
     }
 
+    private static double getRho1DistributionTarget(LfNetwork network, int num, DistributionData data) {
+        LfBranch controllerBranch = network.getBranch(num);
+        LfBranch firstControllerBranch = network.getBranch(data.getFirstControllerElementNum());
+        return controllerBranch.getPiModel().getR1() - firstControllerBranch.getPiModel().getR1();
```
geofjamg, line 164: As far as I understand we keep the gap between rhos constant during regulation?
Maybe just add a comment somewhere?

geofjamg (approved):

Issue #129, opened 2020-07-16
Fast security analysis


geofjamg:
We need to implement a security analysis much faster than default one provided by `com.powsybl.security.SecurityAnalysisImpl` which is an external loop over `com.powsybl.loadflow.LoadFlow` run.

- A security analysis implement `com.powsybl.security.SecurityAnalysis`
- In order to be fast here is some design consideration:
  - A single LfNetwork instance has to be built for all contingency simulation. This one should include all couplers necessary to simulate contingency list.
  - A pre-analysis has to be done on IIDM network (without any modification of it and without any variant creation).
    - `ContingencyTopologyTraverser` could be helpful to identify for each of the contingency the branches and couplers that could be opened to simulate the contingency (fault propagation analysis).
    - A connected component calculation has to be done also for each contingency to check if connected component is split or not byt the equipments removal. As this calculation is a large part of total calculation time we will try first a naive implementation of connected components (like https://jgrapht.org/javadoc-1.4.0/org/jgrapht/alg/connectivity/ConnectivityInspector.html) then an optimized one (like https://dl.acm.org/doi/10.1145/322234.322235).
  - From previous analysis we are able to build a LfNetwork with all necessary bus/branch/coupler and also for each of the contingency a list of bus/branch/coupler to remove in order to simulate the contingency taking into account the fault propagation.
  - Core of security analysis is implemented with an `OuterLoop` instance.
    - This outerloop is executed one time for base analysis and then one time for each contingency of the list.
    - Base iteration is done using full equation system.
    - For each contingency iteration of this outer loop:
       - At beginning of the loop, equation system is updated to remove all equations related to bus/branch/coupler that should be removed for the current contingency.
      - At end of the loop security limits are verified and stored in the  `OuterLoop` instance.
      - Equation system is restored in its original state (full).
    - Each contingency analysis start from base calculation state using a NR warm start.

Some other performance ideas:
 - https://github.com/powsybl/powsybl-open-loadflow/pull/77 is a prerequisite.
 - Equations evaluation has to be optimized to lazy evaluate (because actually lots of equation are not active and don't need to be evaluated at each NR iteration).
 - Equations could be evaluated using GPU (vectorial calculation).
 - In order to avoid recreation of Jacobian structure after any equation add/removal/activation we could imagine a system of mutually exclusive equation pair than could share the same matrix row (Q and V for a generator bus for instance). So we could be able to only have one Jacobian structure for all NR iterations.

[...]

PR #135, opened 2020-08-02
Beta version of security analysis

geofjamg:

**Other information**:


floriand-e2r:
Graph calculations are needed in this PR to calculate the connected components; for each contingency we check if connected component is split or not by the equipments removal. As for each contingency we remove only a few edges from the original graph, we might optimize these calculations by implementing a dynamic graph algorithm. A dynamic graph algorithm is an algorithm which is maintaining a result after each graph modification (here edge deletion) in order to avoid unnecessary computations. 

The algorithm suggested in the related issue #129 is Even-Shiloach'81: complexity of graph update O(n) / of requests O(1), where n is the number of vertices. It has been implemented in previous commit. It is based on 2 processes working simultanously after each edge deletion, the first process which halts stops the other.
  - a classical process A detects if a new connected component is created
  - a more complex process B detects if the connected components remain intact

Drawbacks:
  - only supports edge deletion, hence to deal with contingencies one after the other we cannot use edge insertion to undo the deletion. In current implementation I saved the results of the original graph and cloned them when a new contingency is considered - this might be improved though.
  - when an edge is deleted and it doesn't create a new connected component, all the changes done by process B have to be tediously undone (I still need to implement that, thus so far it does not support several edge deletions in a row)

A few other algorithms exist for this matter:
* Frederickson'84: complexity of graph update 0(sqrt(m)) / of requests O(1) (m: number of edges), which is based on a minimum spanning tree maintained through the changes of the graph
* Eppstein et al'97: complexity of graph update 0(sqrt(n)) / of requests O(1), which is based on a "sparsification" of previous algorithm; we might want to avoid the introduced complexity as this is of interest for dense graphs only
* and others: since then more complex algorithms have been introduced which have better update complexity, but the request cost is not in O(1), and most of them are randomized algorithm. For instance, the (deterministic) algorithm Holm et al'01 has an update cost of O(log²n), and each request has O(log n/log log n) complexity.

I'm wondering if we need to go towards those more complex algorithms?

Note that jgrapht ConnectivityInspector now supports dynamic graphs... nonetheless, in current version, when a edge is removed all the calculations are reinitialized (with following comment: "for now invalidate cached results, in the future need to amend them")

floriand-e2r:
To optimize these algorithms, we can use a bridge detection pre-calculation, which will be useful if many contingencies are tested (a bridge is an edge which creates a new connected component if removed). Indeed, that allows not to compute the connected component if we know that the edge removed is not a bridge. AND if only one edge is removed (no fault propagation...). Computing bridges in a graph is a classical problem (see e.g. Hopcroft & Tarjan's algorithm implemented in in jgrapht BiconnectivityInspector) and is done in a single graph scan. Nonetheless I don't know of any algorithm for dynamic graphs supporting edge deletion, so it's only for the first edge deletion.

This is of interest if we're dealing with graphs which don't have many bridges. In the network I used for testing, there is about 25% of bridges. Note that among those bridges there are only 15% which produce a new connected component which is bigger than just one node.

floriand-e2r:
In BridgesTest I used algorithms computing connected components to find bridges in a basic network. Each branch of the network is deleted and I test if the two buses of the branch are still in the same connected component after that deletion. It doesn't correspond exactly to the use case but it already gives an idea of the algorithms performance. On a more complex network of 10000 branches the results are the following on my laptop:
* Naive implementation through jgrapht ConnectivityInspector (calculation done from scratch after each deletion): ~40s
* Connectivity calculation based on graphstream library: ~15s
* Connectivity calculation based on Even-Shiloach algorithm: ~10s

With that same network, the bridges calculation is done in about 5ms. Therefore, the previous execution times are divided
* by about 4 if the calculation is skipped for non-bridges edge deletion (about 25% of bridges in that network)
* by about 20 if the calculation is also skipped for bridges which produce a new connected component of size 1 (in that test network, 5% of the edges are bridges producing a new connected component of size greater than 1)

annetill:
Hi @geofjamg ! In order to test @floriand-e2r's work on graphs, we are going to work on connexity analysis after a contingency. In your present PR, we will improve the branches and nodes building in considering only switches we have between 2 bus bar sections and droping the two switches we met in each side of a branch when we can. @floriand-e2r is going to try to use his work to increase performances of the whole connexity analysis. This work could lead to improve the slack bus selection that could fail if the slack bus become out of load. 
I am going to perform some functional tests and to try to solve small issues.
I have in mind that we all have to work on:
- How to save the initial state and how to retrieve it after a post contingency LF ; for the moment, only angle and voltage are saved but we could need tap changer position, shunt section number, bus status (PV vs PQ), etc.
- How to detect limits violation on a LfNetwork after each unitary LF ; to begin, the solution is to implement permant limit mangement with a simple ````getPermanentLimit()```` of a LfBranch and a method that would say if the violation occurs or not after a LF calculation using the state variable. This would be a really interesting use case to then validate all we are doing in powsybl-core on operational limits (https://github.com/powsybl/powsybl-core/pull/1359).

src/test/java/com/powsybl/openloadflow/OpenSecurityAnalysisTest.java
```diff
@@ -0,0 +1,213 @@
...
+        OpenSecurityAnalysis securityAnalysis = new OpenSecurityAnalysis(network, new DefaultLimitViolationDetector(),
+            new LimitViolationFilter(), new DenseMatrixFactory(), () -> new NaiveGraphDecrementalConnectivity<>(LfBus::getNum));
+        SecurityAnalysisResult result = securityAnalysis.runSync(saParameters, contingenciesProvider);
+        assertTrue(result.getPreContingencyResult().isComputationOk());
+        assertEquals(0, result.getPreContingencyResult().getLimitViolations().size());
+        assertEquals(2, result.getPostContingencyResults().size());
+        assertTrue(result.getPostContingencyResults().get(0).getLimitViolationsResult().isComputationOk());
+        assertEquals(2, result.getPostContingencyResults().get(0).getLimitViolationsResult().getLimitViolations().size());
+        assertTrue(result.getPostContingencyResults().get(1).getLimitViolationsResult().isComputationOk());
+        assertEquals(4, result.getPostContingencyResults().get(1).getLimitViolationsResult().getLimitViolations().size());
```
annetill, line 211: I could not be more than 2 :-) To be corrected. ````L2```` seems to be still in the network but should be out of load.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    /**
+     * Detect violation limits on one branch and add them to the given list
+     * @param branch branch of interest
+     * @param violations list on which the violation limits encountered are added
+     */
+    private void detectBranchViolations(LfBranch branch, List<LimitViolation> violations) {
+        // Detect violation on the left
+        detectBranchViolationSide(branch.getId(), Branch.Side.ONE, branch.getBus1(), branch.getI1(), branch.getPermanentLimit1(), violations);
```
annetill, line 196: I am not sure that we need this refactoring. In a first version, we just take into account permanent limit for current. We will then support temporary limits for current. After that, we will try to do the same for apparent power limit. I think that it is better to imagine the refactoring with the whole picture or do it at the end.

annetill: ````detectBranchViolationSide```` indeed I am talking about this function.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
```diff
@@ -0,0 +1,474 @@
...
+    private void setBusState(LfBus bus, BusState busState) {
+        bus.setV(busState.v);
+        bus.setAngle(busState.angle);
+        bus.setLoadTargetP(busState.loadTargetP);
+        bus.getGenerators().forEach(g -> {
+            g.setTargetP(busState.generatorsTargetP.get(g.getId()));
+        });
+    }
+
+    private void switchPqPv(LfBus bus, AcloadFlowEngine engine) {
```
annetill, line 299: I prefer not t introduce this function that is confusing for me because already exists.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
```diff
@@ -0,0 +1,474 @@
...
+     * @return the map of the states of given buses, indexed by the bus itself
+     */
+    private Map<LfBus, BusState> getBusStates(List<LfBus> buses, AcloadFlowEngine engine) {
+        Map<LfBus, BusState> result = new HashMap<>(buses.size());
+        buses.forEach(b -> {
+            Map<String, Double> generatorsTargetP = b.getGenerators().stream().collect(
+                Collectors.toMap(LfGenerator::getId, LfGenerator::getTargetP));
+            BusState busState = new BusState(b.getV(), b.getAngle(), b.getLoadTargetP(), generatorsTargetP);
+            result.put(b, busState);
+            switchPqPv(b, engine);
```
annetill, line 273: I think that I made a mistake here: we don't have to call this function ````switchPqPv```` at this place. We have to save the information of which buses are PQ now and where PV at the beginning of the LF and the associated targetQ.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
```diff
@@ -0,0 +1,474 @@
...
+        return new SecurityAnalysisResult(preContingencyResult, postContingencyResults);
+    }
+
+    /**
+     * Get the map of the states of given buses, indexed by the bus itself
+     * @param buses the bus for which the state is returned
+     * @param engine
+     * @return the map of the states of given buses, indexed by the bus itself
+     */
+    private Map<LfBus, BusState> getBusStates(List<LfBus> buses, AcloadFlowEngine engine) {
```
annetill, line 266: I prefer ````saveBusState````.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
```diff
@@ -0,0 +1,474 @@
...
+        });
+        return result;
+    }
+
+    /**
+     * Set the bus states based on the given map of states
+     * @param busStates the map containgin the bus states, indexed by buses
+     * @param engine
+     */
+    private void restoreBusStates(Map<LfBus, BusState> busStates, AcloadFlowEngine engine) {
```
annetill, line 283: I prefer ````restoreBusState````.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
```diff
@@ -0,0 +1,474 @@
...
+        busStates.forEach((b, state) -> {
+            setBusState(b, state);
+            switchPqPv(b, engine);
```
annetill, line 286: So here is wrong indeed. You have to call ````switchPqPv```` only for buses that have become PQ and where PV at pre-contingency state.

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -177,12 +177,16 @@ public double getI2() {
 
     @Override
     public double getPermanentLimit1() {
-        return branch.getCurrentLimits1() != null ? branch.getCurrentLimits1().getPermanentLimit() * getBus1().getNominalV() / PerUnit.SB : Double.NaN;
+        return branch.getCurrentLimits1() != null && getBus1() != null
```
annetill, line 180: I put that at branch violation detections level. See my previous commit. I did that in order to avoid calling permanent limit before. What is the best practice ?

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
```diff
@@ -171,36 +172,42 @@ SecurityAnalysisResult runSync(SecurityAnalysisParameters securityAnalysisParame
...
 
+    /**
+     * Detect violation limits on one branch and add them to the given list
+     * @param branch branch of interest
+     * @param violations list on which the violation limits encountered are added
+     */
     private void detectBranchViolations(LfBranch branch, List<LimitViolation> violations) {
-        // detect violation limits on a branch
-        double scale = 1;
-        if (branch.getBus1() != null && branch.getI1() > branch.getPermanentLimit1()) {
```
annetill: You miss here to keep the open branches treatement.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
```diff
@@ -171,36 +172,42 @@ SecurityAnalysisResult runSync(SecurityAnalysisParameters securityAnalysisParame
...
+        // Detect violation on the left
+        detectBranchViolationSide(branch.getId(), Branch.Side.ONE, branch.getBus1(), branch.getI1(), branch.getPermanentLimit1(), violations);
+
+        // Detect violation on the right
+        detectBranchViolationSide(branch.getId(), Branch.Side.TWO, branch.getBus2(), branch.getI2(), branch.getPermanentLimit2(), violations);
+    }
+
+    private void detectBranchViolationSide(String branchId, Branch.Side branchSide, LfBus busOfSide, double iOfSide,
+                                           double permanentLimitOfSide, List<LimitViolation> violations) {
+        if (busOfSide != null && iOfSide > permanentLimitOfSide) {
```
annetill, line 204: Sorry, you made it here. So why do you need to increase the permanent limit getters ?

annetill (changes requested):
As a general comment: for me you have introduce to many functions. I understand that it is to avoid duplication, but it makes the code not to easy to understand. We have to review together how to save PQ buses that were PV in the pre-contingency state.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
floriand-e2r: yes, this seemed confusing to me too: it's probably because it's doing something wrong... as you mentioned in a later comment.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
floriand-e2r: We do not need this refactoring, sure. It was just meant to avoid code duplication and to improve readability, but as you disagree on that point I'll revert it.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
floriand-e2r: When returning a collection of objects, the plural is usually preferred. About save or get, is this method saving or getting the states...?

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
floriand-e2r: The soon-to-be reverted refactoring needed to also check it here.

src/main/java/com/powsybl/openloadflow/OpenSecurityAnalysis.java
floriand-e2r: To avoid NullPointerException before calling this method.

annetill:
@geofjamg @floriand-e2r this PR is for me ready to be merged. @geofjamg do you agree?

PR #140, opened 2020-08-25
Make the most meshed bus selector deterministic

annetill:

I have performed a test on a classical merged network and on a network merged through the merging view. In some cases, the slack bus selection is not the same in these 2 networks that are functionally identical. 

The selection algorihtm becomes deterministic. 

I have added a sort by bus id before max branches selection.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/MostMeshedSlackBusSelector.java
+++ b/src/main/java/com/powsybl/openloadflow/network/MostMeshedSlackBusSelector.java
@@ -28,6 +28,7 @@ public class MostMeshedSlackBusSelector implements SlackBusSelector {
         // select non fictitious and most meshed bus among buses with highest nominal voltage
         return buses.stream()
                 .filter(bus -> !bus.isFictitious() && bus.getNominalV() == maxNominalV)
+                .sorted(Comparator.comparing(bus -> bus.getId()))
                 .max(Comparator.comparingInt(bus -> bus.getBranches().size()))
                 .orElseThrow(AssertionError::new);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/network/MostMeshedSlackBusSelector.java
```diff
@@ -28,6 +28,7 @@ public LfBus select(List<LfBus> buses) {
         // select non fictitious and most meshed bus among buses with highest nominal voltage
         return buses.stream()
                 .filter(bus -> !bus.isFictitious() && bus.getNominalV() == maxNominalV)
+                .sorted(Comparator.comparing(bus -> bus.getId()))
```
floriand-e2r, line 31: You could be more efficient by avoiding the sort; just take the max size and if equal size the max by id. This can be done by replacing next line by 
`.max(Comparator.comparingInt((LfBus bus) -> bus.getBranches().size()).thenComparing(LfBus::getId))`

floriand-e2r: Attention, you might prefer to have the biggest size and the smallest id; then you need to reverse the 2nd comparator:
`.thenComparing(Comparator.comparing(LfBus::getId).reversed())` instead of `.thenComparing(LfBus::getId)`

annetill: Thanks for your advice, that I understand well. But I have tried both of your sugesstions and for both, I have unit tests failures. In my first proposal, I can keep the same unit tests. I also have a code smell that makes sonar could failed. All of that is not so clear for me...

geofjamg: It is not necessarly a problem to adapt unit tests because of slack bus selection change.

geofjamg (approved):

PR #144, opened 2020-09-03
Slack distribution on conform loads

jeanChristopheGuillotin:

**Other information**:


src/main/java/com/powsybl/openloadflow/ac/DistributedSlackOnLoadOuterLoop.java
```diff
@@ -58,7 +58,14 @@ protected double run(List<ParticipatingElement<LfBus>> participatingElements, in
 
             double targetP = bus.getLoadTargetP();
 
-            double newTargetP = targetP - remainingMismatch * factor;
+            double fixedActivePowed = bus.getFixedActivePower();
+            double variableActivePower = bus.getVariableActivePower();
+
+            double newVariableActivePower = variableActivePower - remainingMismatch * factor;
```
annetill, line 64: You must change the way the factors are computed.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -107,6 +111,16 @@ public double getTargetQ() {
         return getGenerationTargetQ() - getLoadTargetQ();
     }
 
+    @Override
+    public double getFixedActivePower() {
+        return 0;
```
annetill, line 116: You must return the attribute.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -107,6 +111,16 @@ public double getTargetQ() {
...
+    }
+
+    @Override
+    public double getVariableActivePower() {
+        return 0;
```
annetill, line 121: You must return the attribute.

annetill (changes requested):
You must change the class ````LfAbstractBus```` and build a unit test to improve this dev?

annetill:
I will review as soon as possible. Can you see with Marco for the next issue that consists in keeping the power factor constant ? Thanks ! You can start from master branch, we will merge this PR later after review process.

src/main/java/com/powsybl/openloadflow/ac/DistributedSlackOnLoadOuterLoop.java
```diff
@@ -74,6 +76,7 @@ protected double run(List<ParticipatingElement<LfBus>> participatingElements, in
                 }
 
                 bus.setLoadTargetP(newTargetP);
```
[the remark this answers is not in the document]
floriand-e2r: Hello @annetill, in fact I think it is better to have the absolute value of the load fixed part. Indeed, here you need to update the ratio each time you change the load target. If you're dealing with the fixed part in absolute value, you don't have that problem

floriand-e2r (approved):
I don't see the benefit to use the per unit here, but it seems that it's the usual practice?

Issue #84, opened 2020-05-12
Phase control could be improved if needed


annetill:
* **Do you want to request a *feature* or report a *bug*?**

It is more a feature.

* **What is the current behavior?**

We have recently merged the control of active power by phase tap changer. This control can be located at side 1 or at side 2 of the branch, where the tap changer is. As a beta version, we only support two windings transformers. Furthermore, the active power target is often given with a deadband. The open loadflow logs if the active power at regulation terminal is in or out the deadband.  

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

There is no expected behavior. Phase control in this open loadflow is a minimal viable feature and could be improved. I am writing this issue to let some notes of what could be missing.

* **What is the motivation / use case for changing the behavior?**

- The deadband could be taken into account to select the rounded step after the continuious resolution. In fact, we have to choose between 2 steps. In the current implementation, we choose the step which alpha is the closest of the continuious value. 
- The regulating terminal could be remote ;
- We don't support phase control on phase tap changer belonging to three windings transformers ;
- Maybe a connexity computation to evaluate the efficiency of the phase control ;
- Phase control for active power flow (DC mode) ;
- Phase control for phase tap changer which control mode is CURRENT_LIMITER (https://www.powsybl.org/docs/iidm/model/phaseTapChanger.html). The continuous problem is solved only for tap changers that are out of its regulation value.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

If other features are missing, please leave a message in this discussion.

annetill:
We have started to work on current limiter mode in a branch: https://github.com/powsybl/powsybl-open-loadflow/tree/current-limiter-mode
To be finished.

geofjamg:
Solved by #340

PR #153, opened 2020-09-11
Phase control: remote active power mode

tadam50:

#84 

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfLegBranch.java
```diff
@@ -46,25 +49,73 @@ public static LfLegBranch create(LfBus bus1, LfBus bus0, ThreeWindingsTransforme
...
         double nominalV2 = twt.getRatedU0();
         double zb = nominalV2 * nominalV2 / PerUnit.SB;
         PhaseTapChanger ptc = leg.getPhaseTapChanger();
         if (ptc != null
                 && ptc.isRegulating()
                 && ptc.getRegulationMode() != PhaseTapChanger.RegulationMode.FIXED_TAP) {
-            LOGGER.error("3 windings transformer '{}' has a regulating phase tap changer which is not yet supported", twt.getId());
+            PhaseTapChanger.RegulationMode regulationMode = ptc.getRegulationMode();
+            PhaseControl.ControlledSide controlledSide = null;
+            if (ptc.getRegulationTerminal() == twt.getLeg1().getTerminal()) {
```
annetill, line 65: I think the code only allows to regulate locally. It means that a phase tap changer located in a dedicated leg can only regulated at its local terminal. I think that next week I will increase the regulation capability to the 2 other terminals of the transformer (but I don't know if it is easy of not ;-)).

src/main/java/com/powsybl/openloadflow/network/impl/LfLegBranch.java
```diff
@@ -46,25 +49,73 @@ public static LfLegBranch create(LfBus bus1, LfBus bus0, ThreeWindingsTransforme
...
+                controlledSide = PhaseControl.ControlledSide.ONE;
+            } else if (ptc.getRegulationTerminal() == twt.getLeg2().getTerminal()) {
+                controlledSide = PhaseControl.ControlledSide.TWO;
+            } else if (ptc.getRegulationTerminal() == twt.getLeg3().getTerminal()) {
+                controlledSide = PhaseControl.ControlledSide.THREE;
+            } else {
+                LOGGER.error("3 windings transformer '{}' has a regulating phase tap changer which is not yet supported", twt.getId());
```
annetill, line 72: I think that here the best solution could be to consider the phase tap changer as ````FIXED_TAP````. If we do that choice, we have to modify ````LfBranchImpl```` too.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowPhaseShifterTest.java
```diff
@@ -62,7 +54,9 @@ void setUp() {
     }
 
     @Test
-    void baseCaseTest() {
+    void baseCase2WTTest() {
```
annetill, line 57: Can you change for ````baseCaseT2wtTest()```` ? I have forgotten.

annetill (commented):
Just some comments, good beginning!

floriand-e2r:
@tadam50 After discussing with @annetill the PhaseControl design, I refactored a bit your proposition:
- both controller and controlee now hold a reference to the corresponding phase control; indeed we often need only the control of the controlee, and going each time through the controlled branch seemed tedious to me 
- the phase control is now created at once (was previously created in the controller branch and after updated)
- even though phase control and voltage control share the controller/controlee principle, they don't have any common methods/attributes, hence I don't think inheritance from a common class helped

annetill:
@floriand-e2r I think that the limiter mode is more complex than expected. So, what do you think about removing it from this PR and open a new PR for only that feature ? So, this PR will be mergeable ans it will be easier to work on discrete voltage control (#127 ).

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -185,34 +163,31 @@ public double getPermanentLimit2() {
...
-
     @Override
     public void updateState() {
         branch.getTerminal1().setP(p1.eval() * PerUnit.SB);
         branch.getTerminal1().setQ(q1.eval() * PerUnit.SB);
         branch.getTerminal2().setP(p2.eval() * PerUnit.SB);
         branch.getTerminal2().setQ(q2.eval() * PerUnit.SB);
 
-        if (phaseControl != null) { // it means there is a regulating phase tap changer
+        if (isPhaseController()) { // it means there is a regulating phase tap changer
```
annetill, line 173: @floriand-e2r Do not forget to add condition && phaseControl.getMode() == PhaseControl.Mode.OFF

src/main/java/com/powsybl/openloadflow/network/impl/LfLegBranch.java
```diff
@@ -133,10 +155,20 @@ public void updateState() {
         leg.getTerminal().setP(p.eval() * PerUnit.SB);
         leg.getTerminal().setQ(q.eval() * PerUnit.SB);
 
-        if (phaseControl != null) {
+        if (isPhaseController()) {
```
annetill, line 158: @floriand-e2r Do not forget to add condition && phaseControl.getMode() == PhaseControl.Mode.OFF

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
```diff
@@ -33,47 +33,72 @@ public String getType() {
...
-                    a1.setActive(false);
-
-                    // de-activate phase control equation
-                    Equation t = context.getEquationSystem().createEquation(branch.getNum(), EquationType.BRANCH_P);
-                    t.setActive(false);
+        switch (context.getIteration()) {
+            case 0:
+                return checkAtFirstIteration(context);
+            case 1:
+                return checkAtSecondIteration(context);
```
floriand-e2r, line 41: I'm not so happy with those names I gave... any suggestions?

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
```diff
@@ -33,47 +33,72 @@ public String getType() {
 
     @Override
     public OuterLoopStatus check(OuterLoopContext context) {
```
floriand-e2r, line 35: In this outerloop we do more than just check in this method: we also switch off/on phase control. Shouldn't it be done in another method, or is it just a naming issue?

Issue #147, opened 2020-09-07
Open loadflow parameters management


annetill:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

The parameters specific to the open loadflow are set in the Java main program only.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

The parameters specific to the open loadflow could be set in the config.yml file in a specific section ````open-loadflow-default-parameters````. The first developpement is to transfert the parameters in the config file with the same names.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

The parameters of loadflow engines are common and have to be put higher in the LoadflowParameters class. See below some parameters of hades2:

````hades2-default-parameters:
  balanceType: PROPORTIONAL_TO_LOAD
  computedConnectedComponentNumber: 0
  reactiveCapabilityCurveWithMoreThanThreePoints: INFINITE_DIAGRAM
  withMinimumReactance: true
  minimumReactancePerUnit: 0.0007
  anglePerte: false
  remoteVoltageGenerators: true
  dcMode: false
  hvdcAcEmulation: false
  dcUseTransformerRatio: true
  debugMode: ETATS_INTERMEDIAIRES
````

annetill:
Hi! I notice that we miss one parameter: the slack bus selector:     
````private SlackBusSelector slackBusSelector = new MostMeshedSlackBusSelector();````
We should be able to select the slack bus option from the configuration file.

PR #158, opened 2020-09-21
Add SlackBusSelector to config file as open loadflow parameter

tadam50:

#147

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/OpenLoadFlowParametersTest.java
+++ b/src/test/java/com/powsybl/openloadflow/OpenLoadFlowParametersTest.java
@@ -11,6 +11,8 @@ import com.google.common.jimfs.Jimfs;
 import com.powsybl.commons.config.InMemoryPlatformConfig;
 import com.powsybl.commons.config.MapModuleConfig;
 import com.powsybl.loadflow.LoadFlowParameters;
+import com.powsybl.openloadflow.network.FirstSlackBusSelector;
+import com.powsybl.openloadflow.network.SlackBusSelector;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -31,10 +33,16 @@ public class OpenLoadFlowParametersTest {
 
     private FileSystem fileSystem;
 
+    private MapModuleConfig lfModuleConfig;
+
     @Before
     public void setUp() {
         fileSystem = Jimfs.newFileSystem(Configuration.unix());
         platformConfig = new InMemoryPlatformConfig(fileSystem);
+
+        lfModuleConfig = platformConfig.createModuleConfig("load-flow-default-parameters");
+        lfModuleConfig.setStringProperty("voltageInitMode", LoadFlowParameters.VoltageInitMode.DC_VALUES.toString());
+        lfModuleConfig.setStringProperty("transformerVoltageControlOn", Boolean.toString(true));
     }
 
     @After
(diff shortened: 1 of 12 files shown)


floriand-e2r (changes requested):
We cannot choose a `NameSlackBusSelector` (nor a `NetworkSlackBusSelector`) with the current config file. How could we get the busId from the config file? with another parameter?

annetill:
For the ````networkSlackBusSelector````, it seems that we have to use a parameter from ````LoadflowParameter```` called ````readSlackBus````. Do you agree @floriand-e2r ?

floriand-e2r:
Yes, PR [#1416](https://github.com/powsybl/powsybl-core/pull/1416) in powsybl-core introduced that parameter:

> - isReadSlackBus: the extension can be used by the power flow to select the slack bus

That corresponds more or less to what is done in NetworkSlackBusSelector.

geofjamg:
> Yes, PR [#1416](https://github.com/powsybl/powsybl-core/pull/1416) in powsybl-core introduced that parameter:
> 
> > * isReadSlackBus: the extension can be used by the power flow to select the slack bus
> 
> That corresponds more or less to what is done in `NetworkSlackBusSelector`.

This is a little bit more tricky. isReadSlackBus is a kind of addition to `SlackBusSelector` that pre-filter slack bus candidates before going to slack bus selection (this is the purpose of  `NetworkSlackBusSelector` that also take a fallback). So this is complementary.
Here is the code:
```java
    private static SlackBusSelector getSlackBusSelector(Network network, LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
        return parameters.isReadSlackBus() ? new NetworkSlackBusSelector(network, parametersExt.getSlackBusSelector())
                                           : parametersExt.getSlackBusSelector();
    }
```

geofjamg:
I think that to because PlatformConfig has very simple capabilities (this is a flat config model), the only way to deal with `SlackBusSelector` json is to use several variables in case of additional data needed by some implementations.

```json
...
slackBusSelectorType: "Named",
nameSlackBusSelectorBusId: "toto"
...
```

We can maybe introduce a `SlackBusSelectorParametersReader` to keep the generic design (everyone is allowed to provide its own impl) that will be responsible to build a `SlackBusSelector` instance using parameters.
 
```
interface SlackBusSelectorParametersReader {
    String getName();
    SlackBusSelector read(ModuleConfig moduleConfig);
}
```
And `SlackBusSelectorParametersReader` implementations will be discovered using service loader and selected using slackBusSelectorType in the json.

src/test/java/com/powsybl/openloadflow/OpenLoadFlowParametersTest.java
```diff
@@ -69,22 +77,70 @@ public void testConfig() {
...
+        Path cfgDir = Files.createDirectory(fileSystem.getPath("config"));
+        Path cfgFile = cfgDir.resolve("configNameSlackBusSelector.yml");
+
+        Files.copy(getClass().getResourceAsStream("/configNameSlackBusSelector.yml"), cfgFile);
+        PlatformConfig platformConfig = new PlatformConfig(new YamlModuleConfigRepository(cfgFile), cfgDir);
+
+        LoadFlowParameters parameters = LoadFlowParameters.load(platformConfig);
+        OpenLoadFlowParameters olfParameters = parameters.getExtension(OpenLoadFlowParameters.class);
+        assertEquals(NameSlackBusSelector.class, olfParameters.getSlackBusSelector().getClass());
+    }
```
floriand-e2r, line 145: Could you just add an assert to check if the busId is read correctly?

floriand-e2r (changes requested):

src/test/java/com/powsybl/openloadflow/OpenLoadFlowParametersTest.java
tadam50: Good catch

floriand-e2r (approved):

src/test/java/com/powsybl/openloadflow/OpenLoadFlowParametersTest.java
floriand-e2r: I would have tested an existing bus (like VLHV1_0 for instance) instead of testing the exception raised by an unknown bus, but that's ok!

floriand-e2r:
Sorry for the last pushes... I just tried to change the code in order for Sonar to have the correct coverage - without success!

annetill (approved):
The documentation has to be completed, see https://github.com/powsybl/powsybl.github.io/pull/143 !

PR #168, opened 2020-11-10
Refactoring voltage control

floriand-e2r:

No

Refactoring voltage control to have a structure similar to PhaseDiscreteControl / VoltageDiscreteControl, that is, a "control" object which makes the link between the controller and the controllee.


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
```diff
@@ -20,12 +22,71 @@
...
+        Bus connectedBus = getBus(injection.getTerminal(), breakers);
+        if (controlledBus == null || connectedBus == null) {
+            return false;
+        }
+        String controlledBusId = controlledBus.getId();
+        String connectedBusId = connectedBus.getId();
+        return !Objects.equals(controlledBusId, connectedBusId);
+    }
+
+    private static double getTargetVForDeactivatedRemoteControl(Injection<?> injection, Terminal regulatingTerminal, double previousTargetV) {
```
annetill, line 62: Indeed just a `getLocalTargetV` method should be enough.

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineGenerator.java
```diff
@@ -32,7 +33,7 @@ public String getId() {
 
     @Override
     public boolean hasVoltageControl() {
-        return danglingLine.getGeneration().isVoltageRegulationOn();
+        return danglingLine.getGeneration().isVoltageRegulationOn() && !deactivatedVoltageControl;
```
annetill, line 36: I don't understand here: do we want to involve parameters in that level ?

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -81,7 +84,7 @@ public String getId() {
 
     @Override
     public boolean hasVoltageControl() {
-        return generator.isVoltageRegulatorOn();
+        return generator.isVoltageRegulatorOn() && !deactivatedVoltageControl;
```
annetill, line 87: Same remark. What is this parameter?

src/main/java/com/powsybl/openloadflow/network/impl/LfStaticVarCompensatorImpl.java
```diff
@@ -72,7 +76,8 @@ public String getId() {
 
     @Override
     public boolean hasVoltageControl() {
-        return svc.getRegulationMode() == StaticVarCompensator.RegulationMode.VOLTAGE;
+        return svc.getRegulationMode() == StaticVarCompensator.RegulationMode.VOLTAGE
+                && !deactivatedVoltageControl;
```
annetill, line 80: Same remark.

src/main/java/com/powsybl/openloadflow/network/impl/LfVscConverterStationImpl.java
```diff
@@ -49,7 +50,7 @@ public String getId() {
 
     @Override
     public boolean hasVoltageControl() {
-        return station.isVoltageRegulatorOn();
+        return station.isVoltageRegulatorOn() && !deactivatedVoltageControl;
```
annetill, line 53: Same remark. What is this parameter ?

annetill (changes requested):
I have left the review of abstract lf bus and lf network loader impl for latter. I think that we have to review bases before in having a unique voltage control in case of remote shared control but split it in several voltage control if remote voltage control is desactivated. This solution is not clear enough to be prefer at the current one.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -111,106 +103,46 @@ public double getTargetQ() {
...
-                        .ifPresent(otherControllerBus -> {
-                            double otherTargetV = otherControllerBus.getTargetV() * controlledBus.getNominalV();
-                            if (FastMath.abs(otherTargetV - this.targetV) > TARGET_V_EPSILON) {
-                                LOGGER.error("Bus '{}' control voltage of bus '{}' which is already controlled by at least the bus '{}' with a different target voltage: {} (kept) and {} (ignored)",
-                                        getId(), controlledBus.getId(), otherControllerBus.getId(), otherTargetV, this.targetV);
-                                this.targetV = otherTargetV;
-                            }
-                        });
-            }
+    public void setVoltageControl(boolean voltageControl) {
```
annetill, line 131: You have to renamed ths function for sure. I thought that you can make it using the Mode OFF of the voltageControl ?

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -31,24 +31,26 @@
...
                 equationSystem.createEquation(bus.getNum(), EquationType.BUS_PHI).addTerm(new BusPhaseEquationTerm(bus, variableSet));
                 equationSystem.createEquation(bus.getNum(), EquationType.BUS_P).setActive(false);
             }
 
-            if (bus.hasVoltageControl()) {
-                // local voltage control
-                if (!(creationParameters.isVoltageRemoteControl() && bus.getControlledBus().isPresent()) && bus.getControllerBuses().isEmpty()) {
-                    equationSystem.createEquation(bus.getNum(), EquationType.BUS_V).addTerm(new BusVoltageEquationTerm(bus, variableSet));
-                }
+            boolean hasLocalControl = bus.isVoltageController() && bus.getVoltageControl().isVoltageControlLocal();
```
annetill, line 41: This function could be more clear.

annetill (changes requested):
Some small changes. After @geofjamg review, I will perform a functional validation as the feature is central.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -172,9 +172,9 @@ void nonSupportedVoltageControlT2wtTest() {
                 .setRegulationTerminal(network.getGenerator("GEN_1").getTerminal())
                 .setTargetV(33.0);
 
-        LoadFlowResult result = loadFlowRunner.run(network, parameters);
-        assertTrue(result.isOk());
-        assertVoltageEquals(135.0, bus1);
+        LoadFlowResult result = loadFlowRunner.run(network, parameters); //FIXME: result should be ok after merge
```
floriand-e2r, line 175: I don't get why the test does not succeed anymore. Max iteration is now reached with this refactoring

floriand-e2r: Got it! I'll correct this.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -41,7 +37,7 @@
 
     protected boolean voltageControlCapability = false;
```
geofjamg, line 38: But if the bus is attached to a VoltageControl object, it means is has the capability ? What is the pupose of this property?

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
geofjamg: No clear to me too what is the difference between mode of the VoltageControl and property voltageControlEnabled ? Is mode really usefull as mode is kind of result of individual voltage control enable status of all controllers?

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -119,106 +111,46 @@ public double getTargetQ() {
...
-                        });
-            }
+    public void setVoltageControl(boolean voltageControl) {
+        if (this.voltageControlEnabled && !voltageControl) {
+            voltageControlSwitchOffCount++;
         }
-
-        this.controlledBus = controlledBus;
-        controlledBus.addControllerBus(this);
+        this.voltageControlEnabled = voltageControl;
```
geofjamg, line 143: so the method should be setVoltageControlEntabled()?

geofjamg (changes requested):
For we have to clarify the different concepts and check if there is redundancy.
 - Who is responsible to store the structural config of the voltage control: bus or voltage control object?: in the code it looks like both with redundant infos.
 - Who is responsible to store the status of the voltage control? bus or voltage control object? also in the code it looks like both and I am not sure to understand the role of mode property.

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
```diff
@@ -53,27 +51,28 @@ public String getType() {
...
+
         // clean reactive power distribution equations
-        controlledBus.getControllerBuses().forEach(b -> equationSystem.removeEquation(b.getNum(), EquationType.ZERO_Q));
+        controllerBuses.forEach(b -> equationSystem.removeEquation(b.getNum(), EquationType.ZERO_Q));
 
         // controlled bus has a voltage equation only if one of the controller bus has voltage control on
-        List<LfBus> controllerBusesWithVoltageControlOn = controlledBus.getControllerBuses().stream()
-                .filter(LfBus::hasVoltageControl)
+        List<LfBus> controllerBusesWithVoltageControlOn = controllerBuses.stream()
+                .filter(LfBus::isVoltageController)
```
annetill, line 75: This is weird even if I understand why.

src/main/java/com/powsybl/openloadflow/equations/Equation.java
```diff
@@ -114,21 +114,14 @@ private static double getBusTargetV(LfBus bus) {
...
-                    .stream()
-                    .filter(LfBus::hasVoltageControl)
-                    .collect(Collectors.toList());
-            if (bus.hasVoltageControl()) {
-                controllerBuses.add(bus);
-            }
-            if (controllerBuses.isEmpty()) {
-                throw new IllegalStateException("None of the controller buses of bus '" + bus.getId()
-                        + "'has voltage control on");
+        if (bus.getVoltageControl().isPresent() && bus.isVoltageControlled()) {
```
annetill, line 117: I don't understand why you have the right to do that `(bus.getVoltageControl().isPresent() && bus.isVoltageControlled()` as the voltage control is optional. How you deal with that ?

annetill: Indeed Sonar says that it is a bug :-)

floriand-e2r: I corrected the sonar bug, I hope it is more clear now!

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
floriand-e2r: Maybe we should add for clarity some kind of `LfBus::isVoltageControllerEnabled()` method? or a `VoltageControl::getEnabledControllerBuses()`?

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
floriand-e2r: I removed the mode, as `VoltageControl` is only structural, the state (`voltageControlEnabled`) remains in `LfBus`.

floriand-e2r: `setVoltageControl()` is moreover renamed into `setVoltageControlEnabled()`

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
```diff
@@ -36,7 +38,7 @@
 
     private static final Comparator<PvToPqBus> BY_NOMINAL_V_COMPARATOR = Comparator.comparingDouble(pvToPqBus -> {
         LfBus bus = pvToPqBus.bus;
-        LfBus controlledBus = bus.getControlledBus().orElse(null);
+        LfBus controlledBus = bus.hasVoltageControllerCapability() ? bus.getVoltageControl().getControlledBus() : null;
```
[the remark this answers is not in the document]
floriand-e2r: `hasVoltageControllerCapability()` now only checks if `getVoltageControl()` is not null. I kept the method anyway as I find it clearer when looking for pqToPvBuses, but it's the only place where it is called

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
floriand-e2r: Property removed

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
floriand-e2r: renamed.

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -130,6 +137,13 @@ public double getParticipationFactor() {
         return participationFactor;
     }
 
+    @Override
+    public String getControlledBusId(boolean breakers) {
+        Terminal terminal = generator.getRegulatingTerminal();
+        Bus controlled = breakers ? terminal.getBusBreakerView().getBus() : terminal.getBusView().getBus();
```
annetill, line 143: You have to fix this: you have to check if the bus is not `null` and not creating the voltage control if it is. We do that for the discrete voltage control.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -68,7 +68,7 @@ private static void createVoltageControls(LfNetwork lfNetwork, List<LfBus> lfBus
                 // check target voltage
                 checkUniqueTargetVControllerBus(lfGenerator, controllerTargetV, controllerBus, generatorControlledBus);
 
-                if (lfGenerator.hasVoltageControl()) {
+                if (lfGenerator.hasVoltageControl() && controllerBus.isVoltageControllerEnabled()) {
```
annetill, line 71: This condition can be done before to avoid not usufruit checks ?

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
floriand-e2r: Done. In the last commit I do it earlier, in the LfGenerator constructor, so that the voltage control is disabled from the start.

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -335,16 +337,8 @@ public void writeJson(Writer writer) {
...
-        int controllerBusCount = 0;
-        for (LfBus bus : busesById.values()) {
-            if (bus.getControlledBus().isPresent()) {
-                controlledBusCount++;
-            }
-            if (!bus.getControllerBuses().isEmpty()) {
-                controllerBusCount++;
-            }
-        }
+        long controllerBusCount = busesById.values().stream().filter(LfBus::isVoltageControllerEnabled).count();
```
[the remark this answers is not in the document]
floriand-e2r: Corrected, but what was counted before included the voltage controls discarded for too small max reactive ranges and for not started generators... so it's still annoying to compare the results.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
floriand-e2r: Done; in fact some refactoring has been done here, to avoid the null/NaN initialization and the useless checks.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -39,21 +43,82 @@
...
+
+                    // check that remote control bus is the same for the generators of current controller bus which have voltage control on
+                    checkUniqueControlledBus(finalControlledBus, generatorControlledBus, controllerBus);
+
+                    // check that target voltage is the same for the generators of current controller bus which have voltage control on
+                    checkUniqueTargetVControllerBus(lfGenerator, controllerTargetV, controllerBus, generatorControlledBus);
+                });
+
+                if (!voltageRemoteControl && controlledBus != controllerBus) {
+                    // if voltage remote control deactivated and remote control, set local control instead
```
floriand-e2r, line 78: This could be done earlier to have the lfGenerator consistent with the local voltage control, but for now it's easier to do it here.

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -335,18 +337,10 @@ public void writeJson(Writer writer) {
...
-        int controllerBusCount = 0;
-        for (LfBus bus : busesById.values()) {
-            if (bus.getControlledBus().isPresent()) {
-                controlledBusCount++;
-            }
-            if (!bus.getControllerBuses().isEmpty()) {
-                controllerBusCount++;
-            }
-        }
+        long remoteControllerBusCount = busesById.values().stream().filter(b -> b.isVoltageControllerEnabled() && !b.isVoltageControlled()).count();
```
geofjamg, line 340: Not really equivalent to previous car as we iterate 2 time over all buses.
Maybe not an issue?

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
```diff
@@ -0,0 +1,44 @@
+package com.powsybl.openloadflow.network;
```
geofjamg, line 1: License

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
```diff
@@ -0,0 +1,44 @@
...
+        return controllers;
+    }
+
+    public void addControllerBus(LfBus controllerBus) {
+        Objects.requireNonNull(controllerBus);
+        controllers.add(controllerBus);
+        controllerBus.setVoltageControl(this);
+    }
+
+    public boolean isVoltageControlLocal() {
```
geofjamg, line 41: The name is not clear, at first read I was thinking it means this is only a local control

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -77,6 +65,8 @@
 
     protected final List<LfBranch> branches = new ArrayList<>();
 
+    private Optional<VoltageControl> voltageControl = Optional.empty();
```
geofjamg, line 68: Never store an optional as an attribute!!!

src/test/java/com/powsybl/openloadflow/OpenLoadFlowParametersTest.java
```diff
@@ -164,7 +164,7 @@ void testMaxIterationReached() {
         parameters.setWriteSlackBus(true);
         Network network = EurostagTutorialExample1Factory.create();
         LoadFlow.Runner loadFlowRunner = new LoadFlow.Runner(new OpenLoadFlowProvider(new SparseMatrixFactory()));
-        network.getGenerator("GEN").setTargetV(5);
+        network.getVoltageLevelStream().forEach(vl -> vl.setNominalV(21));
```
geofjamg, line 167: Why ?

src/test/java/com/powsybl/openloadflow/ac/GeneratorTargetVoltageInconsistencyTest.java
```diff
@@ -45,7 +45,7 @@ void localTest() {
                 .setMinP(0)
                 .setMaxP(200)
                 .setTargetP(100)
-                .setTargetV(413)
+                .setTargetV(23)
```
geofjamg, line 48: Could you explain why you changed unit tests?

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreationParameters.java
```diff
@@ -11,29 +11,22 @@
...
-
     private final boolean phaseControl;
 
     private final boolean transformerVoltageControl;
 
     private final boolean forceA1Var;
 
-    public AcEquationSystemCreationParameters(boolean voltageRemoteControl, boolean phaseControl, boolean transformerVoltageControl) {
-        this(voltageRemoteControl, phaseControl, transformerVoltageControl, false);
+    public AcEquationSystemCreationParameters(boolean phaseControl, boolean transformerVoltageControl) {
```
geofjamg, line 20: Ok so remote voltage control de-activation is only managed at network creation.

geofjamg (changes requested):
Just minor remarks or questions. This is a large refactoring, lots of code has been moved and so this is very difficult to review. But new design looks good and is clearly better than previous one.

src/test/java/com/powsybl/openloadflow/ac/GeneratorTargetVoltageInconsistencyTest.java
annetill: Because the previous test was not real with two generators connected to a 20 kV voltage level and with target V around 400 kV, without any remote regulating terminal.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
```diff
@@ -111,4 +129,49 @@ public void setCalculatedQ(double calculatedQ) {
...
+        }
+        return consistency;
+    }
+
+    protected void setTargetV(double targetV) {
+        double newTargetV = targetV;
+        // check that targetV has a plausible value (wrong nominal voltage issue)
+        if (targetV < PlausibleValues.MIN_TARGET_VOLTAGE_PU || targetV > PlausibleValues.MAX_TARGET_VOLTAGE_PU) {
+            LOGGER.warn("Generator '{}' has an inconsistent target voltage: {} pu", getId(), targetV);
+            newTargetV = (targetV > PlausibleValues.MAX_TARGET_VOLTAGE_PU) ? PlausibleValues.MAX_TARGET_VOLTAGE_PU : PlausibleValues.MIN_TARGET_VOLTAGE_PU;
```
annetill, line 172: @geofjamg Just tell if if this fix on targetV is okay for you.

src/test/java/com/powsybl/openloadflow/OpenLoadFlowParametersTest.java
floriand-e2r: That was just a way to get max iteration reached, as setting a target value too far from nominal voltage is not possible anymore. I improved it with a clearer and cleaner change, and added a comment.

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
floriand-e2r: Indeed, but in fact we call it only once, to check just after if it is only local! hence I changed its content to return whether it is a local control only.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreationParameters.java
floriand-e2r: Indeed, after that we do not need that boolean anymore.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
geofjamg: why not!

geofjamg (approved):

annetill (approved):
What a great work @floriand-e2r ! Thanks a lot !

PR #177, opened 2020-12-03
Discard phase control if controlled branch is null or open

annetill:

No check is done.

Two checks have been added and in case of null controlled branch or open controlled branch, the phase control is not created.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
@@ -249,6 +249,14 @@ public class LfNetworkLoaderImpl implements LfNetworkLoader {
                 controlledBranchId += legId;
             }
             LfBranch controlledBranch = lfNetwork.getBranchById(controlledBranchId);
+            if (controlledBranch == null) {
+                LOGGER.warn("Phase controlled branch {} is null: no phase control created", controlledBranchId);
+                return;
+            }
+            if (controlledBranch.getBus1() == null || controlledBranch.getBus2() == null) {
+                LOGGER.warn("Phase controlled branch {} is open: no phase control created", controlledBranch.getId());
+                return;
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -249,6 +249,14 @@ private static void createPhaseControl(LfNetwork lfNetwork, PhaseTapChanger ptc,
                 controlledBranchId += legId;
             }
             LfBranch controlledBranch = lfNetwork.getBranchById(controlledBranchId);
+            if (controlledBranch == null) {
+                LOGGER.warn("Phase controlled branch {} is null: no phase control created", controlledBranchId);
+                return;
+            }
+            if (controlledBranch.getBus1() == null || controlledBranch.getBus2() == null) {
```
floriand-e2r, line 256: Is this check really useful? It seems equivalent to the one at line 265 `ptc.getRegulationTerminal().getBusView().getBus() == null`. In the tests at least there's no counterexample. Thus we never go through the line 266 anymore in the tests.

floriand-e2r (changes requested):

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
annetill: The controlled branch could be open at one side: it could be the side of the regulating terminal or the other one.

floriand-e2r: Ok. But then we should remove the test of line 265 which is useless. And it would be maybe better to add such a test of a null branch on the other side? Indeed if I comment out the lines 256-259 all tests stay passed, which shows that when it's null it's always on the regulating side.

PR #178, opened 2020-12-03
Distributed slack management in sensitivity analysis

Djazouli:

Create a feature allowing the user to use distributed slack in sensitivity analysis

**Other information**:

This is currently WIP, I opened the PR to request feedback


src/main/java/com/powsybl/openloadflow/ac/AbstractDistributedSlackOuterLoop.java
```diff
@@ -30,11 +30,11 @@
      */
     private static final double SLACK_P_RESIDUE_EPS = Math.pow(10, -5);
 
-    protected static class ParticipatingElement<T> {
+    public static class ParticipatingElement<T> {
```
Djazouli, line 33: I really want to get rid of these "public" changes, but I need to get those values somewhere else in the code
Should I just create a protected method that returns something understandable by the rest of the code ?
(returning Pair<String, Number> instead of participatingElement for example)

src/main/java/com/powsybl/openloadflow/dc/equations/AbstractClosedBranchDcFlowEquationTerm.java
```diff
@@ -71,4 +71,8 @@ public int getSubjectNum() {
     public boolean hasRhs() {
         return true;
     }
+
```
Djazouli, line 74: Shouldn't be here, I will remove it

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +195,94 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
 
-    public List<SensitivityValue> analyse(Network network, List<SensitivityFactor> factors, OpenLoadFlowParameters lfParametersExt,
+    /**
+     * If the slack is distributed, we need to create a factor for every generator
+     * @param network
+     * @param lfNetwork
+     * @param oldFactors
+     * @return
+     */
+    private List<SensitivityFactor> getDistributedFactors(Network network, LfNetwork lfNetwork, List<SensitivityFactor> oldFactors) {
```
Djazouli, line 205: this method takes the network as input, because I need to extract generators. But this may be an issue, so I'd like to use just lfNetwork, but I do not know how to extract generators from it

src/test/java/com/powsybl/openloadflow/sensi/OpenSensitivityAnalysisProviderTest.java
```diff
@@ -131,6 +145,38 @@ void testDc4buses() {
         assertEquals(0.25d, getValue(result, "g4", "l13"), LoadFlowAssert.DELTA_POWER);
     }
 
+    @Test
+    void testDc4busesDistributed() {
```
Djazouli, line 149: Test does not pass at the moment, it's [WIP]

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +192,100 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+     * @param openLoadFlowParameters
+     * @return
+     */
+    private List<Pair<String, Number>> getParticipatingElements(LfNetwork lfNetwork, LoadFlowParameters loadFlowParameters, OpenLoadFlowParameters openLoadFlowParameters) {
+        List<Pair<String, Number>> contributionList = new ArrayList<>();
+
+        if (loadFlowParameters.isDistributedSlack()) {
+            switch (loadFlowParameters.getBalanceType()) {
+                case PROPORTIONAL_TO_GENERATION_P_MAX:
+                    DistributedSlackOnGenerationOuterLoop outerLoop = new DistributedSlackOnGenerationOuterLoop(openLoadFlowParameters.isThrowsExceptionInCaseOfSlackDistributionFailure());
```
Djazouli, line 256: This is were I need all this outer loop stuff to be Public, I would love to avoid this

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +192,100 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
 
-    public List<SensitivityValue> analyse(Network network, List<SensitivityFactor> factors, OpenLoadFlowParameters lfParametersExt,
+    /**
+     * If the slack is distributed, we need to create a factor for every generator
+     * @param network
+     * @param lfNetwork
+     * @param oldFactors
+     * @return
+     */
+    private List<SensitivityFactor> getDistributedFactors(Network network, LfNetwork lfNetwork,
```
Djazouli, line 202: This method uses network to get generators. But lfNetwork may not contain all these generators (because it is a connected component of network), so I would love to use only lfNetwork, but I do not know how to access generators from LfNetwork

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +192,100 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+        }
+        return contributionList;
+    }
+
+    /**
+     * Create a map containing sensibility values, indexed first by FunctionId, then by factorId
+     * @param sensitivityValues
+     * @param remainder
+     * @return
+     */
```
Djazouli, line 273: hardly understandable, need refactoring

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -225,6 +320,46 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
         DenseMatrix states = solveTransposed(rhs, j);
 
         // calculate sensitivity values
-        return calculateSensitivityValues(lfNetwork, equationSystem, factorsByVarConfig, states);
+        List<SensitivityValue> sensitivityValues = calculateSensitivityValues(lfNetwork, equationSystem, factorsByVarConfig, states);
+
+        if (!lfParameters.isDistributedSlack()) {
+            return sensitivityValues;
```
Djazouli, line 326: Having a return in the middle of the function makes it harder to understand, I need to find a better way to distinguish the distributed/ not distributed case

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +197,146 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+            if (initialFactor instanceof BranchFlowPerInjectionIncrease) {
+                BranchFlowPerInjectionIncrease branchFactor = (BranchFlowPerInjectionIncrease) initialFactor;
+                monitoredLines.add(branchFactor.getFunction().getBranchId());
+            } else {
+                factors.add(initialFactor);
+            }
+        }
+
+        switch (balanceType) {
+            case PROPORTIONAL_TO_GENERATION_P_MAX:
```
annetill, line 223: @Djazouli Geoffory prefers that we do not use factors that whould remain in the API. So we have to fix that.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +197,146 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+    }
+
+    /**
+     * Return a list of id from participating element, and its participation value
+     *
+     * @param lfNetwork
+     * @param loadFlowParameters
+     * @return
+     */
+    private List<Pair<String, Number>> getParticipatingElements(LfNetwork lfNetwork, LoadFlowParameters loadFlowParameters) {
```
annetill, line 259: I think here we just have to have as argument the balance type.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
annetill: Indeed, it is the part of the code that is the less understable for me.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +197,146 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+        for (String lfBranchId : sensitivityValuesMap.keySet()) {
+            Map<String, SensitivityValue> sensitivitiesByBranch = sensitivityValuesMap.get(lfBranchId);
+            double sum = participatingElements.stream()
+                    .mapToDouble(element -> element.getRight().doubleValue() * sensitivitiesByBranch.get(element.getLeft()).getValue())
+                    .sum();
+            slackDistributionCorrection.put(lfBranchId, sum);
+        }
+
+        // compute sensitivities corrected from the slack distribution effects
+        for (SensitivityValue sensitivityValue : sensitivityValues) {
```
annetill, line 326: Here a fix is needed because I don't have sort the sensitivity values to give only those that have been ask in the API. @Djazouli your code was functionaly right but it was to remove the use of factors.

annetill (commented):
Tests have to be added. I think that this PR should not have more features but should be tested and improved before review.

src/test/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysisTest.java
```diff
@@ -79,6 +83,63 @@ void test4buses() {
...
+                sensiParameters, LocalComputationManager.getDefault())
+                                                        .join();
+
+        assertEquals(5, result.getSensitivityValues().size());
+        assertEquals(0.175d, getValue(result, "g1", "l14"), LoadFlowAssert.DELTA_POWER);
+        assertEquals(0.275d, getValue(result, "g1", "l12"), LoadFlowAssert.DELTA_POWER);
+        assertEquals(-0.125d, getValue(result, "g1", "l23"), LoadFlowAssert.DELTA_POWER);
+        assertEquals(0.025d, getValue(result, "g1", "l34"), LoadFlowAssert.DELTA_POWER);
+        assertEquals(0.15d, getValue(result, "g1", "l13"), LoadFlowAssert.DELTA_POWER);
     }
```
Djazouli, line 143: Todo: 
add tests:

- where we are injecting on a load node that has no generator
- Where the slack bus is participating in the compensation
- Where the balance type is PROPORTIONAL_TO_LOAD

Djazouli:
We still need to test balance type : PROPORTIONAL_TO_LOAD to improve coverage

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -153,8 +156,11 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
         return rhs;
     }
 
-    private Map<SensitivityVariableConfiguration, SensitivityFactorGroup> indexFactorsByVariableConfig(Network network, List<SensitivityFactor> factors) {
+    private Map<SensitivityVariableConfiguration, SensitivityFactorGroup> indexFactorsByVariableConfig(Network network, List<SensitivityFactor> factors, LfNetwork lfNetwork, LoadFlowParameters loadFlowParameters) {
         Map<SensitivityVariableConfiguration, SensitivityFactorGroup> factorsByVarConfig = new LinkedHashMap<>(factors.size());
+        Map<String, Number> participationMap = getParticipationMap(network, lfNetwork, loadFlowParameters); // empty if slack is not distributed
```
annetill, line 161: Maybe participationFactorMap ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -153,8 +156,11 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+        participationMap.remove(lfNetwork.getSlackBus().getId()); // the injection on the slack bus will not appear in the rhs
+        participationMap.replaceAll((key, value) -> -value.doubleValue()); // the compensation on a bus will be the opposite of its participation
```
annetill, line 163: compensation -> slack distribution

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -164,7 +170,12 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
                 Bus bus = injection.getTerminal().getBusView().getBus();
                 // skip disconnected injections
                 if (bus != null) {
-                    SensitivityVariableConfiguration varConfig = new SensitivityVariableConfiguration(Collections.singleton(bus.getId()), Collections.emptySet());
+                    Map<String, Number> busCompensation = new HashMap<>(participationMap);
+                    // add 1 where we are making the injection
+                    busCompensation.put(bus.getId(), busCompensation.getOrDefault(bus.getId(), 0).doubleValue() + 1);
```
annetill, line 175: busCompensation is not the good name choice indeed.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +203,50 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+            ParticipatingElement.normalizeParticipationFactors(participatingElements, "bus");
+
+            participatingElements.forEach(participatingElement -> {
+                String busId;
+                if (participatingElement.getElement() instanceof LfGenerator) {
+                    busId = network.getGenerator(((LfGenerator) participatingElement.getElement()).getId()).getTerminal().getBusView().getBus().getId();
+                } else if (participatingElement.getElement() instanceof LfBus) {
+                    busId = ((LfBus) participatingElement.getElement()).getId();
+                } else {
+                    throw new UnsupportedOperationException("Only buses can have an impact in slack distribution");
```
annetill, line 241: Indeed I am not sure that you can cover easily this line.

annetill (commented):
Maybe some slight variable names have to be changed to improve readability. I have a question about injection increase for a generator or a load that is connected to the slack bus: for the moment, we throw an exception. Is it the final behaviour that we aim ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -142,7 +142,7 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
             for (Map.Entry<String, Number> busAndInjection : configuration.busInjectionById().entrySet()) {
                 LfBus lfBus = lfNetwork.getBusById(busAndInjection.getKey());
                 if (lfBus.isSlack()) {
-                    throw new PowsyblException("Cannot analyse sensitivity of slack bus");
```
annetill, line 142: I am not sure to get why this line is not cover by our test.

Djazouli: It is because we never reach this line, the slack bus will never be inside the businjectionbyId. But the exception here stays for the safety of future developments

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +209,50 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+                    step = new LoadActivePowerDistributionStep(false, false);
+                    break;
+                default:
+                    throw new UnsupportedOperationException("Balance type not yet supported: " + loadFlowParameters.getBalanceType());
+            }
+
+            participatingElements = step.getParticipatingElements(lfNetwork);
+            ParticipatingElement.normalizeParticipationFactors(participatingElements, "bus");
+
+            participatingElements.forEach(participatingElement -> {
```
geofjamg, line 240: You can directly create the map like:
```java
Map<String, Number> participationFactorByBusMap = participatingElements.stream().collect(Collectors.toMap(..));
```

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -203,12 +263,22 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
         List<LfNetwork> lfNetworks = LfNetwork.load(network, lfParametersExt.getSlackBusSelector());
         LfNetwork lfNetwork = lfNetworks.get(0);
 
-        // create DC equation system
-        EquationSystem equationSystem = DcEquationSystem.create(lfNetwork, new VariableSet(), new DcEquationSystemCreationParameters(false, true, true, lfParametersExt.isDcUseTransformerRatio()));
+        // run DC load
+        Map<String, Double> functionReferenceByBranch = new HashMap<>();
+        DcLoadFlowParameters dcLoadFlowParameters = new DcLoadFlowParameters(lfParametersExt.getSlackBusSelector(),
+                new SparseMatrixFactory(), true, lfParametersExt.isDcUseTransformerRatio(), lfParameters.isDistributedSlack(),  lfParameters.getBalanceType());
```
geofjamg, line 269: MatrixFactory should be taken from parameters.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +209,50 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
     }
 
-    public List<SensitivityValue> analyse(Network network, List<SensitivityFactor> factors, OpenLoadFlowParameters lfParametersExt,
+    /**
+     * Return a mapping between the participating element id through its connected bus, and its participation value in the slack distribution
+     * @param lfNetwork
+     * @param loadFlowParameters
+     * @return
+     */
+    private Map<String, Number> getParticipationFactorByBus(Network network, LfNetwork lfNetwork, LoadFlowParameters loadFlowParameters) {
```
geofjamg, line 218: Map<String, Double> ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -36,17 +39,17 @@
 
     static class SensitivityVariableConfiguration {
```
geofjamg, line 40: Maybe we should remove SensitivityVariableConfiguration? I added it to avoid computing same sensibility multiple times but there is no reason for that to happen frequently (user input mistake?) and it adds useless complexity.
So instead of `Map<SensitivityVariableConfiguration, SensitivityFactorGroup> factorsByVarConfig` we would only have `List<SensitivityFactor<?, ?>>`

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -203,12 +263,22 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
         List<LfNetwork> lfNetworks = LfNetwork.load(network, lfParametersExt.getSlackBusSelector());
         LfNetwork lfNetwork = lfNetworks.get(0);
 
-        // create DC equation system
-        EquationSystem equationSystem = DcEquationSystem.create(lfNetwork, new VariableSet(), new DcEquationSystemCreationParameters(false, true, true, lfParametersExt.isDcUseTransformerRatio()));
+        // run DC load
+        Map<String, Double> functionReferenceByBranch = new HashMap<>();
+        DcLoadFlowParameters dcLoadFlowParameters = new DcLoadFlowParameters(lfParametersExt.getSlackBusSelector(),
+                new SparseMatrixFactory(), true, lfParametersExt.isDcUseTransformerRatio(), lfParameters.isDistributedSlack(),  lfParameters.getBalanceType());
+        DcLoadFlowResult dcLoadFlowResult = new DcLoadFlowEngine(lfNetworks, dcLoadFlowParameters).run();
```
geofjamg, line 270: As we calculate sensitivity only on first component do we really need to run the initial DC loadflow on all components?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -192,7 +209,50 @@ private DenseMatrix initRhs(LfNetwork lfNetwork, EquationSystem equationSystem,
...
+                    throw new UnsupportedOperationException("Balance type not yet supported: " + loadFlowParameters.getBalanceType());
+            }
+
+            participatingElements = step.getParticipatingElements(lfNetwork);
+            ParticipatingElement.normalizeParticipationFactors(participatingElements, "bus");
+
+            participatingElements.forEach(participatingElement -> {
+                String busId;
+                if (participatingElement.getElement() instanceof LfGenerator) {
+                    busId = network.getGenerator(((LfGenerator) participatingElement.getElement()).getId()).getTerminal().getBusView().getBus().getId();
```
geofjamg, line 243: We should avoid searching generator into IIDM model. We could just get LfBus from LfGenerator.

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
geofjamg, line 270: Maybe for later but running a DC LF using DcLoadFlowEngine is not efficient as we later need to build the same equation system and jacobian for sensitivity analysis.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
annetill: Done, but tp be checked by @Djazouli !

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
annetill: Indeed, in the DCLoadFlowEngine, we explicitely compute the LF just on the first connected component. So I leave it like that?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
annetill: Indeed there is no existing method, you mean that we have to implement the getter on all LfGenerator or there is something easier?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: I feel like SensitivityVariableConfigurations are useful when the user gives a provider that contains all the generators of the network. Not sure if this is considered as an "user input mistake". But in this case, the SensitivityVariableConfiguration is useful to remove duplicate (generators connected on the same bus).

geofjamg: Ok so let s keep it

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
geofjamg: Ok

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
geofjamg: Yes

Djazouli: I have tried something in [the latest commit](https://github.com/powsybl/powsybl-open-loadflow/pull/178/commits/c4342720a2a8c6388f03aac3da1af2cd8f6a329b), tell me if this works for you:

geofjamg (approved):

annetill (approved):
Thanks a lot @Djazouli and @Hadrien-Godard for this great PR !

Issue #80, opened 2020-05-04
SVC reactive limits modeling could be improved


annetill:
* **Do you want to request a *feature* or report a *bug*?**

More a *feature* about SVC reactive limits.

* **What is the current behavior?**

Currently, minQ and maxQ are defined using Bmin and Bmax and the nominal voltage of the connected bus. It is in fact an approximation.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

In a way, reactive power furniture or reactive power consumption could be over estimated or under estimated using nominal voltage.

* **What is the expected behavior?**

That the limits depend on calculated voltage. 

* **What is the motivation / use case for changing the behavior?**

To have the right reactive limits in order to have the more right power flow solution.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

adriguir:
The behavior of the Static Var Compensator should be modified so that the QMin and QMax are calculated on the fly during the external loop using BMin * the calculated value of U from the terminal.

PR #180, opened 2020-12-08
Static Var Compensator : MinQ and MaxQ are computed using local voltage

frigaux:

https://github.com/powsybl/powsybl-open-loadflow/issues/80

improvement

```
double nominalV = svc.getTerminal().getVoltageLevel().getNominalV();
double minQ = svc.getBmin() * nominalV * nominalV;
```

```
double v = bus.getV() * bus.getNominalV();
double minQ = svc.getBmin() * v * v;
```

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfStaticVarCompensatorImpl.java
```diff
@@ -24,23 +24,21 @@
...
-        double nominalV = svc.getTerminal().getVoltageLevel().getNominalV();
-        // min and  max reactive limit are calculated at nominal voltage
-        double minQ = svc.getBmin() * nominalV * nominalV;
-        double maxQ = svc.getBmax() * nominalV * nominalV;
         reactiveLimits = new MinMaxReactiveLimits() {
 
             @Override
             public double getMinQ() {
-                return minQ;
+                double v = bus.getV() * bus.getNominalV();
```
annetill, line 34: The bus could be `null` (it means that the SVC is not connected). You are going to met a null pointer exception here. Same remark to get the nomial voltage, you have to get it from the terminal. I have to check but we could met problems in security analysis when some equipements are disconnected from the network. @geofjamg what is your opinion about this desgin ?

src/main/java/com/powsybl/openloadflow/network/impl/LfStaticVarCompensatorImpl.java
```diff
@@ -24,23 +24,21 @@
...
+                return svc.getBmin() * v * v;
             }
 
             @Override
             public double getMaxQ() {
-                return maxQ;
+                double v = bus.getV() * bus.getNominalV();
```
annetill, line 40: Same remark here.

annetill (commented):
We need @geofjamg opinion about that and maybe adding test of the security analysis in a network that contains a SVC and that becomes disconnected to see if all is okay (in this outerloop for exalmple).

src/main/java/com/powsybl/openloadflow/network/impl/LfStaticVarCompensatorImpl.java
frigaux: Hello,
1/ For the moment, bus cannot be null because **LfStaticVarCompensatorImpl** is created by **AbstractLfBus** himself (AbstractLfBus : line 297).
```
    void addStaticVarCompensator(StaticVarCompensator staticVarCompensator, double scaleV, LfNetworkLoadingReport report) {
        if (staticVarCompensator.getRegulationMode() != StaticVarCompensator.RegulationMode.OFF) {
            add(LfStaticVarCompensatorImpl.create(staticVarCompensator, this),
                    staticVarCompensator.getRegulationMode() == StaticVarCompensator.RegulationMode.VOLTAGE,
                    staticVarCompensator.getVoltageSetPoint() * scaleV, -staticVarCompensator.getReactivePowerSetPoint(),
                    report);
        }
    }
```

2/ **bus.getNominalV()** returns a primitive type and thus cannot be null (but could be NaN ?)

3/ Considering `bus.getV() * bus.getNominalV();`
Here is code we can found in **AbstractLfBus** :
```
    public double getV() {
        return v / getNominalV();
    }
```
That's why I have to multiply by bus.getNominalV() to get v, is that correct ?

geofjamg (approved):

PR #183, opened 2020-12-09
Slack distribution support for DC power flow

annetill:

Slack distribution was not supported before for DC power flow. The slack was entirely put at the slack node.

Slack distribution is now supported for balance type PROPORTIONAL_TO_LOAD, PROPORTIONAL_TO_P_MAX and PROPORTIONAL_TO_CONFORM_LOAD. The power factor of loads is ignored as we are in a DC approximation. Looses are neglected. I need to add unit tests. A refactoring is needed for pre-distribution of slack used here and in the security analysis.

**Other information**:


src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -37,24 +40,22 @@
...
-        updateFlows = parameters.isUpdateFlows();
-        useTransformerRatio = parameters.isUseTransformerRatio();
+        this.networks = LfNetwork.load(network, new LfNetworkParameters(parameters.getSlackBusSelector(), false, false, false, false));
+        this.parameters = Objects.requireNonNull(parameters);
+    }
+
+    private void distributeSlack(LfNetwork network) {
+        double mismatch = network.getActivePowerMismatch();
+        ActivePowerDistribution activePowerDistribution = ActivePowerDistribution.create(parameters.getBalanceType(), false);
+        activePowerDistribution.run(network, -mismatch);
```
geofjamg, line 58: Maybe we shoud invert the sign of Network.getActivePowerMismatch to avoid this minus anf be consistent everywhere?

annetill: I have changed the sign. Do you think that I need to perform a functional validation ?

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
geofjamg: I should be ok

PR #187, opened 2020-12-15
Improve parameters management

annetill:

No.

The default values of  `remoteVoltageControl` is true. The default value of `dcuseTransformerRatio` is true.
I have added `updateFlows` as a parameter that can be changed in the config file, but I am not sure to really understand the need of this parameter.

**Other information**:

I wonder about the sensitivity specific parameter that are not managed through the configuration file.

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
@@ -31,7 +31,7 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
 
     private boolean throwsExceptionInCaseOfSlackDistributionFailure = THROWS_EXCEPTION_IN_CASE_OF_SLACK_DISTRIBUTION_FAILURE_DEFAULT_VALUE;
 
-    private boolean voltageRemoteControl = VOLTAGE_REMOTE_CONTROLE_DEFAULT_VALUE;
+    private boolean voltageRemoteControl = VOLTAGE_REMOTE_CONTROL_DEFAULT_VALUE;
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -109,6 +111,15 @@ public OpenLoadFlowParameters setDcUseTransformerRatio(boolean dcUseTransformerR
         return this;
     }
 
+    public boolean isUpdateFlows() {
```
annetill, line 114: I am not sure to understand if we really need this kind of parameters. In which use case are we doing not to update flows ?

geofjamg: At this level (external API), we always wang to update flows so we should remove it here.
Internally (in AC/DC engines) we sometimes don't want to update flows.

src/main/java/com/powsybl/openloadflow/util/ParameterConstants.java
```diff
@@ -32,7 +32,10 @@
     public static final boolean LOAD_POWER_FACTOR_CONSTANT_DEFAULT_VALUE = false;
 
     public static final String DC_USE_TRANSFORMER_RATIO_PARAM_NAME = "dcUseTransformerRatio";
-    public static final boolean DC_USE_TRANSFORMER_RATIO_DEFAULT_VALUE = false;
+    public static final boolean DC_USE_TRANSFORMER_RATIO_DEFAULT_VALUE = true;
+
+    public static final String UPDATE_FLOWS_PARAM_NAME = "updateFlows";
```
geofjamg, line 37: Not use anymore?

geofjamg (approved):

Issue #186, opened 2020-12-11
Manage Contingencies in Sensitivity Analysis 


Djazouli:
* **Do you want to request a *feature* or report a *bug*?**
A feature

* **What is the current behavior?**
Currently, if a ContingenciesProvider is supplied to the OpenSensitivityAnalysisProvider.run method, an UnsupportedOperationException is raised.

* **What is the expected behavior?**
We should be able to supply sensitivity values for each contingency.

annetill:
**Use case**
	- Sensitivity of +1 MW on a bus n on the flow of the line (i,j) with the contingency of the line (m,k)
	- In a first approach, we will not support phase shifter at line (m, k)
	
**Right hand side B**
	- First vector with +1 MW at bus n
	- Second vector with +1 MW at bus m and -1 MW at bus k
	         - If bus m is the slack bus, the second vector is only +1 MW at bus k
	         - If bus k is the slack bus, the second vector is only -1 MW at bus m
	
We solve the system JT = B from the LU factorisation to obtain T, the state vector of angles. Then:
	
1) Computation of sensitivity value of bus n on line (i,j) (already exists):
	Snij = (Ti1 - Tj1)/Xij
	
2) Computation of sensitivity value N-1 of bus n one line (i,j) with a line contingency:
	Snijmk = Snij + (Tm1-Tk1)/(Xmk - (Tm2 - Tk2)) * (Ti2-Tj2)/Xij

Djazouli:
The 2. mentioned above comes from the inversion of a linear system. This method will always work and give the right result, except if the system cannot be inverted. This can only happen if adding +1-1 on the buses of the line (m,k) produce a flow of 1 on this line. This can only happen if there no other path between m and k than this branch. So, the line (m,k) is a "bridge" between two parts of the network, let's call them A and B. The slack bus is in A with m, and k is in B.

In this case, adding a +1-1 on m and k cannot change the flow on the branch (i,j) that we are watching.
So, our N-1 sensitivity can only be equal to the N sensitivity, if the N-1 sensitivity makes sense.

Let's call b the bus where we are making an injection. 

If b is in A, then the aforementioned system becomes 0=0, so, even if it is not invertible, we can solve it, and the N-1 sensitivity makes sense.

If b is in , the system becomes 0=1, and we cannot solve it. We then have two cases:

- If (i,j) was in A, the sensitivity is obviously 0.
[...]

PR #189, opened 2020-12-17
Add systematic analysis with contingencies to sensitivity analysis

Djazouli:

#186


src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis/DcSlowContingencyAnalysis.java
```diff
@@ -0,0 +1,116 @@
...
+import com.powsybl.openloadflow.network.LfNetwork;
+import com.powsybl.openloadflow.sensi.OpenSensitivityAnalysisParameters;
+import com.powsybl.sensitivity.SensitivityFactor;
+import com.powsybl.sensitivity.SensitivityValue;
+import org.jgrapht.alg.util.Pair;
+
+import javax.annotation.Nullable;
+import java.util.*;
+
+public class DcSlowContingencyAnalysis extends AbstractDcSensitivityAnalysis {
```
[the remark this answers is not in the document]
Djazouli: Yes this is definitely a temporary dev

src/main/java/com/powsybl/openloadflow/sensi/AbstractDcSensitivityAnalysis.java
```diff
@@ -131,57 +156,60 @@ private SensitivityValue createBranchSensitivityValue(LfNetwork lfNetwork, Equat
...
 
         // index factors by variable config
         for (SensitivityFactor<?, ?> factor : factors) {
             if (factor instanceof BranchFlowPerInjectionIncrease) {
                 BranchFlowPerInjectionIncrease injectionFactor = (BranchFlowPerInjectionIncrease) factor;
-                Injection<?> injection = getInjection(network, injectionFactor.getVariable().getInjectionId());
+                Injection<?> injection = getInjection(network, injectionFactor.getVariable().getInjectionId(), throwExceptionIfNullInjection());
+                if (injection == null) {
+                    LOGGER.debug("Injection {} not found in the network", injectionFactor.getVariable().getInjectionId());
+                    // FIXME
```
annetill, line 204: I have added here a `FIXME` because after the code review I think that we need a decidate analysis of contingencies and factors to avoid treatements all over the code. This analysis will be done before the base case sensitivity analysis and before each N-k analysis. This graph analysis is needed because :
- We only support sensivity analysis on main connected component but the external API exposes the whole network. The list of contingencies provided by the external API is also over the whole network and these lead to a lot of if/else in the code ;
- We will need this kind of analysis on factors and injections involved in the slack distribution when a N-k leads to a separate connected component (connected or off load).

src/main/java/com/powsybl/openloadflow/sensi/DcSlowSensitivityAnalysis.java
```diff
@@ -0,0 +1,141 @@
...
+        JacobianMatrix j = createJacobianMatrix(equationSystem, voltageInitializer);
+
+        // solve system
+        DenseMatrix states = solveTransposed(rhs, j);
+
+        // calculate sensitivity values
+        return calculateSensitivityValues(lfNetwork, equationSystem, factorsByVarConfig, states, functionReferenceByBranch);
+    }
+
+    public Pair<List<SensitivityValue>, Map<String, List<SensitivityValue>>> analyse(Network network, List<SensitivityFactor> factors,
```
annetill, line 98: This part is never used in your unit tests:
- I am not sure it is relevant to intoduce it in that PR or later ;
- Or we could make some unit tests that compare results from the fast and the slow approches.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,409 @@
...
+        DenseMatrix states = solveTransposed(rhs, j); // states contains angles for the sensitivity factors, but also for the +1-1 related to contingencies
+
+        // sensitivities without contingency
+        List<SensitivityValue> sensitivityValues = calculateSensitivityValues(lfNetwork, equationSystem, factorsGroups,
+                states, functionReferenceByBranch, Collections.emptyList(), Collections.emptyMap());
 
-        // solve system
-        DenseMatrix states = solveTransposed(rhs, j);
+        Collection<Contingency> straightforwardContingencies = new LinkedList<>();
+        Map<List<Set<ComputedContingencyElement>>, List<Contingency>> losingConnectivityContingency = new HashMap<>();
```
geofjamg, line 341: Not clear to me why you need this data structure, why do yu need to index by element that loose connectivity? It is very unlikely that 2 contingencies share the same elements.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,409 @@
...
+                        element -> getParticipatingElementBus(element).getId(),
+                        element -> -element.getFactor(),
+                        Double::sum
+                    ))
+                ));
+                getParticipationForBus = busId -> participationPerCc.get(connectivity.getConnectivity().getComponentNumber(lfNetwork.getBusById(busId)));
+                computeFactorsInjection(getParticipationForBus, factorsGroups, predefinedResults);
+
+                ComputedContingencyElement.setGlobalIndexes(contingenciesElements.values(), factorsGroups.size());
+                rhs = initRhs(lfNetwork, equationSystem, factorsGroups, contingenciesElements);
```
geofjamg, line 379: Why do you solve multiple times (one for each list of element that loose connectivity) ?
Why you you keep in your rhs all elements and not only ones needed for contingencies that loose connectivity?

src/main/java/com/powsybl/openloadflow/sensi/AbstractDcSensitivityAnalysis.java
```diff
@@ -0,0 +1,303 @@
...
+        // assign an index to each factor group
+        int index = 0;
+        for (SensitivityFactorGroup factorGroup : groupIndexedById.values()) {
+            factorGroup.setIndex(index++);
+        }
+
+        return new ArrayList<>(groupIndexedById.values());
+    }
+
+    protected void computeFactorsInjection(Function<String, Map<String, Double>> getParticipationForBus, List<SensitivityFactorGroup> factorsGroups, Map<SensitivityFactor, Double> predefinedResult) {
```
geofjamg, line 201: Not clear to me what is the purpose of this method?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,409 @@
...
+        // compute the contingencies with no loss of connectivity
+        for (Contingency contingency : straightforwardContingencies) {
+            contingenciesValue.put(contingency.getId(), calculateSensitivityValues(lfNetwork, equationSystem, factorsGroups,
+                    states, functionReferenceByBranch, contingency.getElements().stream().map(element -> contingenciesElements.get(element.getId())).collect(Collectors.toList()),
+                    Collections.emptyMap()
+            ));
+        }
+
+        for (Map.Entry<List<Set<ComputedContingencyElement>>, List<Contingency>> contingencyEntry : losingConnectivityContingency.entrySet()) {
+            Map<SensitivityFactor, Double> predefinedResults = new HashMap<>();
```
geofjamg, line 355: What is the purpose of predefinedResults? Is it for part of the grid where connectivity has been lost where we want to overwritte results with zero?

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,409 @@
...
                 } else if (factor instanceof BranchFlowPerPSTAngle) {
-                    BranchFlowPerPSTAngle pstAngleFactor = (BranchFlowPerPSTAngle) factor;
-                    sensitivityValues.add(createBranchSensitivityValue(lfNetwork, equationSystem, pstAngleFactor.getFunction().getBranchId(),
-                            factor, factorGroup, states, functionReferenceByBranch));
+                    branchId = ((BranchFlowPerPSTAngle) factor).getFunction().getBranchId();
                 } else {
                     throw new UnsupportedOperationException("Factor type '" + factor.getClass().getSimpleName() + "' not yet supported");
                 }
+                LfBranch lfBranch = lfNetwork.getBranchById(branchId);
+                ClosedBranchSide1DcFlowEquationTerm p1 = getAndSaveEquationTerm(equationSystem, lfBranch, equationTermByBranchId);
```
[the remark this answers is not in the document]
Djazouli: Yes, but I also need to store the equation term for every factor, so I figured out that I would store everything in the same data structure

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,409 @@
...
-            SensitivityVariableConfiguration configuration = e.getKey();
-            SensitivityFactorGroup factorGroup = e.getValue();
-            for (Map.Entry<String, Double> busAndInjection : configuration.busInjectionById().entrySet()) {
-                LfBus lfBus = lfNetwork.getBusById(busAndInjection.getKey());
-                if (lfBus.isSlack()) {
-                    throw new PowsyblException("Cannot set injection increase at slack bus");
+    private List<Set<ComputedContingencyElement>> getElementsBreakingConnectivity(LfNetwork lfNetwork, DenseMatrix preContingencyStates,
+                                                                             List<ComputedContingencyElement> contingencyElements,
+                                                                             EquationSystem equationSystem) {
+        List<Set<ComputedContingencyElement>> responsibleAssociations = new LinkedList<>();
```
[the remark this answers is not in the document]
Djazouli: These are the associations of contingency elements that loose connectivity. For example, if in the network, loosing line L1 breaks connectivity, and loosing L2 and L3 together breaks connectivity, the responsible associations would be:
`[{L1}, {L2, L3}]`

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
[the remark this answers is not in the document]
geofjamg: The equation term is the same for every factor bu just depend on the branch, so if we attach it  in the ComputedContingencyElement we also avoid a map lookup as today?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,409 @@
...
+
+        if (factorsGroups.isEmpty()) {
+            return Pair.of(Collections.emptyList(), Collections.emptyMap());
+        }
 
-        if (factorsByVarConfig.isEmpty()) {
-            return Collections.emptyList();
+        // Compute the partipation for every factor
+        List<ParticipatingElement> participatingElements = null;
+        Function<String, Map<String, Double>> getParticipationForBus;
```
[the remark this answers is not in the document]
Djazouli: On this particular line, it is not needed because connectivity is guaranteed, but when we process the contingencies, we need to know on what bus the injection is made, to know in which connected component we will take the elements participating to the load balancing

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,409 @@
...
+        List<SensitivityFactorGroup> factorsGroups = createFactorGroups(network, factors);
+
+        if (factorsGroups.isEmpty()) {
+            return Pair.of(Collections.emptyList(), Collections.emptyMap());
+        }
 
-        if (factorsByVarConfig.isEmpty()) {
-            return Collections.emptyList();
+        // Compute the partipation for every factor
+        List<ParticipatingElement> participatingElements = null;
```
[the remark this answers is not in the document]
Djazouli: This variable is reused later in the code (when we process contingencies), so we need to declare it in the upper scope

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: Yes

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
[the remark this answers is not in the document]
Djazouli: This map is also used on line 152 to get equation for each branch of the sensitivity factors. So we would need to attach it to the ComputedContingencyElement, and to the SensitivityFactor

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: The idea was that if a given line is responsible for the loss of connectivity (let's say L1), then you don't have to recompute the participating factors for every N-2.
For example, all the contingencies [L1, L2], [L1, L3], [L1, Ln] will be indexed by the same `List<Set<ComputedContingencyElement>>`, and we will process all of them with only one solve.

src/main/java/com/powsybl/openloadflow/sensi/AbstractDcSensitivityAnalysis.java
Djazouli: Each FactorGroup will be represented by a column in the rhs. This method is used to know what data will be put in this column (this data changes depending on the elements that will be participating in the slack distribution, so it needs to be called for every loss of connectivity).

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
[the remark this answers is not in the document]
geofjamg: Or also have a SensitivityFactor wrapper as an internal model to allow attaching and caching informations

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
geofjamg: ok I understand

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: We need to solve it multiple times, because the distribution of the slack changed, so we need to recompute the rhs for all the injection factors.
We actually do not need to recompute anything for the contingency elements, only for the factors, but once the LU decomposition is computed, I feel like the solve isn't really expensive, so I do not mind adding a few columns for nothing.

But you're right, we could optimize it by creating a big rhs with one column per combination (factor + elements responsible for the loss of connectivity) and solve everything at once

geofjamg: The way solveTransposed is implemented does not allow reuse LU decomposition (but we could change it easily).
For solve, I did some test a few years ago and I think it was a little bit faster to group solve than calling solve multiple time (even if very fast in any cases so maybe not a bottleneck).

geofjamg: > But you're right, we could optimize it by creating a big rhs with one column per combination (factor + elements responsible for the loss of connectivity) and solve everything at once

Bu are elements have already been computed before, we can just reuse it even for contingency that loose connectivity?

geofjamg:
As a general comment, we really need to add more comments int the code to describe what is going to be calculated for each block of code and why, also explain the data structure design etc.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: Not sure I understand here. What exactly has already been computed, and can be reused ?

geofjamg: elements responsible for the loss of connectivity

src/test/resources/logback-test.xml
```diff
@@ -16,7 +16,7 @@
     <root level="info">
         <appender-ref ref="STDOUT" />
     </root>
-    <logger name="com.powsybl.openloadflow" level="DEBUG" additivity="false">
```
annetill, line 19: I am going to revert that change, it is a mistake from myself.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,405 @@
...
+                    Equation p2 = equationSystem.getEquation(bus2.getNum(), EquationType.BUS_P).orElseThrow(IllegalStateException::new);
+                    rhs.set(p1.getColumn(), element.getGlobalIndex(), 1 / PerUnit.SB);
+                    rhs.set(p2.getColumn(), element.getGlobalIndex(), -1 / PerUnit.SB);
                 }
-                SensitivityVariableConfiguration varConfig = new SensitivityVariableConfiguration(Collections.emptyMap(), Collections.singleton(phaseTapChangerHolderId));
-                factorsByVarConfig.computeIfAbsent(varConfig, k -> new SensitivityFactorGroup())
-                        .getFactors().add(pstAngleFactor);
             } else {
-                throw new UnsupportedOperationException("Factor type '" + factor.getClass().getSimpleName() + "' not yet supported");
+                throw new UnsupportedOperationException("Only contingency on a branch is yet supported");
```
annetill, line 237: You check contingencies in the very beginning of your code. Do we really need to keep that exception ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,405 @@
...
+                             - states.get(p1.getVariables().get(1).getRow(), element.getGlobalIndex()));
+            double b = states.get(p1.getVariables().get(0).getRow(), sensitivityFactorGroup.getIndex())
+                       - states.get(p1.getVariables().get(1).getRow(), sensitivityFactorGroup.getIndex());
+            element.setAlpha(b / a);
+        } else {
+            DenseMatrix rhs = new DenseMatrix(contingencyElements.size(), 1);
+            DenseMatrix matrix = new DenseMatrix(contingencyElements.size(), contingencyElements.size());
+            for (ComputedContingencyElement element : contingencyElements) {
+                if (!element.getElement().getType().equals(ContingencyElementType.BRANCH)) {
+                    throw new UnsupportedOperationException("Only contingency on a branch is yet supported");
```
annetill, line 164: You check contingencies in the very beginning of your analysis. Do we really need this exception?

src/main/java/com/powsybl/openloadflow/sensi/AbstractDcSensitivityAnalysis.java
```diff
@@ -0,0 +1,368 @@
...
+                connectivity.addVertex(bus);
+            }
+            for (LfBranch branch : network.getBranches()) {
+                connectivity.addEdge(branch.getBus1(), branch.getBus2());
+            }
+            return connectivity;
+        }
+    }
+
+    static class SensitivityFactorWrapped {
```
annetill, line 73: `LfSensitivityFactor` is maybe a better name choice.

src/main/java/com/powsybl/openloadflow/sensi/AbstractDcSensitivityAnalysis.java
```diff
@@ -0,0 +1,368 @@
...
+        private Double functionReference;
+
+        public SensitivityFactorWrapped(SensitivityFactor factor, LfNetwork lfNetwork, EquationSystem equationSystem) {
+            this.factor = factor;
+            if (factor instanceof BranchFlowPerInjectionIncrease) {
+                functionBranch = lfNetwork.getBranchById(((BranchFlowPerInjectionIncrease) factor).getFunction().getBranchId());
+            } else if (factor instanceof  BranchFlowPerPSTAngle) {
+                functionBranch = lfNetwork.getBranchById(((BranchFlowPerPSTAngle) factor).getFunction().getBranchId());
+            } else {
+                throw new UnsupportedOperationException("Only factors of type BranchFlowPerInjectionIncrease and BranchFlowPerPSTAngle are supported");
```
annetill, line 94: You already check that earlier in your code. Do you really need that line that will not be covered?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,405 @@
...
+        // index factors by variable configuration to compute minimal number of states
+        List<SensitivityFactorGroup> factorsGroups = createFactorGroups(network, factorsWrapped);
+
+        if (factorsGroups.isEmpty()) {
+            return Pair.of(Collections.emptyList(), Collections.emptyMap());
         }
 
-        // initialize right hand side
-        DenseMatrix rhs = initRhs(lfNetwork, equationSystem, factorsByVarConfig);
+        // Compute the partipation for every factor
```
annetill, line 294: Improve comment: "// Compute the participation to slack distribution for every factor". I can make some improvements in a dedicated commit if you agree, it will be easier.

Djazouli: Yes sure

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,405 @@
...
+                        element -> getParticipatingElementBus(element).getId(),
+                        element -> -element.getFactor(),
+                        Double::sum
+                    ))
+                ));
+
+                // compute the participation for each injection factor (+1 on the injection, and then -participation on all
+                // elements participating to load balancing, that are still in the connected component of the factor's variable)
+                Function<String, Map<String, Double>> getParticipationForBus = busId -> participationPerCc.get(connectivity.getConnectivity().getComponentNumber(lfNetwork.getBusById(busId)));
+                computeFactorsInjection(getParticipationForBus, factorsGroups);
```
annetill, line 687: Maybe rename it with `participationPerBus ` as earlier in the code.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -7,297 +7,405 @@
...
         JacobianMatrix j = createJacobianMatrix(equationSystem, voltageInitializer);
+        // compute pre-contingency sensitivity values + the states for making +1-1 on the contingencies
+        DenseMatrix rhs = initRhs(lfNetwork, equationSystem, factorsGroups, contingenciesElements);
+        DenseMatrix states = solveTransposed(rhs, j); // states contains angles for the sensitivity factors, but also for the +1-1 related to contingencies
+
+        // sensitivities without contingency
+        List<SensitivityValue> sensitivityValues = calculateSensitivityValues(factorsGroups,
+                states, Collections.emptyList());
+
+        Collection<Contingency> straightforwardContingencies = new LinkedList<>();
```
annetill, line 647: `straightforwardContingencies` should be renamed to reflect that these contingencies do not loose connectivity.

src/test/java/com/powsybl/openloadflow/network/ConnectedComponentNetworkFactory.java
```diff
@@ -0,0 +1,196 @@
+package com.powsybl.openloadflow.network;
```
geofjamg, line 1: License header

src/test/java/com/powsybl/openloadflow/network/ConnectedComponentNetworkFactory.java
```diff
@@ -0,0 +1,196 @@
...
+
+import com.powsybl.iidm.network.Bus;
+import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.network.TwoWindingsTransformer;
+
+public class ConnectedComponentNetworkFactory extends AbstractLoadFlowNetworkFactory {
```
geofjamg, line 7: author

geofjamg (changes requested):

geofjamg (approved):

annetill (approved):

PR #190, opened 2020-12-21
LoadFlowResult management after a DC loadflow

annetill:

No.

The LoadFlowResult from powsybl-core is now filled with metrics.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
@@ -88,6 +88,12 @@ public class OpenLoadFlowProvider implements LoadFlowProvider {
                                prefix + "status", result.getNewtonRaphsonStatus().name());
     }
 
+    private static ImmutableMap<String, String> createMetrics(DcLoadFlowResult result) {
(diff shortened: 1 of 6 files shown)


geofjamg (approved):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -88,12 +88,6 @@ private static AcLoadFlowObserver getObserver(OpenLoadFlowParameters parametersE
                                prefix + "status", result.getNewtonRaphsonStatus().name());
     }
 
-    private static ImmutableMap<String, String> createMetrics(DcLoadFlowResult result) {
```
annetill: If you remove that, it means that we have no metrics at the end of the computation using itools or pypowsybl. The number of iterations is not needed as it means nothing but the status is usefull.

geofjamg: Now there is a way to fill status and iteration on component result these metrics are useless ( and this is true also for ac)

geofjamg: Metrics might still be useful fior ourput infos not yet normmilzed in api output

PR #195, opened 2021-01-12
Increase security analysis with temporary limits and violations filter

annetill:

Yes, issue: https://github.com/powsybl/powsybl-open-loadflow/issues/129

- Expected: current temporary limits.
- Expected: a sort on limit violations. There is a rule not to detect violations in state N-1/K if they already exist in state N, expect if they are worst (violation of a limit that has a smaller acceptable duration). For a contingency and for a branch side, if several temporary limit are violated, only the worst one is displayed (the worst one means the smaller acceptable duration). The `LimitViolationFilter` is not used for the moment.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/LfBranch.java
```diff
@@ -45,6 +47,10 @@
 
     double getPermanentLimit2();
 
+    Map<Integer, Double> getTemporaryLimits1();
```
floriand-e2r, line 50: It's a bit strange to use a map, I think returning a List<TemporaryLimit> would be clearer.

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -205,6 +203,30 @@ public double getPermanentLimit2() {
         return branch.getCurrentLimits2() != null ? branch.getCurrentLimits2().getPermanentLimit() * getBus2().getNominalV() / PerUnit.SB : Double.NaN;
     }
 
+    @Override
+    public Map<Integer, Double> getTemporaryLimits1() {
+        Map<Integer, Double> map = new HashMap<>();
```
floriand-e2r, line 208: This code is duplicated in all classes extending LfBranch, shouldn't we add a `getTemporaryLimits(CurrentLimits currentLimits, double nominalVoltage)` method in `AbstractLfBranch`?

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -205,6 +203,30 @@ public double getPermanentLimit2() {
...
+        if (branch.getCurrentLimits1() != null) {
+            for (CurrentLimits.TemporaryLimit temporaryLimit : branch.getCurrentLimits1().getTemporaryLimits()) {
+                map.put(temporaryLimit.getAcceptableDuration(),
+                        temporaryLimit.getValue() != Double.MAX_VALUE ? temporaryLimit.getValue() * getBus1().getNominalV() / PerUnit.SB : Double.MAX_VALUE);
```
floriand-e2r, line 212: Why is it `Double.MAX_VALUE` here and `Double.NaN` in others?

floriand-e2r (changes requested):

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
```diff
@@ -21,6 +25,9 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractLfBranch.class);
 
+    private static final Comparator<CurrentLimits.TemporaryLimit> TEMPORARY_LIMITS_COMPARATOR =
+        Comparator.comparingDouble(CurrentLimits.TemporaryLimit::getValue).reversed();
```
floriand-e2r, line 29: I'm wondering, isn't it the same to sort based on descending values as to sort based on ascending `acceptableDuration`? Then we wouldn't need to sort, as they are already sorted in `TemporaryLimitImpl` based on ascending acceptableDuration

src/main/java/com/powsybl/openloadflow/network/LfBranch.java
floriand-e2r: I commited with a `SortedSet<CurrentLimits.TemporaryLimit>` instead, as we're only using it sorted anyway. Using this class means that the values are not in PerUnit though.

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
```diff
@@ -39,6 +46,14 @@ protected AbstractLfBranch(LfBus bus1, LfBus bus2, PiModel piModel) {
         this.piModel = Objects.requireNonNull(piModel);
     }
 
+    protected static SortedSet<CurrentLimits.TemporaryLimit> getSortedTemporaryLimits(CurrentLimits currentLimits) {
```
annetill, line 49: Indeed you have change the initial code too much for me. The values of current limits attached to a LfBranch sould be stored in per unit system. This `<CurrentLimits.TemporaryLimit>` should disappear too I think.

annetill: Why all these treatments have disappear: 
`temporaryLimit.getValue() != Double.NaN ? temporaryLimit.getValue() * getBus1().getNominalV() / PerUnit.SB : Double.NaN);`

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -200,20 +200,36 @@ private void detectViolations(Stream<LfBranch> branches, Stream<LfBus> buses, Li
...
-        double scale = 1;
-        if (branch.getBus1() != null && branch.getI1() > branch.getPermanentLimit1()) {
-            scale = PerUnit.SB / branch.getBus1().getNominalV();
-            LimitViolation limitViolation1 = new LimitViolation(branch.getId(), LimitViolationType.CURRENT, (String) null,
-                    2147483647, branch.getPermanentLimit1() * scale, (float) 1., branch.getI1() * scale, Branch.Side.ONE);
-            violations.add(limitViolation1);
+        if (branch.getBus1() != null) {
+            double scale1 = PerUnit.SB / branch.getBus1().getNominalV();
+            // permanent limit detection
+            if (branch.getI1() > branch.getPermanentLimit1()) {
```
annetill, line 206: Here you compare without scale but after for temporary limits you compare with scaling. The coherence should be maintained. All data at `Lf` level are and should be in per unit system.

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
floriand-e2r: Alright. Then we could introduce a new `LfTemporaryLimit` class. Or come back to the map.

annetill:
Now, we have to imagine a post-processing that, for a branch and for s side:
- For the base case: keep the limit violation with the more important value ;
- For the post-contingency case: keep the limit violation with the more important value and put it in the result only if not present in the base case.

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
floriand-e2r: (and the treatments you're referring to have disappeared as it's the ones translating in per unit)

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
```diff
@@ -39,6 +66,17 @@ protected AbstractLfBranch(LfBus bus1, LfBus bus2, PiModel piModel) {
         this.piModel = Objects.requireNonNull(piModel);
     }
 
+    protected static SortedSet<LfTemporaryLimit> createSortedTemporaryLimitsSet(CurrentLimits currentLimits, LfBus bus) {
+        SortedSet<LfTemporaryLimit> temporaryLimits = new TreeSet<>(TEMPORARY_LIMITS_COMPARATOR);
+        if (currentLimits != null) {
+            for (CurrentLimits.TemporaryLimit temporaryLimit : currentLimits.getTemporaryLimits()) {
+                double valuePerUnit = temporaryLimit.getValue() != Double.MAX_VALUE ? temporaryLimit.getValue() * bus.getNominalV() / PerUnit.SB : Double.MAX_VALUE;
```
annetill, line 73: `Double.NaN`?

annetill: Indeed I think that we don't need to make the distinction between temporary and permanent limit. The permanent limit is just a temporary limit with infinite duration time. And with that, the comparator works for all limits and it will ease the post processing on limit violations.

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
```diff
@@ -38,6 +74,19 @@ protected AbstractLfBranch(LfNetwork network, LfBus bus1, LfBus bus2, PiModel pi
         this.piModel = Objects.requireNonNull(piModel);
     }
 
+    protected static List<LfLimit> createSortedLimitsList(CurrentLimits currentLimits, LfBus bus) {
+        LinkedList<LfLimit> sortedLimits = new LinkedList<>();
+        if (currentLimits != null) {
+            double toPerUnit = bus.getNominalV() / PerUnit.SB;
+            for (LoadingLimits.TemporaryLimit temporaryLimit : currentLimits.getTemporaryLimits()) {
+                double valuePerUnit = temporaryLimit.getValue() != Double.MAX_VALUE ? temporaryLimit.getValue() * toPerUnit : Double.MAX_VALUE;
```
annetill, line 82: `Why the default value for limit value is `Double.MAX_VALUE` ?

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -271,21 +288,28 @@ private PostContingencyResult runPostContingencySimulation(LfNetwork network, Ac
...
         if (postContingencyComputationOk) {
             detectViolations(
                 network.getBranches().stream().filter(b -> !lfContingency.getBranches().contains(b)),
                 network.getBuses().stream().filter(b -> !lfContingency.getBuses().contains(b)),
                 postContingencyLimitViolations);
         }
 
+        preContingencyLimitViolations.forEach((subjectSideId, preContingencyViolation) -> {
+            LimitViolation postContingencyViolation = postContingencyLimitViolations.get(subjectSideId);
+            if (postContingencyViolation != null && postContingencyViolation.getAcceptableDuration() >= preContingencyViolation.getAcceptableDuration()) {
```
annetill, line 301: @geofjamg I think that the comparator should rely on getValue and not on getAcceptableDuration. What do you think ?

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
floriand-e2r: Because Double.NaN is not supported as value in the implementation `TemporaryLimitAdder` of powsybl-iidm-impl. Do you think we should still support a NaN value?

geofjamg (approved):

PR #196, opened 2021-01-13
[WIP] Sensitivity API v2

geofjamg:

**Other information**:


src/main/java/com/powsybl/openloadflow/sensi/apiv2/MatrixSensitivityFactor.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    public DenseMatrix getValues() {
+        if (values == null) {
+            values = new DenseMatrix(variablesIds.size(), functionsIds.size());
+        }
+        return values;
+    }
+
+    public double[] getFunctionsReferences() {
```
Djazouli, line 77: It's worth noting that the behavior would be weird if someone did something like
```
matrixSensitivityFactor.getValues();
matrixSensitivityFactor.getFunctionsIds().add("another function");
matrixSensitivityFactor.getValues();
```
But I don't see any use case where this would happen so I guess it is fine

src/main/java/com/powsybl/openloadflow/sensi/apiv2/MatrixSensitivityFactor.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public class MatrixSensitivityFactor {
+
+    private final List<String> functionsIds;
+
+    private final FunctionType functionType;
+
+    private final List<String> variablesIds;
```
Djazouli, line 25: At some point, do we want to be able to have a matrix of MultiVariables ? For example, if you want to compute multiple GLSKs on the same branches. In which case, I guess this List<String> would be replaced by a List<Variable>

src/main/java/com/powsybl/openloadflow/sensi/apiv2/SensitivityFactorConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+                                                                         List<MultiVariablesSensitivityFactor> multiVarsFactors) {
+        Objects.requireNonNull(contingencyId);
+        return new SensitivityFactorConfiguration(ContingencyContext.ONE_CONTINGENCY, contingencyId, simpleFactors, matrixFactors, multiVarsFactors);
+    }
+
+    public static SensitivityFactorConfiguration createForOneContingency(String contingencyId) {
+        return createForOneContingency(contingencyId, Collections.emptyList(), Collections.emptyList(), Collections.emptyList());
+    }
+
+    public static SensitivityFactorConfiguration createForAllContingency(List<SimpleSensitivityFactor> simpleFactors,
```
Djazouli, line 64: I am not sure I understand how this fits with the factors being mutable. From what I have understood, the results of the sensitivity analysis will be written inside the factors. But how do you write multiple values (one for each contingency) ?

src/main/java/com/powsybl/openloadflow/sensi/apiv2/MatrixSensitivityFactor.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private final VariableType variableType;
+
+    private DenseMatrix values;
+
+    private double[] functionsReferences;
+
+    public MatrixSensitivityFactor(List<String> functionsIds, FunctionType functionType, List<String> variablesIds, VariableType variableType) {
```
annetill, line 33: I am not sure to understand the limitations to have multiple `variableType` in a same `MatrixSensitivityFactor` because it would be more efficient to have a unique matrix. I can understand for `functionType`.

Issue #184, opened 2020-12-09
Manage BranchFlowPerLinearGLSK in DC sensitivity analysis


Djazouli:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Currently, we are managing sensitivity analysis for BranchFlowPerInjectionIncrease and BranchFlowPerPSTAngle factors in DC mode. We need still to manage BranchFlowPerLinearGLSK factor.

* **What is the motivation / use case for changing the behavior?**
Being coherent with powsybl-sensitivity-analysis-api

PR #197, opened 2021-01-14
[WIP] Security analysis: controller separated from controlled by contingency

floriand-e2r:
**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem ?**
No (but related to #129)

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
Feature

**What is the current behavior?** *(You can also link to an open issue here)*
No branch tripping for bus breaker view, which is needed for security analysis 

**What is the new behavior (if this is a feature change)?**
Branch tripping for bus breaker view, hence security analysis can be done on bus breaker view networks

**Does this PR introduce a breaking change or deprecate an API?**
No

PR #199, opened 2021-01-18
GLSK support and sensitivity n minus k

Djazouli:

#184 
#186

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
@@ -147,6 +147,10 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis {
         public boolean areVariableAndFunctionDisconnected(GraphDecrementalConnectivity<LfBus> connectivity) {
             throw new NotImplementedException("areVariableAndFunctionDisconnected should have an override");
         }
+
+        public boolean isConnectedToSlack(Integer slackBusComponent, GraphDecrementalConnectivity<LfBus> connectivity) {
+            throw new NotImplementedException("isConnectedToSlack should have an override");
+        }
     }
 
     static class LfBranchFlowPerInjectionIncrease extends LfSensitivityFactor {
@@ -162,6 +166,11 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis {
         public boolean areVariableAndFunctionDisconnected(final GraphDecrementalConnectivity<LfBus> connectivity) {
             return connectivity.getComponentNumber(injectionLfBus) != connectivity.getComponentNumber(getFunctionLfBranch().getBus1())
                     || connectivity.getComponentNumber(injectionLfBus) != connectivity.getComponentNumber(getFunctionLfBranch().getBus2());
+        }
+
+        @Override
+        public boolean isConnectedToSlack(Integer slackBusComponent, GraphDecrementalConnectivity<LfBus> connectivity) {
+            return connectivity.getComponentNumber(injectionLfBus) == slackBusComponent;
         }
 
         public LfBus getInjectionLfBus() {
@@ -183,6 +192,12 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis {
             return connectivity.getComponentNumber(phaseTapChangerLfBranch.getBus1()) != connectivity.getComponentNumber(getFunctionLfBranch().getBus1())
                     || connectivity.getComponentNumber(phaseTapChangerLfBranch.getBus1()) != connectivity.getComponentNumber(getFunctionLfBranch().getBus2());
         }
+
+        @Override
+        public boolean isConnectedToSlack(Integer slackBusComponent, GraphDecrementalConnectivity<LfBus> connectivity) {
+            return slackBusComponent == connectivity.getComponentNumber(phaseTapChangerLfBranch.getBus1());
+        }
+
     }
 
     static class SensitivityFactorGroup {
@@ -267,7 +282,8 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis {
 
     static class ComputedContingencyElement {
 
-        private int contingencyIndex = -1; // local index of the element inside of a contingency : index of the element in the matrix used in the setAlphas method
+        private int contingencyIndex = -1; // index of the element in the rhs for +1-1
+        private int localIndex = -1; // local index of the element : index of the element in the matrix used in the setAlphas method
         private double alpha = Double.NaN;
         private final ContingencyElement element;
         private final LfBranch lfBranch;
@@ -285,6 +301,14 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis {
 
         public void setContingencyIndex(final int index) {
             this.contingencyIndex = index;
+        }
+
+        public int getLocalIndex() {
+            return localIndex;
+        }
+
+        public void setLocalIndex(final int index) {
+            this.localIndex = index;
         }
 
         public double getAlpha() {
@@ -311,6 +335,13 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis {
             int index = 0;
             for (ComputedContingencyElement element : elements) {
                 element.setContingencyIndex(index++);
+            }
+        }
+
+        public static void setLocalIndexes(Collection<ComputedContingencyElement> elements) {
+            int index = 0;
+            for (ComputedContingencyElement element : elements) {
+                element.setLocalIndex(index++);
             }
         }
 
@@ -518,12 +549,13 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis {
             element.setAlpha(b / a);
         } else {
             // FIXME: direct resolution if contingencyElements.size() == 2
+            ComputedContingencyElement.setLocalIndexes(contingencyElements);
             DenseMatrix rhs = new DenseMatrix(contingencyElements.size(), 1);
             DenseMatrix matrix = new DenseMatrix(contingencyElements.size(), contingencyElements.size());
             for (ComputedContingencyElement element : contingencyElements) {
                 LfBranch lfBranch = element.getLfBranch();
                 ClosedBranchSide1DcFlowEquationTerm p1 = element.getLfBranchEquation();
-                rhs.set(element.getContingencyIndex(), 0, factorStates.get(p1.getVariables().get(0).getRow(), sensitivityFactorGroup.getIndex())
+                rhs.set(element.getLocalIndex(), 0, factorStates.get(p1.getVariables().get(0).getRow(), sensitivityFactorGroup.getIndex())
                         - factorStates.get(p1.getVariables().get(1).getRow(), sensitivityFactorGroup.getIndex())
                 );
                 for (ComputedContingencyElement element2 : contingencyElements) {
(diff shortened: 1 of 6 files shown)


geofjamg:
> The fault propagation will certainly be easier to write if we merge #197 first.
> The computation of functionReference will be done in another PR, as well as the HVDC management, so I think we're done with this PR atm

Looking at the code, I am ok with this PR. @annetill will do some functional validation and then we can merge it.

geofjamg (approved):

annetill:
I have perform some functional tests of GLSK support. It seems that the sum of the GLSK must be equal to 1. In case of difference, we have 2 options: throw an exception, or rescale. If we loose a generator and a load after a contingency, for sure we have to rescale in order that the sum equals 1. I will test N-K tomorrow!

annetill (changes requested):

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -265,9 +333,41 @@ void fillRhs(LfNetwork lfNetwork, EquationSystem equationSystem, Matrix rhs) {
...
+            glskMapInMainComponent = glskMap.entrySet().stream().collect(Collectors.toMap(
+                entry -> entry.getKey().getId(),
+                Map.Entry::getValue
+            ));
+        }
+
+        @Override
+        public void setInjectionByBus(final Map<String, Double> participationToSlackByBus) {
+            Double glskWeightSum = glskMapInMainComponent.values().stream().mapToDouble(Math::abs).sum();
+            glskMapInMainComponent.forEach((busId, weight) -> participationToSlackByBus.merge(busId, weight / glskWeightSum, Double::sum));
```
Djazouli, line 354: @annetill the rescaling to 1 here is not working ?

annetill: Indeed you are right. The GLSK with this map:
`        glskMap.put("d2", 30f);
        glskMap.put("g2", 10f);
        glskMap.put("d3", 50f);
        glskMap.put("g1", 10f);`
should give the same result as:
        `glskMap.put("d2", 0.3f);
        glskMap.put("g2", 0.1f);
        glskMap.put("d3", 0.5f);
        glskMap.put("g1", 0.1f);`
So maybe, just add a warnning or something that says that the initial map is not well filled. Maybe change the test too (and for sure the IIDM documentation is missing).

annetill: But is it used after a contingency that looses an element of the GLSK ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -713,22 +905,33 @@ private void detectConnectivityLoss(LfNetwork lfNetwork, DenseMatrix states, Lis
...
+                for (LfSensitivityFactor factor : lfFactors) {
+                    // check if the factor function and variable are in different connected components
+                    if (factor.areVariableAndFunctionDisconnected(connectivity.getConnectivity())) {
+                        factor.setPredefinedResult(0d);
+                    } else if (!factor.isConnectedToSlack(slackBusComponent, connectivity.getConnectivity())) {
+                        factor.setPredefinedResult(Double.NaN);
+                    }
+                }
+
+                rescaleGlsk(factorGroups, connectivity.getConnectivity(), slackBusComponent);
```
Djazouli, line 926: @annetill here, the call to this function changes the variable glskMapInMainComponent from the glsk factor group, and then, it will change the output of setInjectionByBus when it is called by computeInjectionFactors line 951

annetill (approved):

jonenst:
I forcepushed a squashed version of this pr

annetill:
Thanks Jon, indeed I think that @Djazouli makes the merged by mistake. I am going to review the rights.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -697,11 +888,12 @@ private void detectConnectivityLoss(LfNetwork lfNetwork, DenseMatrix states, Lis
...
             // connectivity analysis by connectivity
-            // we will index contingencies by a list of branch associations that breaks connectivity
+            // we will index contingencies by a list of branch that may breaks connectivity
             // for example, if in the network, loosing line L1 breaks connectivity, and loosing L2 and L3 together breaks connectivity,
-            // the associations would be: [{L1}, {L2, L3}]
+            // the index would be: {L1, L2, L3}
+            // todo: There may be a better way to group the contingencies that will have the same connectivities afterwards
             Collection<Contingency> nonLosingConnectivityContingencies = new LinkedList<>();
-            Map<List<Set<ComputedContingencyElement>>, List<Contingency>> contingenciesByGroupOfElementsBreakingConnectivity = new HashMap<>();
+            Map<Set<ComputedContingencyElement>, List<Contingency>> contingenciesByGroupOfElementsBreakingConnectivity = new HashMap<>();
```
floriand-e2r, line 896: I'm a bit confused with the map here, as we don't need the indexing. Just a list of pairs would be ok (or a list of other custom objects). But anyway that doesn't come from this PR, it was already like that before.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -608,20 +756,62 @@ protected DenseMatrix initContingencyRhs(LfNetwork lfNetwork, EquationSystem equ
...
         for (Contingency contingency : contingencies) {
-            List<Set<ComputedContingencyElement>> groupOfElementsBreakingConnectivity = getGroupOfElementsBreakingConnectivity(lfNetwork, states,
+            Set<ComputedContingencyElement> groupOfElementsBreakingConnectivity = getGroupOfElementsBreakingConnectivity(lfNetwork, states,
                     contingency.getElements().stream().map(element -> contingenciesElements.get(element.getId())).collect(Collectors.toList()), equationSystem);
             if (groupOfElementsBreakingConnectivity.isEmpty()) { // connectivity not broken
                 nonLosingConnectivityContingencies.add(contingency);
-            } else if (groupOfElementsBreakingConnectivity.stream().mapToInt(Set::size).max().getAsInt() > 1) {
-                throw new PowsyblException("The contingency " + contingency.getId() + " breaks the connectivity on more than one branch.");
             } else {
                 contingenciesByGroupOfElementsBreakingConnectivity.computeIfAbsent(groupOfElementsBreakingConnectivity, key -> new LinkedList<>()).add(contingency);
```
floriand-e2r, line 766: From another PR, but a `Set.put()` would be clearer (always absent).

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -608,20 +756,62 @@ protected DenseMatrix initContingencyRhs(LfNetwork lfNetwork, EquationSystem equ
...
             }
         }
     }
 
+    private Set<String> getElementsToReconnect(GraphDecrementalConnectivity<LfBus> connectivity, Set<ComputedContingencyElement> breakingConnectivityCandidates) {
+        Set<String> elementsToReconnect = new HashSet<>();
+
+        Map<Pair<Integer, Integer>, Set<ComputedContingencyElement>> elementsByConnectedComponents =
+                breakingConnectivityCandidates.stream().collect(Collectors.toMap(
+                    element -> Pair.of(connectivity.getComponentNumber(element.getLfBranch().getBus1()), connectivity.getComponentNumber(element.getLfBranch().getBus2())),
```
floriand-e2r, line 776: be careful, Pair.equals() takes into account the order, hence you can have in the map (2, 3) and (3, 2) as keys

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -608,20 +756,62 @@ protected DenseMatrix initContingencyRhs(LfNetwork lfNetwork, EquationSystem equ
...
+                        return existing;
+                    }
+                ));
+
+        Map<Integer, Set<Integer>> connections = new HashMap<>();
+        connections.put(0, Collections.singleton(0));
+        for (Set<LfBus> smallComponent : connectivity.getSmallComponents()) {
+            int ccNumber = connectivity.getComponentNumber(smallComponent.iterator().next());
+            connections.put(ccNumber, new HashSet<>(Collections.singletonList(ccNumber)));
+        }
```
floriand-e2r, line 789: It would be clearer to replace the loop with a `for (int i=0; i<nbComponents; i++) connections.put(i, Collections.singleton(i));`

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -608,20 +756,62 @@ protected DenseMatrix initContingencyRhs(LfNetwork lfNetwork, EquationSystem equ
...
+            Integer cc2 = elementsByCc.getKey().getValue();
+            if (connections.get(cc1).contains(cc2)) {
+                // cc are already connected
+                continue;
+            }
+            elementsToReconnect.add(elementsByCc.getValue().iterator().next().getElement().getId());
+            Set<Integer> newCc = new HashSet<>();
+            newCc.addAll(connections.get(cc1));
+            newCc.addAll(connections.get(cc2));
+            connections.get(cc1).forEach(integer -> connections.put(integer, newCc));
```
floriand-e2r, line 802: Why is it only for integers in `connections.get(cc1)`? Shouldn't it be `newCc.forEach()` instead?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -608,20 +756,62 @@ protected DenseMatrix initContingencyRhs(LfNetwork lfNetwork, EquationSystem equ
...
+        }
+
+        for (Map.Entry<Pair<Integer, Integer>, Set<ComputedContingencyElement>> elementsByCc : elementsByConnectedComponents.entrySet()) {
+            Integer cc1 = elementsByCc.getKey().getKey();
+            Integer cc2 = elementsByCc.getKey().getValue();
+            if (connections.get(cc1).contains(cc2)) {
+                // cc are already connected
+                continue;
+            }
+            elementsToReconnect.add(elementsByCc.getValue().iterator().next().getElement().getId());
```
floriand-e2r, line 798: Could you just add a comment saying why you take only the first element?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -608,20 +756,62 @@ protected DenseMatrix initContingencyRhs(LfNetwork lfNetwork, EquationSystem equ
...
 
+    private Set<String> getElementsToReconnect(GraphDecrementalConnectivity<LfBus> connectivity, Set<ComputedContingencyElement> breakingConnectivityCandidates) {
+        Set<String> elementsToReconnect = new HashSet<>();
+
+        Map<Pair<Integer, Integer>, Set<ComputedContingencyElement>> elementsByConnectedComponents =
+                breakingConnectivityCandidates.stream().collect(Collectors.toMap(
+                    element -> Pair.of(connectivity.getComponentNumber(element.getLfBranch().getBus1()), connectivity.getComponentNumber(element.getLfBranch().getBus2())),
+                    element -> new HashSet<>(Collections.singletonList(element)),
+                    (existing, replacement) -> {
+                        existing.addAll(replacement);
```
floriand-e2r, line 779: In the next lines we only consider the first element. So why should it be a `Set<ComputedContingencyElement>` and not just a `ComputedContingencyElement`?

src/test/java/com/powsybl/openloadflow/network/ConnectedComponentNetworkFactory.java
```diff
@@ -14,6 +14,18 @@
  * @author Gael Macherel <gael.macherel at artelys.com>
  */
 public class ConnectedComponentNetworkFactory extends AbstractLoadFlowNetworkFactory {
+
+    /**
+     * b1 ----------+
```
floriand-e2r, line 19: Could you include the `<PRE>` `</PRE>` tags around the network drawings? so that it's displayed properly in the javadoc

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -608,20 +756,62 @@ protected DenseMatrix initContingencyRhs(LfNetwork lfNetwork, EquationSystem equ
...
                 nonLosingConnectivityContingencies.add(contingency);
-            } else if (groupOfElementsBreakingConnectivity.stream().mapToInt(Set::size).max().getAsInt() > 1) {
-                throw new PowsyblException("The contingency " + contingency.getId() + " breaks the connectivity on more than one branch.");
             } else {
                 contingenciesByGroupOfElementsBreakingConnectivity.computeIfAbsent(groupOfElementsBreakingConnectivity, key -> new LinkedList<>()).add(contingency);
             }
         }
     }
 
+    private Set<String> getElementsToReconnect(GraphDecrementalConnectivity<LfBus> connectivity, Set<ComputedContingencyElement> breakingConnectivityCandidates) {
```
floriand-e2r, line 771: I'm wondering if `GraphDecrementalConnectivity` is the appropriate interface... as we see in this method that the interface is missing some features. We should either modify that interface, or create a new one, or use a standard library like jgrapT (which takes more time to compute the connectivity though)

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: Yes, indeed, I wasn't sure that the connectivity guaranteed the component number to always stay between 0 and the number of components.
But reading at the documentation in the interface, this is the case, so I will change this

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: It shouldn't always be absent
The very purpose of this function is to index contingencies by group of elements that breaks connectivity.
For example, if you have two contingencies removing the following lines: {L1, L2}, and {L1, L3}, and the line L1 breaks the connectivity: then groupOfElementsBreakingConnectivity should be {L1} for both contingencies

floriand-e2r: Thanks for explaining. My mistake was due to the fact that I had in mind that Set.equals() was not checking more than the object equality, but in fact it's doing more than that. A test with the example you're describing would have avoided my mistake! :) But again, this is not the purpose of this PR anyway.

PR #204, opened 2021-01-25
BranchTripping for bus breaker view security analysis

floriand-e2r:

No (but related to #129)

Feature

No branch tripping for bus breaker view, which is needed for security analysis 

Branch tripping for bus breaker view, hence security analysis can be done on bus breaker view networks

No


src/main/java/com/powsybl/openloadflow/sa/BusBreakerTraverser.java
```diff
@@ -0,0 +1,42 @@
...
+        // so to keep things simple we do not propagate the fault
+        if (connected) {
+            traversedTerminals.add(terminal);
+        }
+        return false;
+    }
+
+    @Override
+    public boolean traverse(Switch aSwitch) {
+        throw new AssertionError("Should not be called");
```
geofjamg, line 39: It could if for instance we start from a busbarsection?

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -172,11 +169,13 @@ SecurityAnalysisResult runSync(SecurityAnalysisParameters securityAnalysisParame
         String tmpVariantId = "olf-tmp-" + UUID.randomUUID().toString();
         network.getVariantManager().cloneVariant(network.getVariantManager().getWorkingVariantId(), tmpVariantId);
         try {
-            for (Switch sw : network.getSwitches()) {
-                sw.setRetained(false);
-            }
-            for (Switch sw : allSwitchesToOpen) {
-                sw.setRetained(true);
+            if (network.getVoltageLevelStream().anyMatch(vl -> vl.getTopologyKind() == TopologyKind.NODE_BREAKER)) {
```
geofjamg, line 172: So it means that we don't support hybrid networks with node/breaker and bus/breaker voltage levels?

geofjamg: Not an issue but in that case I will check it at startup and throws an exception in such a case, because here I feel like a propagation between mix of node/breaker and bus/breaker would not work.

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
floriand-e2r: I added support of hybrid networks

src/main/java/com/powsybl/openloadflow/sa/BusBreakerTraverser.java
floriand-e2r: No, we always stop before arriving to a switch!

floriand-e2r:
See https://sonarcloud.io/dashboard?id=com.powsybl%3Apowsybl-open-loadflow&pullRequest=204 for sonar cloud results

geofjamg (approved):

Issue #205, opened 2021-01-26
Recompute function references when doing sensitivity analysis with contingencies


Djazouli:
* **Do you want to request a *feature* or report a *bug*?**
This is a bug.

* **What is the current behavior?**
Currently, the function reference of each sensitivity value returned by the dc sensitivity analysis is the value of the flow in the original network.

* **What is the expected behavior?**
It needs to be the value of the flow for the network where the contingency has been taken into account.

PR #206, opened 2021-01-26
[WIP] Compute function references from within sensitivity analysis

Djazouli:

- [?] Docs have been added / updated (for bug fixes / features)

#205


src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -366,6 +367,17 @@ public double getActivePowerMismatch() {
         return -mismatch;
     }
 
+    public double getActivePowerMismatchWithinMainComponent(GraphDecrementalConnectivity<LfBus> connectivity) {
```
annetill, line 370: I prefer `getActivePowerMismatchWithinMainComponent`.

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
```diff
@@ -26,6 +26,10 @@
 
         String getElementType();
 
+        void setOriginalValues(List<ParticipatingElement> participatingElements);
```
annetill, line 29: Can you hide this setter in a constructor?

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
```diff
@@ -26,6 +26,10 @@
...
+
+        void resetOriginalValues();
```
annetill, line 31: Maybe `restoreDistributionKeys` ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -90,6 +64,8 @@ public LazyConnectivity(LfNetwork lfNetwork) {
 
         private Double functionReference;
 
+        private Double baseSensitivityValue = Double.NaN; // the sensitivity value without any +1-1 (needs to be recomputed if the stack distribution changes)
```
annetill, line 67: I prefer `baseCaseSensitivityValue`.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -368,7 +352,8 @@ public void setGlskMapInMainComponent(final Map<String, Double> glskMapInMainCom
 
         private int contingencyIndex = -1; // index of the element in the rhs for +1-1
         private int localIndex = -1; // local index of the element : index of the element in the matrix used in the setAlphas method
-        private double alpha = Double.NaN;
+        private double alphaSensi = Double.NaN;
```
annetill, line 355: To be coherent, `alphaForSensitivityValue` and `alphaForFunctionReference` ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -431,18 +424,104 @@ public static void setLocalIndexes(Collection<ComputedContingencyElement> elemen
 
     }
 
+    static class LfContingency {
```
annetill, line 427: This class already exists in the code, it is a bit confusing to have twice.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -431,18 +424,104 @@ public static void setLocalIndexes(Collection<ComputedContingencyElement> elemen
...
+        private final Contingency contingency;
+        private final Collection<ComputedContingencyElement> elements;
+
+        public LfContingency(Contingency contingency, Map<ContingencyElement, Collection<ComputedContingencyElement>> defaultPropagation) {
```
annetill, line 431: `defaultPropagation` is not well chosen. Maybe you should introduce something close to `LfBranchTripping` world.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -451,6 +530,17 @@ protected void fillRhsSensitivityVariable(LfNetwork lfNetwork, EquationSystem eq
         }
     }
 
+    private void setPredefinedResults(Collection<LfSensitivityFactor> lfFactors, GraphDecrementalConnectivity<LfBus> connectivity, int slackBusComponent) {
```
annetill, line 533: `slackBusComponent` -> `mainComponent`

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -488,19 +578,12 @@ protected void fillRhsSensitivityVariable(LfNetwork lfNetwork, EquationSystem eq
         return new ArrayList<>(groupIndexedById.values());
     }
 
-    protected void computeInjectionFactors(Function<String, Map<String, Double>> getSlackParticipationByBus, List<SensitivityFactorGroup> factorGroups) {
+    protected void computeInjectionFactors(Map<String, Double> participationMap, List<SensitivityFactorGroup> factorGroups) {
```
annetill, line 581: `participationFactorByBus`

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -617,52 +694,105 @@ public void checkContingencies(LfNetwork lfNetwork, List<Contingency> contingenc
...
-                value = value + contingencyElement.getAlpha() * p1.calculate(contingenciesStates, contingencyElement.getContingencyIndex());
+                double contingencySensitivity = p1.calculate(contingenciesStates, contingencyElement.getContingencyIndex());
+                flowValue += contingencyElement.getAlphaFlow() * contingencySensitivity;
+                sensiValue +=  contingencyElement.getAlphaSensi() * contingencySensitivity;
+            }
+        }
+        return new SensitivityValue(factor.getFactor(), sensiValue * PerUnit.SB, flowValue * PerUnit.SB, 0);
+    }
+
+    protected void setBaseSensitivityValues(List<SensitivityFactorGroup> factorGroups, DenseMatrix factorsState) {
```
annetill, line 724: `setBaseSensitivityValues`

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -617,52 +694,105 @@ public void checkContingencies(LfNetwork lfNetwork, List<Contingency> contingenc
...
             for (LfSensitivityFactor factor : factorGroup.getFactors()) {
-                sensitivityValuesContingencies.add(createBranchSensitivityValue(factor, factorGroup, factorStates, contingenciesStates, contingencyElements));
+                sensitivityValuesContingencies.add(createBranchSensitivityValue(factor, contingenciesStates, contingencyElements));
             }
         }
         return sensitivityValuesContingencies;
     }
 
-    private void setAlphas(List<ComputedContingencyElement> contingencyElements, SensitivityFactorGroup sensitivityFactorGroup,
+    private void setAlphasFlow(Collection<ComputedContingencyElement> contingencyElements,
```
annetill, line 745: Can you try to factorize this function with `setAlphaSensi` for maintainability?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -768,6 +897,17 @@ private void detectConnectivityLoss(LfNetwork lfNetwork, DenseMatrix states, Lis
         }
     }
 
+    private Map<ContingencyElement, Collection<ComputedContingencyElement>> getPropagatedDefaults(Collection<ContingencyElement> contingencyElements, LfNetwork lfNetwork, EquationSystem equationSystem) {
```
annetill, line 900: See above, the method name is not clear enough to be understandable.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -855,112 +992,112 @@ public void cutConnectivity(LfNetwork lfNetwork, GraphDecrementalConnectivity<Lf
...
+
+                // some elements of the GLSK may not be in the connected component anymore, we recompute the injections
+                rescaleGlsk(factorGroups, connectivity, slackBusComponent);
+
+                // null and unused if slack is not distributed
+                List<ParticipatingElement> participatingElementsForThisConnectivity = participatingElements;
+
+                // we need to recompute the factor states because the connectivity changed
+                if (lfParameters.isDistributedSlack() || hasGlsk) {
+                    Map<String, Double> slackParticipation;
```
annetill, line 1067: I think that we have chosen `slackParticipationByBus`?

annetill (changes requested):

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -535,22 +615,31 @@ protected LfBus getParticipatingElementLfBus(ParticipatingElement participatingE
...
+            case PROPORTIONAL_TO_GENERATION_P_MAX:
+                return new GenerationActionPowerDistributionStep();
+            case PROPORTIONAL_TO_LOAD:
+                return new LoadActivePowerDistributionStep(false, false);
+            default:
+                throw new UnsupportedOperationException("Balance type not yet supported: " + loadFlowParameters.getBalanceType());
+        }
+    }
 
+    private ActivePowerDistribution.Step getStep(LoadFlowParameters loadFlowParameters, List<ParticipatingElement> participatingElements) {
```
annetill, line 629: We have to move this function elsewhere.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -535,22 +615,31 @@ protected LfBus getParticipatingElementLfBus(ParticipatingElement participatingE
...
-                step = new LoadActivePowerDistributionStep(false, false);
-                break;
+                return new LoadActivePowerDistributionStep(false, false, participatingElements);
             default:
                 throw new UnsupportedOperationException("Balance type not yet supported: " + loadFlowParameters.getBalanceType());
         }
+    }
 
-        participatingElements = step.getParticipatingElements(lfNetwork);
+    protected List<ParticipatingElement> getParticipatingElements(LfNetwork lfNetwork, LoadFlowParameters loadFlowParameters) {
```
annetill, line 640: We have to move this function elsewhere.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -535,22 +615,31 @@ protected LfBus getParticipatingElementLfBus(ParticipatingElement participatingE
         }
     }
 
-    protected List<ParticipatingElement> getParticipatingElements(LfNetwork lfNetwork, LoadFlowParameters loadFlowParameters) {
-        ActivePowerDistribution.Step step;
-        List<ParticipatingElement> participatingElements;
+    private ActivePowerDistribution.Step getStep(LoadFlowParameters loadFlowParameters) {
```
annetill, line 618: We have to move this function elsewhere.

annetill (changes requested):

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -431,18 +425,104 @@ public static void setLocalIndexes(Collection<ComputedContingencyElement> elemen
...
+
+        equationSystem.updateEquations(dx);
+        equationSystem.updateNetwork(dx);
+
+        // set all calculated voltages to NaN
+        for (LfBus bus : network.getBuses()) {
+            bus.setV(Double.NaN);
+        }
         for (LfSensitivityFactor factor : factors) {
-            factor.setFunctionReference(dcLoadFlowResult.getNetwork().getBranchById(factor.getFunctionLfBranchId()).getP1() * PerUnit.SB);
```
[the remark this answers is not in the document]
Djazouli: I just moved it to the createBranchSensitivityValue function, so that it stays coherent with the rest of the code

geofjamg:
> Because in DC approximation, the power factor does not exist for example. And we have not implemented the slack distribution on conform load (and we will not in the short term) in sensitivity analysis.

But this could be solved by a pre-check on sensi DC side and keep the utility class generic

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
```diff
@@ -94,6 +100,22 @@ public static ActivePowerDistribution create(LoadFlowParameters.BalanceType bala
             default:
                 throw new UnsupportedOperationException("Unknown balance type mode: " + balanceType);
         }
-        return new ActivePowerDistribution(step);
+        return step;
+    }
+
+    public static Step getStep(LoadFlowParameters.BalanceType balanceType) {
```
geofjamg, line 106: This is a very dangerous design. How someone is using it by calling getStep wan now it is only for DC purpose?

src/main/java/com/powsybl/openloadflow/network/util/GenerationActionPowerDistributionStep.java
```diff
@@ -23,6 +25,29 @@
...
 
+    Map<LfGenerator, Double> initialValues = new HashMap<>();
+
+    public GenerationActionPowerDistributionStep() {
+        super();
+    }
+
+    public GenerationActionPowerDistributionStep(List<ParticipatingElement> elementsToMemorize) {
+        this();
+        initialValues = elementsToMemorize.stream()
```
geofjamg, line 36: For me this is a bad design, this utility class should not have a state

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
annetill: Ok, we will follow your advice.

src/main/java/com/powsybl/openloadflow/network/util/GenerationActionPowerDistributionStep.java
annetill: We have started to work on the network to avoid that: https://github.com/powsybl/powsybl-open-loadflow/pull/211
This aim is to have the same mechanism for security analysis and for sensitivity analysis. Please just tell me if we can follow this solution or not.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -453,7 +453,7 @@ private void reactivateEquations(List<Equation> deactivatedEquations, List<Equat
      * @param buses the bus for which the state is returned
      * @return the map of the states of given buses, indexed by the bus itself
      */
-    private Map<LfBus, BusState> getBusStates(List<LfBus> buses) {
```
annetill, line 456: We have to move these three methods in `BusState` class ? @floriand-e2r

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -435,14 +429,90 @@ public DcSensitivityAnalysis(MatrixFactory matrixFactory) {
         super(matrixFactory);
     }
 
-    protected void setFunctionReferenceOnFactors(List<LfNetwork> lfNetworks, List<LfSensitivityFactor> factors,
-                                                 LoadFlowParameters lfParameters, OpenLoadFlowParameters lfParametersExt) {
-        DcLoadFlowParameters dcLoadFlowParameters = new DcLoadFlowParameters(lfParametersExt.getSlackBusSelector(), matrixFactory,
-                true, lfParametersExt.isDcUseTransformerRatio(), lfParameters.isDistributedSlack(),  lfParameters.getBalanceType());
-        DcLoadFlowResult dcLoadFlowResult = new DcLoadFlowEngine(lfNetworks, dcLoadFlowParameters).run();
+    private GraphDecrementalConnectivity<LfBus> createConnectivity(LfNetwork lfNetwork) {
```
annetill, line 432: @floriand-e2r Can you move this function to a network util class ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -435,14 +429,90 @@ public DcSensitivityAnalysis(MatrixFactory matrixFactory) {
...
+    }
+
+    protected DenseMatrix setFunctionReferenceOnFactors(LfNetwork network, EquationSystem equationSystem, LUDecomposition lu, List<LfSensitivityFactor> factors,
+                                                 LoadFlowParameters lfParameters, List<ParticipatingElement> participatingElements, GraphDecrementalConnectivity<LfBus> connectivity) {
+
+        double[] x = equationSystem.createStateVector(new UniformValueVoltageInitializer());
+        Map<LfBus, BusState> busStates = new HashMap<>();
+        if (lfParameters.isDistributedSlack()) {
+            double mismatch;
+            if (connectivity != null) {
```
annetill, line 450: @floriand-e2r we have to think about making this "common".

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -435,14 +429,90 @@ public DcSensitivityAnalysis(MatrixFactory matrixFactory) {
...
+                mismatch = network.getActivePowerMismatchWithinMainComponent(connectivity);
+            } else {
+                mismatch = network.getActivePowerMismatch();
+            }
+            busStates = OpenSecurityAnalysis.getBusStates(participatingElements.stream()
+                                                                               .map(ParticipatingElement::getLfBus)
+                                                                               .collect(Collectors.toSet()));
+            int iteration = 0;
+            List<ParticipatingElement> participatingElementsCopy = new ArrayList<>(participatingElements);
```
annetill, line 459: Why do you still need a copy of participating elements?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -521,38 +595,15 @@ protected void rescaleGlsk(List<SensitivityFactorGroup> factorGroups, GraphDecre
...
-                step = new LoadActivePowerDistributionStep(false, false);
-                break;
-            default:
-                throw new UnsupportedOperationException("Balance type not yet supported: " + loadFlowParameters.getBalanceType());
-        }
-
-        participatingElements = step.getParticipatingElements(lfNetwork);
-        ParticipatingElement.normalizeParticipationFactors(participatingElements, "bus");
-        return participatingElements;
+        return getParticipatingElements(lfNetwork, loadFlowParameters, element -> true);
```
annetill, line 606: I am not sure to get the aim of this function...

Djazouli: It allows you to get the normalized participating elements

annetill: Indeed I get it, it is to filter.

Djazouli: It is possible to use the isDisabled attribute, but it less efficient. 
To use the isDisabled parameter, I think we would need to:
- Save all the buses states
- Go through all the buses and disconnect the one that needs to be disconnected
- Go through all the buses to get the non-disconnected participating elements
- Restore all the buses states

With the current filtering in the DC getParticipatingElements, we are just doing:
- Go through all the buses to get participating elements and get the one in the right connected component

annetill: I was thinking about save the state of the bus that initial contains a participating element. Then, you disabled those that are not in the main connected component of this connectivity. Then, you can use the classical getParticipatingElements. Then you restore the buses that initially contain participating element. It is just to ensure coherence in the code, but it is not a bit deal.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
floriand-e2r: I've put it in LfNetwork directly (similarly to `createZeroImpedanceSubGraph()`)

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -435,14 +432,79 @@ public DcSensitivityAnalysis(MatrixFactory matrixFactory) {
...
+
+        equationSystem.updateEquations(x);
+
+        double[] dx = equationSystem.createTargetVector();
+
+        if (connectivity != null) {
+            // set buses injections and transformers to 0 outside the main connected component
+            int mainComponentNumber = connectivity.getComponentNumber(network.getSlackBus());
+            List<Integer> columnsToSetToZero = network.getBuses().stream()
+                    .filter(lfBus -> connectivity.getComponentNumber(lfBus) != mainComponentNumber)
```
floriand-e2r, line 468: To get the buses which are not in the main component you should use directly `connectivity.getSmallComponents()`

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -822,145 +883,142 @@ public void cutConnectivity(LfNetwork lfNetwork, GraphDecrementalConnectivity<Lf
...
+
             // compute the contingencies with loss of connectivity
             for (Map.Entry<Set<ComputedContingencyElement>, List<Contingency>> entry : contingenciesByGroupOfElementsBreakingConnectivity.entrySet()) {
                 Set<ComputedContingencyElement> breakingConnectivityCandidates = entry.getKey();
                 List<Contingency> contingencyList = entry.getValue();
-
                 lfFactors.forEach(factor -> factor.setPredefinedResult(null));
-                cutConnectivity(lfNetwork, connectivity.getConnectivity(), breakingConnectivityCandidates);
+                cutConnectivity(lfNetwork, connectivity, breakingConnectivityCandidates);
+                int mainComponent = connectivity.getComponentNumber(lfNetwork.getSlackBus());
```
floriand-e2r, line 980: Always equal to 0

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -451,6 +513,17 @@ protected void fillRhsSensitivityVariable(LfNetwork lfNetwork, EquationSystem eq
         }
     }
 
+    private void setPredefinedResults(Collection<LfSensitivityFactor> lfFactors, GraphDecrementalConnectivity<LfBus> connectivity, int mainComponent) {
+        for (LfSensitivityFactor factor : lfFactors) {
+            // check if the factor function and variable are in different connected components
+            if (factor.areVariableAndFunctionDisconnected(connectivity)) {
+                factor.setPredefinedResult(0d);
+            } else if (!factor.isConnectedToComponent(mainComponent, connectivity)) {
```
floriand-e2r, line 521: You always call `isConnectedToComponent` with mainComponent, so in fact you could rename it to `isConnectedToMaintComponent` and use `getSmallComponents()` instead of `getComponentNumber()`

floriand-e2r (changes requested):
I first thought that you needed a change in `GraphDecrementalConnectivity` to simplify your connectivity code, but in fact you can use `getSmallComponents()` instead of `getComponentNumber()` almost everywhere, if not everywhere. There is two places where I'm not so sure about that:
- `areVariableAndFunctionDisconnected()`: do you really need to know if variable and function are in two distinct "small" connected components?
- `getElementsToReconnect()`: I don't think we can avoid using `getComponentNumber()`, do you agree on that?
Why am I trying to make those changes? Because in implementations of decremental connectivity the getComponentNumber needs some additional time to compute the map giving the number from the vertex.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: Unfortunately, I think getSmallComponents() returns the smallest components in size, but it has nothing to do with the slack bus (the slack could be in a small component). What I want here are the buses that are not in the same component than the slack bus.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: See previous comment, I am not sure about this

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Djazouli: See previous conversation

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
floriand-e2r: Indeed, but couldn't we make the assumption that the slack bus is in the biggest component in size? I thought you could, but well I might be mistaken - calling @annetill and @geofjamg to the rescue!

PR #209, opened 2021-01-31
[WIP] Fix ratio + phase tap changer branch model

geofjamg:

No

Bug fix

When a transformer has a non regulating ratio tap changer + phase tap changer, the resulting combination of ratio and phase is not correct. 

When a transformer has a non regulating ratio tap changer + phase tap changer, a correct combination of ratio and phase is calculated. 

**Other information**:


annetill:
We have to keep in mind that it would lead to a difference bewt

PR #210, opened 2021-02-01
Refactor transformer conversion

geofjamg:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/Transformers.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/Transformers.java
@@ -274,9 +274,9 @@ public final class Transformers {
     /**
      * Find the tap position of a ratio tap changer corresponding to a given rho shift.
      */
-    public static int findTapPosition(RatioTapChanger rtc, double rho) {
+    public static int findTapPosition(RatioTapChanger rtc, double ptcRho, double rho) {
         for (int position = rtc.getLowTapPosition(); position <= rtc.getHighTapPosition(); position++) {
-            if (Math.abs(rho - rtc.getStep(position).getRho()) < EPS_ALPHA) {
+            if (Math.abs(rho - ptcRho * rtc.getStep(position).getRho()) < EPS_ALPHA) {
                 return position;
             }
         }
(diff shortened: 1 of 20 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/Transformers.java
```diff
@@ -25,238 +23,153 @@
...
+                angle0, r0, x0, g0, b0);
     }
 
-    /**
-     * Get the nominal magnetizing susceptance of a three windings transformer, located on bus star side.
-     */
-    public static double getB1(ThreeWindingsTransformer.Leg leg, boolean twtSplitShuntAdmittance) {
-        return getB1(leg, getCurrentPosition(leg.getRatioTapChanger()), getCurrentPosition(leg.getPhaseTapChanger()), twtSplitShuntAdmittance);
+    public static Integer getCurrentPosition(RatioTapChanger rtc) {
+        return rtc != null ? rtc.getTapPosition() : null;
```
annetill, line 128: I am not sure to understand why I have initialy chosen `null` if there is not ratio tap changer.

src/main/java/com/powsybl/openloadflow/network/impl/Transformers.java
```diff
@@ -25,238 +23,153 @@
...
     }
 
-    public static double getB1(ThreeWindingsTransformer.Leg leg, Integer rtcPosition, Integer ptcPosition, boolean twtSplitShuntAdmittance) {
-        return getB1(twtSplitShuntAdmittance ? leg.getB() / 2 : leg.getB(), leg.getRatioTapChanger(), leg.getPhaseTapChanger(), rtcPosition, ptcPosition);
+    public static Integer getCurrentPosition(PhaseTapChanger rtc) {
+        return rtc != null ? rtc.getTapPosition() : 0;
```
annetill, line 132: `ptc` should be better. I am not sure to understand why I have initialy chosen `0` when we have no phase tap changer.

src/main/java/com/powsybl/openloadflow/network/impl/Transformers.java
geofjamg: It was to not confuse with a possible value

src/main/java/com/powsybl/openloadflow/network/impl/Transformers.java
geofjamg: I renamed to ptc and replace 0 by null to be consistent with the other method.

annetill (approved):
No regression on functional tests.

PR #213, opened 2021-02-02
Fix DC equations when a1 is a variable

geofjamg:

No

Bug fix

When a1 is considered as a variable the rhs, so the part of the equation which has to be moved to right hand side is not correct.

rhs has been fixed

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/dc/equations/ClosedBranchSide1DcFlowEquationTerm.java
+++ b/src/main/java/com/powsybl/openloadflow/dc/equations/ClosedBranchSide1DcFlowEquationTerm.java
@@ -47,7 +47,7 @@ public final class ClosedBranchSide1DcFlowEquationTerm extends AbstractClosedBra
         Objects.requireNonNull(x);
         double ph1 = x.get(ph1Var.getRow(), column);
         double ph2 = x.get(ph2Var.getRow(), column);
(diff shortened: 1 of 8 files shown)


Djazouli (approved):
I tried it on the data I was having issues with, and it seems to be working, so it sounds good to me !

src/main/java/com/powsybl/openloadflow/dc/equations/ClosedBranchSide1DcFlowEquationTerm.java
```diff
@@ -58,7 +58,11 @@ public void update(double[] x) {
         double ph2 = x[ph2Var.getRow()];
         double a1 = a1Var != null ? x[a1Var.getRow()] : branch.getPiModel().getA1();
         p1 = calculate(ph1, ph2, a1);
-        rhs = -power * (A2 - a1);
+        if (a1Var == null) {
```
annetill, line 61: In `ac` module, you have introduced the notion of active variable. I think that we have to test if `(a1Var != null && a1Var.isActive())`, as we do in `ac` module.

geofjamg: Done

annetill (approved):

PR #217, opened 2021-02-08
Refactor dc sensivity analysis

floriand-e2r:

No

Refactor

1. Common code in DcLoadFlowEngine and DcSensitivityAnalysis
2. Redundant calls to connectivity to compute the list of disconnected LfBuses or its complement (and therefore maniplulating the GraphDecrementalConnectivity interface in many methods).

No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
+++ b/src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
@@ -32,7 +32,7 @@ public class DcLoadFlowEngine {
 
     private final DcLoadFlowParameters parameters;
 
-    private double[] dx;
+    private double[] targetVector;
 
     public DcLoadFlowEngine(LfNetwork network, MatrixFactory matrixFactory) {
         this.networks = Collections.singletonList(network);
@@ -97,9 +97,7 @@ public class DcLoadFlowEngine {
 
         equationSystem.updateEquations(x);
 
-        double[] targets = equationSystem.createTargetVector();
-
-        this.dx = Arrays.copyOf(targets, targets.length);
+        this.targetVector = equationSystem.createTargetVector();
 
         if (!removedBuses.isEmpty()) {
             // set buses injections and transformers to 0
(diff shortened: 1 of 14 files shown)


Djazouli (approved):
I haven't checked the implementation of the new methods in the connectivity API, but I trust you
The refactoring looks really great ! I hope that at some point, we may have another way of computing "predefinedResults" than using component numbers within the Sensitivity Analysis, but this would be for another time.

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -59,10 +55,18 @@ public DcLoadFlowEngine(List<LfNetwork> networks, DcLoadFlowParameters parameter
...
-    private void distributeSlack(LfNetwork network) {
-        double mismatch = network.getActivePowerMismatch();
+    protected void distributeSlack(Collection<LfBus> buses) {
+        double mismatch = getActivePowerMismatch(buses);
         ActivePowerDistribution activePowerDistribution = ActivePowerDistribution.create(parameters.getBalanceType(), false);
-        activePowerDistribution.run(network, mismatch);
+        activePowerDistribution.run(buses, mismatch);
+    }
+
+    public static double getActivePowerMismatch(Collection<LfBus> buses) {
```
annetill, line 64: I don't like the idea of having as argument a list of buses because before it was a network that represents a unique synchronous component and now you hide that. Do you have to protect/private the functions that use this list as argument.

floriand-e2r:
> I hope that at some point, we may have another way of computing "predefinedResults" than using component numbers within the Sensitivity Analysis, but this would be for another time.

Thanks, I just forgot that point! I think I can improve this a bit at least.

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -71,45 +75,73 @@ public DcLoadFlowResult run() {
...
-            equationSystem.updateEquations(dx);
-            equationSystem.updateNetwork(dx);
+        this.targetVector = equationSystem.createTargetVector();
 
-            // set all calculated voltages to NaN
-            for (LfBus bus : network.getBuses()) {
-                bus.setV(Double.NaN);
-            }
+        if (!removedBuses.isEmpty()) {
+            // set buses injections and transformers to 0
```
Djazouli, line 114: Losing a transformer in a contingency requires us to do
equationSystem.getEquation(transformer.getNum(), EquationType.BRANCH_ALPHA1).get().getColumn() and set this column to 0.
So we may also need to pass a list of removedTransformers

Djazouli: (I'll do it in another PR and ask for a merge on this one)

Djazouli: (see #237 )

geofjamg (approved):

PR #228, opened 2021-02-18
Remove observer

geofjamg:

**Other information**:


floriand-e2r (approved):
Introducing a `Profiler` instead of the existing `Observer` interface makes things neater and clearer! But I'm wondering, how interesting is it to have our own profiler instead of using an external one?

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -36,37 +40,39 @@
...
         this.equationSystem = Objects.requireNonNull(equationSystem);
         this.j = Objects.requireNonNull(j);
         this.stoppingCriteria = Objects.requireNonNull(stoppingCriteria);
     }
 
-    private NewtonRaphsonStatus runIteration(double[] fx, double[] targets, double[] x) {
-        observer.beginIteration(iteration);
+    public void logLargestMismatches(double[] mismatch, EquationSystem equationSystem, int count) {
+        if (LOGGER.isTraceEnabled()) {
+            Map<Equation, Double> mismatchByEquation = new HashMap<>(equationSystem.getSortedEquationsToSolve().size());
```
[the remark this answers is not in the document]
geofjamg: Done

geofjamg:
> Introducing a `Profiler` instead of the existing `Observer` interface makes things neater and clearer! But I'm wondering, how interesting is it to have our own profiler instead of using an external one?

Agree, even if more synthetic and clear that what we have with a generic profiler, maybe it does not worth the cost of adding specific code everywhere . I removed to profiler and this PR now only contains observer removal (which was not useful and was responsible of code clutterring)

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -36,37 +37,33 @@
...
         this.j = Objects.requireNonNull(j);
         this.stoppingCriteria = Objects.requireNonNull(stoppingCriteria);
     }
 
+    public void logLargestMismatches(double[] mismatch, EquationSystem equationSystem, int count) {
+        if (LOGGER.isTraceEnabled()) {
+            equationSystem.getSortedEquationsToSolve().keySet().stream()
+                    .map(equation -> Pair.of(equation, mismatch[equation.getColumn()]))
+                    .filter(e -> Math.abs(e.getValue()) > Math.pow(10, -7))
+                    .sorted(Comparator.comparingDouble((Map.Entry<Equation, Double> e) -> Math.abs(e.getValue())).reversed())
```
floriand-e2r, line 54: Strange, it should be `Pair<Equation, Double>` instead of `Map.Entry<Equation, Double>`

floriand-e2r (approved):

src/main/java/com/powsybl/openloadflow/equations/JacobianMatrix.java
```diff
@@ -185,8 +186,9 @@ public Matrix getMatrix() {
     }
 
     private LUDecomposition getLUDecomposition() {
+        Matrix matrix = getMatrix();
```
floriand-e2r, line 189: I think this is not needed anymore

geofjamg: it is, this is a serious bug that was hidden by a log that was calling luckily getMatrix before computation

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
geofjamg: I don't understand, Pair has the same methods that Entry.

src/main/java/com/powsybl/openloadflow/equations/JacobianMatrix.java
floriand-e2r: ok, my bad!

PR #236, opened 2021-03-02
Change disconnected sensitivities management

Djazouli:

Do not throw an error if we are asking a sensitivity on a component that is not in the network

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
@@ -9,10 +9,7 @@ package com.powsybl.openloadflow.sensi;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.contingency.ContingencyElement;
 import com.powsybl.contingency.ContingencyElementType;
-import com.powsybl.iidm.network.Bus;
-import com.powsybl.iidm.network.Injection;
-import com.powsybl.iidm.network.Network;
-import com.powsybl.iidm.network.TwoWindingsTransformer;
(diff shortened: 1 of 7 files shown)


src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -98,6 +104,12 @@ protected JacobianMatrix createJacobianMatrix(EquationSystem equationSystem, Vol
     }
 
     static class LfSensitivityFactor<T extends EquationTerm> {
+
+        enum Status {
+            VALID,
+            SKIP,
+            NULL
```
geofjamg, line 111: You don't need this enum value, just use a null status

Djazouli: For me, null and Status.NULL would represent two different things:
- If the status is Status.NULL, it means that the value for this sensitivity factor will be null (0), because it is a sensitivity on a disconnected branch
- if the status is null (meaning: the sensitivity factor has no value), then this is a bug

geofjamg: So here NULL means zero. this make me think it is NULL as null in Java so an enum with a mix of business and technical values which is a bad practice

geofjamg:
@Djazouli I just tested, it works very well, but it is quite verbose. When a transformer is disconnected and is the variable of many factors we end up with as many warn as functions for this variables. Maybe we can in this skip have just one warning saying: "skipping all factors with variable ..." ?

geofjamg (approved):

PR #244, opened 2021-03-16
Manage additional factors

Djazouli:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
@@ -25,6 +25,7 @@ import com.powsybl.openloadflow.util.BusState;
 import com.powsybl.openloadflow.util.LfContingency;
 import com.powsybl.openloadflow.util.PropagatedContingency;
 import com.powsybl.sensitivity.SensitivityFactor;
+import com.powsybl.sensitivity.SensitivityFactorsProvider;
 import com.powsybl.sensitivity.SensitivityValue;
 import com.powsybl.sensitivity.factors.BranchIntensityPerPSTAngle;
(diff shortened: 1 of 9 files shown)


src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -112,37 +115,57 @@ protected void setFunctionReferences(List<LfSensitivityFactor> factors) {
...
         // create LF network (we only manage main connected component)
         List<LfNetwork> lfNetworks = LfNetwork.load(network, new LfNetworkParameters(lfParametersExt.getSlackBusSelector(), false, true, lfParameters.isTwtSplitShuntAdmittance(), false, lfParametersExt.getPlausibleActivePowerLimit(), false));
         LfNetwork lfNetwork = lfNetworks.get(0);
         checkContingencies(lfNetwork, contingencies);
-        checkSensitivities(network, factors);
         checkLoadFlowParameters(lfParameters);
+
+        List<SensitivityFactor> factors = factorsProvider.getCommonFactors(network);
+        checkSensitivities(network, factors);
+        List<SensitivityFactor> additionalFactors = factorsProvider.getAdditionalFactors(network);
```
geofjamg, line 160: The SensitivityFactoryProvider API is terrible, I add to read to JavaDoc to figure out that this is factors that have to be calculated for only base case.... In our code maybe we should be more clear by naming like factorsOnBaseCase or something like this?

geofjamg (changes requested):
I have difficulties to understand the code. The 3 ugly concepts that are in the API should not be impact our internal implementation. It should only be a preprocessing.
Everywhere in implementation we should have only one list of factor and a list of contingency for which we want to compute the sensitivity (could be empty)
The case where we only want to compute post contingency sensitivity and not base case one should not be handle (we compute base case sensitivity in any case).

PR #247, opened 2021-03-19
Remove variable activation

geofjamg:

No

Bug fix

Variable has a active property which allows a variable considered either as a variable or as a constant in an equation term. The problem is that it is not fully implemented (see the ugly FIXME in Variable.java). By luck when we change the active status of a variable, we always also change the active status of its term, so it works but this is really fragile...

That would be hard to support a robust variable active status. Instead it is simpler to just rework the equation design by just adding a new equation v = constant when we want the variable to become constant. 

**Other information**:


geofjamg:
@annetill @floriand-e2r this is the comeback a  change I revert in the network listener PR :-). I added a description/explanation of the change in the PR, tell me if it is clear or not.

floriand-e2r (approved):
more robust and clearer indeed!

annetill (approved):

annetill (approved):

PR #248, opened 2021-03-20
Refactor network and equation link

geofjamg:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
@@ -207,10 +207,17 @@ public final class AcEquationSystem {
 
             // add a dummy reactive power variable to both sides of the non impedant branch and with an opposite sign
             // to ensure we have the same number of equation and variables
-            equationSystem.createEquation(bus1.getNum(), EquationType.BUS_Q)
-                    .addTerm(EquationTerm.createVariableTerm(branch, VariableType.DUMMY_Q, variableSet));
-            equationSystem.createEquation(bus2.getNum(), EquationType.BUS_Q)
-                    .addTerm(EquationTerm.multiply(EquationTerm.createVariableTerm(branch, VariableType.DUMMY_Q, variableSet), -1));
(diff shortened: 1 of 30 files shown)


geofjamg:
@Djazouli I added to this PR a direct link between bus and injection equation, coud you review and check this could avoid CacheBus in your PR #239 ?

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -295,25 +302,41 @@ private static void createImpedantBranch(LfBranch branch, LfBus bus1, LfBus bus2
...
+            sp1.addTerm(p1);
             branch.setP1(p1);
             if (creationParameters.isPhaseControl()) {
                 createBranchActivePowerTargetEquation(branch, DiscretePhaseControl.ControlledSide.ONE, equationSystem, p1);
             }
         }
         if (q1 != null) {
-            equationSystem.createEquation(bus1.getNum(), EquationType.BUS_Q).addTerm(q1);
+            Equation sq1 = equationSystem.createEquation(bus1.getNum(), EquationType.BUS_Q);
+            if (sq1.getTerms().isEmpty()) {
```
Djazouli, line 317: Not sure I fully understand the purpose of these three lines, is it just a performance optimization, or do they have some deeper implications ?

Djazouli (approved):

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
geofjamg: This is just to avoid setting injection equation for each of the branch connected to the bus

annetill: So I understand that our code only works for networks with at least one branch? I have never tested but can we have a network with only one generator and one load, connected in the same "bus"?

annetill (approved):
I think that it is okay for branch refactoring. For bus refactoring, as it is not used, we have to be aware in the future. The first test will be to base #239 on that PR. @Djazouli if you find issue, we will make fixes in your PR.

PR #253, opened 2021-03-23
Introduce voltage sensitivity

Djazouli:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
@@ -215,9 +215,10 @@ public final class AcEquationSystem {
             PiModel piModel = branch.getPiModel();
             double rho = PiModel.R2 / piModel.getR1();
             EquationTerm vTerm = EquationTerm.createVariableTerm(bus1, VariableType.BUS_V, variableSet, bus1.getV().eval());
+            EquationTerm bus2vTerm = EquationTerm.createVariableTerm(bus2, VariableType.BUS_V, variableSet, bus2.getV().eval());
             equationSystem.createEquation(branch.getNum(), EquationType.ZERO_V)
                     .addTerm(vTerm)
-                    .addTerm(EquationTerm.multiply(EquationTerm.createVariableTerm(bus2, VariableType.BUS_V, variableSet), -1 * rho));
+                    .addTerm(EquationTerm.multiply(bus2vTerm, -1 * rho));
             bus1.setV(vTerm);
             // add a dummy reactive power variable to both sides of the non impedant branch and with an opposite sign
             // to ensure we have the same number of equation and variables
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
@@ -48,9 +48,10 @@ public final class AcEquationSystem {
             Equation v = equationSystem.createEquation(bus.getNum(), EquationType.BUS_V);
             if (v.getTerms().isEmpty()) {
                 v.setActive(false);
-                EquationTerm vTerm = EquationTerm.createVariableTerm(bus, VariableType.BUS_V, variableSet);
+                EquationTerm vTerm = EquationTerm.createVariableTerm(bus, VariableType.BUS_V, variableSet, bus.getV().eval());
                 v.addTerm(vTerm);
-                v.setUpdateType(EquationSystem.EquationUpdateType.NEVER);
+                bus.setV(vTerm);
+                v.setUpdateType(EquationSystem.EquationUpdateType.AFTER_NR);
             }
         }
     }
@@ -58,7 +59,9 @@ public final class AcEquationSystem {
     private static void createVoltageControlEquations(VoltageControl voltageControl, LfBus bus, VariableSet variableSet,
                                                       EquationSystem equationSystem, AcEquationSystemCreationParameters creationParameters) {
         if (voltageControl.isVoltageControlLocal()) {
-            equationSystem.createEquation(bus.getNum(), EquationType.BUS_V).addTerm(EquationTerm.createVariableTerm(bus, VariableType.BUS_V, variableSet));
+            EquationTerm vTerm = EquationTerm.createVariableTerm(bus, VariableType.BUS_V, variableSet, bus.getV().eval());
+            equationSystem.createEquation(bus.getNum(), EquationType.BUS_V).addTerm(vTerm);
+            bus.setV(vTerm);
         } else if (bus.isVoltageControlled()) {
             // remote controlled: set voltage equation on this controlled bus
             createVoltageControlledBusEquations(voltageControl, equationSystem, variableSet, creationParameters);
@@ -82,8 +85,10 @@ public final class AcEquationSystem {
         LfBus controlledBus = voltageControl.getControlledBus();
 
         // create voltage equation at voltage controlled bus
+        EquationTerm vTerm = EquationTerm.createVariableTerm(controlledBus, VariableType.BUS_V, variableSet, controlledBus.getV().eval());
         Equation vEq = equationSystem.createEquation(controlledBus.getNum(), EquationType.BUS_V)
-                .addTerm(EquationTerm.createVariableTerm(controlledBus, VariableType.BUS_V, variableSet));
+                .addTerm(vTerm);
+        controlledBus.setV(vTerm);
 
         List<LfBus> controllerBuses = voltageControl.getControllerBuses().stream()
                 .filter(LfBus::isVoltageControllerEnabled)
@@ -209,10 +214,11 @@ public final class AcEquationSystem {
             // 0 = v1 - v2 * rho
             PiModel piModel = branch.getPiModel();
             double rho = PiModel.R2 / piModel.getR1();
+            EquationTerm vTerm = EquationTerm.createVariableTerm(bus1, VariableType.BUS_V, variableSet, bus1.getV().eval());
             equationSystem.createEquation(branch.getNum(), EquationType.ZERO_V)
-                    .addTerm(EquationTerm.createVariableTerm(bus1, VariableType.BUS_V, variableSet))
+                    .addTerm(vTerm)
                     .addTerm(EquationTerm.multiply(EquationTerm.createVariableTerm(bus2, VariableType.BUS_V, variableSet), -1 * rho));
-
+            bus1.setV(vTerm);
             // add a dummy reactive power variable to both sides of the non impedant branch and with an opposite sign
             // to ensure we have the same number of equation and variables
             Equation sq1 = equationSystem.createEquation(bus1.getNum(), EquationType.BUS_Q);
@@ -257,7 +263,9 @@ public final class AcEquationSystem {
 
     private static void createDiscreteVoltageControlEquation(LfBus bus,  VariableSet variableSet, EquationSystem equationSystem) {
         if (bus.isDiscreteVoltageControlled()) {
-            equationSystem.createEquation(bus.getNum(), EquationType.BUS_V).addTerm(EquationTerm.createVariableTerm(bus, VariableType.BUS_V, variableSet));
+            EquationTerm vTerm = EquationTerm.createVariableTerm(bus, VariableType.BUS_V, variableSet, bus.getV().eval());
+            equationSystem.createEquation(bus.getNum(), EquationType.BUS_V).addTerm(vTerm);
+            bus.setV(vTerm);
             if (bus.getDiscreteVoltageControl().getControllers().size() > 1) {
                 createR1DistributionEquations(equationSystem, variableSet, bus.getDiscreteVoltageControl().getControllers());
             }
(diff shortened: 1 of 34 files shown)


Djazouli:
We will need to update the core version once the sensitivity is released on core's side

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -55,6 +54,27 @@ protected AbstractSensitivityAnalysis(MatrixFactory matrixFactory) {
         this.matrixFactory = Objects.requireNonNull(matrixFactory);
     }
 
+    protected static Terminal getEquipmentRegulatingTerminal(Network network, String equipmentId) {
```
Djazouli, line 57: @annetill Here I am trying to get the terminal that corresponds to the equipmentId, but I am not sure of the way to do it

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -125,44 +147,62 @@ protected JacobianMatrix createJacobianMatrix(EquationSystem equationSystem, Vol
...
             } else if (factor instanceof BranchFlowPerLinearGlsk) {
                 BranchFlow branchFlow = ((BranchFlowPerLinearGlsk) factor).getFunction();
                 functionLfBranch = lfNetwork.getBranchById(branchFlow.getBranchId());
                 equationTerm = functionLfBranch != null ? (EquationTerm) functionLfBranch.getP1() : null;
+                functionLfBus = null;
+            } else if (factor instanceof BusVoltagePerTargetV) {
+                functionLfBranch = null;
+                Bus bus = ((BusVoltagePerTargetV) factor).getFunction().getBusRef().resolve(network).orElseThrow();
+                functionLfBus = lfNetwork.getBusById(bus.getId());
+                equationTerm = equationSystem.getEquationTerm(ElementType.BUS, functionLfBus.getNum(), EquationTerm.VariableEquationTerm.class);
```
Djazouli, line 176: @geofjamg @annetill I would like to do the same thing as the branches (like using functionLfBus.getV() to get an Evaluable), but I am not sure if the Evaluable should be the bus_V equation, or the bus_V equation term ? Moreover, currently, getV() returns v / nominalV, so I am not sure whether the nominal V should be included in the Evaluable or not

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
annetill: I have pushed a commit to help you (but the code could be improved if you want). Indeed it was what I have tried to avoid with another design of the factor, without any success from my colleagues :-)

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
geofjamg: You can use the BUS_V equation term.
No the / nominalV should not be included and just be careful to add it when updating the IIDM network (LfBusImpl.updateState method)

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -111,7 +146,9 @@ protected JacobianMatrix createJacobianMatrix(EquationSystem equationSystem, Vol
         // Wrap factors in specific class to have instant access to their branch, and their equation term
         private final SensitivityFactor factor;
 
-        private final LfBranch functionLfBranch;
+        private final LfBranch functionLfBranch; // todo: create two classes: branch function and bus function ?
```
geofjamg, line 149: Ideally yes, if possible we should have another class with commons abstract for branch or bus monitored factor.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -345,6 +404,40 @@ public boolean isConnectedToComponent(Set<LfBus> connectedComponent) {
         }
     }
 
+    static class LfBusVoltagePerTargetV extends LfSensitivityFactor {
+
+        private final LfBus targetVBus;
+
+        LfBusVoltagePerTargetV(SensitivityFactor factor, Network network, LfNetwork lfNetwork, EquationSystem equationSystem) {
+            super(factor, network, lfNetwork, equationSystem);
+            String targetVTerminalId = ((TargetV) factor.getVariable()).getEquipmentId();
```
geofjamg, line 413: This is not a terminal id, but an element id

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -345,6 +404,40 @@ public boolean isConnectedToComponent(Set<LfBus> connectedComponent) {
...
+            Terminal regulatingTerminal = getEquipmentRegulatingTerminal(network, targetVTerminalId);
```
geofjamg, line 414: The factor/bus ref is crappy, this why you are doing all this ugly code. Here factor should provide you a TerminaRef directly.

geofjamg (changes requested):

geofjamg:
> @Djazouli Now that we have a sensi internal API, I propose that for this PR, you don't care about powsybl sensi API and just make voltage sensi available from internal API. In Gridpy I made an evolution to directly connect to sensi internal API, so even without new powsybl API addition I will be able to use it.

The idea would be to merge this, before switching to next powsybl release. Adaptation to new factor from powsybl api would be done in another PR.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -66,11 +172,9 @@ protected JacobianMatrix createJacobianMatrix(EquationSystem equationSystem, Vol
         // Wrap factors in specific class to have instant access to their branch and their equation term
         private final Object context;
 
-        private final String functionId;
-
         private final String variableId;
 
-        protected final LfBranch functionLfBranch;
+        ILfSensitivityFunction function;
```
geofjamg, line 177: LfBranch and LfBus has a common base class LfElement. Would it be possible to use it here in order to avoid adding ILfSensitivityFunction?

geofjamg: Maybe we could try in this PR to adapt LfFactor to internal API. To avoid LfFactor subclassing hell (with all function/variable combinations), we could try to only have one LfFactor class and one LfFactorGroup. So having LfElement type for variable and function.

Djazouli: So, we would to store a LfElement + a SensitivityFunctionType instead of having a class wrapping the element for each Function type ? And then, to get the equation term from that, we write a big switch statement, instead of overriding methods ? (and do the same thing for the variables)

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -49,13 +47,52 @@ protected AbstractSensitivityAnalysis(MatrixFactory matrixFactory, Supplier<Grap
...
+    }
+
     protected JacobianMatrix createJacobianMatrix(EquationSystem equationSystem, VoltageInitializer voltageInitializer) {
         double[] x = equationSystem.createStateVector(voltageInitializer);
         equationSystem.updateEquations(x);
         return new JacobianMatrix(equationSystem, matrixFactory);
     }
 
-    static class LfSensitivityFactor {
+    abstract static class AbstractLfSensitivityFactor {
```
geofjamg, line 95: Thanks, for this great refactoring! Maybe we can keep a LfSensitivityFactor interface in addition to the AbstractLfSensitivityFactor to avoid using AbstractLfSensitivityFactor everywhere in the code (Abstract class are not supposed to be used as an API)?

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -49,217 +47,294 @@ protected AbstractSensitivityAnalysis(MatrixFactory matrixFactory, Supplier<Grap
...
+        @Override
+        public SensitivityFunctionType getFunctionType() {
+            return functionType;
         }
 
+        @Override
         public EquationTerm getEquationTerm() {
-            throw new NotImplementedException("getEquationTerm should have an override");
+            switch (functionType) {
+                case BRANCH_ACTIVE_POWER:
```
annetill, line 198: Just a small remark here: maybe it could be better to use the terminology we already introduced: `BRANCH_P`, `BRANCH_I` (I have it on a working branch for transformer in current limiter mode) and `BUS_V`?

geofjamg: I would keep it like this as it is supposed one day to be move to core as API V2 and as an officieal API is it more meaningful to keep  BRANCH_ACTIVE_POWER

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -49,217 +47,294 @@ protected AbstractSensitivityAnalysis(MatrixFactory matrixFactory, Supplier<Grap
...
+    protected static Terminal getEquipmentRegulatingTerminal(Network network, String equipmentId) {
+        Generator generator = network.getGenerator(equipmentId);
+        if (generator != null) {
+            return generator.getRegulatingTerminal();
+        }
+        StaticVarCompensator staticVarCompensator = network.getStaticVarCompensator(equipmentId);
+        if (staticVarCompensator != null) {
+            return staticVarCompensator.getRegulatingTerminal();
+        }
+        TwoWindingsTransformer t2wt = network.getTwoWindingsTransformer(equipmentId);
```
annetill, line 59: This PR is indeed not working with transformers as we don't have any PV bus linked to the transformer regulation. In your test, the PV bus is linked to the generator. For phase tap changers, we have activated the phase variable. Here, we have to active the rho variable and the linked bus_v equation I think.

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -263,7 +263,7 @@ private static void createBranchActivePowerTargetEquation(LfBranch branch, Discr
 
     private static void createDiscreteVoltageControlEquation(LfBus bus,  VariableSet variableSet, EquationSystem equationSystem) {
         if (bus.isDiscreteVoltageControlled()) {
-            EquationTerm vTerm = EquationTerm.createVariableTerm(bus, VariableType.BUS_V, variableSet, bus.getV().eval());
+            EquationTerm vTerm = EquationTerm.createVariableTerm(bus, VariableType.BUS_V, variableSet, bus.getDiscreteVoltageControl().getTargetValue());
```
annetill, line 266: @geofjamg I think it is ` bus.getV().eval()`

geofjamg: Using a numeric value of a state variable in the equation system is not correct.
bus.getDiscreteVoltageControl().getTargetValue() is better, but not correct too. It should be redesign to move it in the RHS.

geofjamg: After thinking again, I am not sure to understand why do we need to set an initial value for equation term at this place. Normally equation initial value should be computed from initial state vector. To investigate before merging this PR.

annetill: Yes indeed I understand now the purpose of this new term and we have to investigate why we need it now.

Djazouli: The initial value is only used if the voltage initializer type is set to `PREVIOUS_VALUES`. (because it causes a call to lfBus.getV().eval() ). Otherwise it is indeed replaced immediately by the value from the state vector.

geofjamg: But with PREVIOUS_VALUES the bus voltage is initialized from IIDM values (if defined). Why do you need to override this value?

Djazouli: The PREVIOUS_VALUES was initialized with LfBus.getV() (now lfBus.getV().eval()). The issue is that after the creation of the system, lfBus.getV() returns the EquationTerm representing V in the system, and not the value found in the IIDM, so we need to pass the value from the IIDM to the EquationTerm so it does not get lost. (or we could also try to call the voltage initialization before the creation of the equation system)

geofjamg: Ok I understand, so @annetill is right in her comment, it should be bus.getV().eval() in that case to really init from IIDM network state as it was before.

annetill (approved):

src/main/java/com/powsybl/openloadflow/equations/EquationTerm.java
```diff
@@ -165,21 +166,30 @@ public double rhs() {
...
+        }
+
         @Override
         public void write(Writer writer) throws IOException {
             variables.get(0).write(writer);
         }
     }
 
     static VariableEquationTerm createVariableTerm(LfElement element, VariableType variableType, VariableSet variableSet) {
+        return createVariableTerm(element, variableType, variableSet, 0d);
```
annetill, line 181: I think it will be better to put `Double.NaN` here.

annetill (approved):

PR #260, opened 2021-03-28
Sensi internal api

geofjamg:

**Other information**:


src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -282,45 +214,50 @@ public boolean isConnectedToComponent(Set<LfBus> connectedComponent) {
     }
 
     static class LfBranchFlowPerPSTAngle extends LfBranchPerPSTAngle {
```
geofjamg, line 216: Ideally we should also refactor LfFactor subclasses to reflect new more generic API

src/main/java/com/powsybl/openloadflow/sensi/SensitivityFactorReaderAdapter.java
```diff
@@ -0,0 +1,89 @@
...
+        this.contingencies = Objects.requireNonNull(contingencies);
+    }
+
+    @Override
+    public void read(Handler handler) {
+        for (SensitivityFactor factor : sensitivityFactorsProvider.getCommonFactors(network)) {
+            read(handler, factor);
+        }
+        // FIXME additional factors are not yet supported
+        if (!sensitivityFactorsProvider.getAdditionalFactors(network).isEmpty()) {
```
Djazouli, line 51: How do you think additional factors should be managed ? Do we pass additional info inside the context, or do we need to change the handler's methods ?

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
Djazouli: Yep, this would be great, do you want to do it within this PR, or another one ?

Djazouli (approved):

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
geofjamg: This PR is already very large, I think we should do it in another PR. Any idea how to do it?

src/main/java/com/powsybl/openloadflow/sensi/SensitivityFactorReaderAdapter.java
geofjamg: We can add contingency ID to handler

PR #261, opened 2021-03-29
Fix default parameter loading

mathbagu:

N/A

Bug fix

When we run OLF without specific parameters, the parameters are not read from the config.yml file. The default constructor is used.

Load the OLF parameters from config.yml.
The default value for `throwsExceptionInCaseOfSlackDistributionFailure` also changed to `false`. @geofjamg do you agree with that change?

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/util/ParameterConstants.java
+++ b/src/main/java/com/powsybl/openloadflow/util/ParameterConstants.java
@@ -21,7 +21,7 @@ public final class ParameterConstants {
     public static final SlackBusSelector SLACK_BUS_SELECTOR_DEFAULT_VALUE = new MostMeshedSlackBusSelector();
 
     public static final String THROWS_EXCEPTION_IN_CASE_OF_SLACK_DISTRIBUTION_FAILURE_PARAM_NAME = "throwsExceptionInCaseOfSlackDistributionFailure";
-    public static final boolean THROWS_EXCEPTION_IN_CASE_OF_SLACK_DISTRIBUTION_FAILURE_DEFAULT_VALUE = true;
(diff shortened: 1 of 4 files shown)


src/main/java/com/powsybl/openloadflow/util/ParameterConstants.java
```diff
@@ -21,7 +21,7 @@
     public static final SlackBusSelector SLACK_BUS_SELECTOR_DEFAULT_VALUE = new MostMeshedSlackBusSelector();
 
     public static final String THROWS_EXCEPTION_IN_CASE_OF_SLACK_DISTRIBUTION_FAILURE_PARAM_NAME = "throwsExceptionInCaseOfSlackDistributionFailure";
-    public static final boolean THROWS_EXCEPTION_IN_CASE_OF_SLACK_DISTRIBUTION_FAILURE_DEFAULT_VALUE = true;
+    public static final boolean THROWS_EXCEPTION_IN_CASE_OF_SLACK_DISTRIBUTION_FAILURE_DEFAULT_VALUE = false;
```
geofjamg, line 24: I don't known if it should be the defautl behaviour. I think we did that because of network with multiple synchronous components. Very often there is no way to fully distribute slack on small synchronous components and we don't want the simulation to fail

annetill: Yes but this modification, the default behavior is to log an error.

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/util/ParameterConstants.java
geofjamg: Ok

annetill (approved):

PR #262, opened 2021-03-30
Target vector update refactoring

geofjamg:

**Other information**:


src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -102,7 +102,7 @@ public DcLoadFlowResult run(Reporter reporter) {
 
         equationSystem.updateEquations(x);
 
-        this.targetVector = equationSystem.createTargetVector();
+        this.targetVector = TargetVector.createArray(network, equationSystem);
```
flo-dup, line 105: Shouldn't we do something similar for DC? later maybe. Indeed, I think that the target vector update could be used for sensitivity analyses, where we could avoid to create the vector several times.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -283,7 +297,14 @@ public double getLoadTargetQ() {
 
     @Override
     public void setLoadTargetQ(double loadTargetQ) {
-        this.loadTargetQ = loadTargetQ * PerUnit.SB;
+        double newLoadTargetQ = loadTargetQ * PerUnit.SB;
+        if (newLoadTargetQ != this.loadTargetQ) {
```
flo-dup, line 301: I realized this old/new test is only there because of restoring all bus states instead of only those which have changed, maybe I should have a look there if there's something to do. and if it's worth it.

flo-dup (approved):
Nice exending the use of network listener. Just a couple of comments

src/main/java/com/powsybl/openloadflow/equations/TargetVector.java
```diff
@@ -0,0 +1,108 @@
...
+        // nothing to do
+    }
+
+    @Override
+    public void onStateUpdate(double[] x) {
+        // nothing to do
+    }
+
+    public double[] toArray() {
+        if (status == Status.VALID) {
```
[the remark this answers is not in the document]
geofjamg: Yes, done.

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
geofjamg: Maybe it could be interesting but not as much as for AC power flow because for DC power we have just either one or two iteration depending on stack distribution activated or not. For AC power we have often a lot of iterations and target vector number updates, that is why it is in this case very interesting to automate the update.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
geofjamg: Indeed we could probably check if restoration is relevant on BusState/BranchState side but in any case I think it is safer to add this guard here.

PR #265, opened 2021-04-07
manage additional factors

Djazouli:

**Other information**:


geofjamg (commented):
Looks good to me, the way it is implemented is very clear and simple.
I am just wondering if there is a better design for ContingencyContext. Marker interfaces are often not considered as a good practice, but not easy to do otherwise.

src/main/java/com/powsybl/openloadflow/sensi/SensitivityFactorReader.java
```diff
@@ -13,11 +13,29 @@
  */
 public interface SensitivityFactorReader {
 
+    interface ContingencyContext { }
```
flo-dup, line 16: Coudn't we replace this marker interface by something like:
```
enum ContingencyContextType {
   ALL, NONE, SPECIFIC
}
class ContingencyContext {
   private final String contingencyId;
   private final ContingencyContextType contextType;
}
```
Not the prettiest design either I know, but at least this avoids casting and instance of.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -675,12 +676,45 @@ private static void checkRegulatingTerminal(Network network, String equipmentId)
...
+        private final List<LfSensitivityFactor> commonFactors = new ArrayList<>();
+
+        public List<LfSensitivityFactor> getAllFactors() {
+            List<LfSensitivityFactor> allFactors = new ArrayList<>(commonFactors);
+            allFactors.addAll(additionalFactorsNoContingency);
+            allFactors.addAll(additionalFactorsPerContingency.values().stream().flatMap(List::stream).collect(Collectors.toSet()));
+            return allFactors;
+        }
+
+        public List<LfSensitivityFactor> getFactorsForContingency(String contingencyId) {
```
flo-dup, line 691: Just a detail, but I think you could avoid passing null arguments to a function to have a specific behaviour, and instead call a specific method directly. This avoids a test in the method but also avoids to give a sense to the null value (here I admit that the sense is easily guessed though!).

flo-dup (commented):
Neat implementation - as already said above! Just a couple of remarks

src/main/java/com/powsybl/openloadflow/sensi/SensitivityFactorReaderAdapter.java
```diff
@@ -46,50 +46,54 @@ public SensitivityFactorReaderAdapter(Network network, SensitivityFactorsProvide
 
     @Override
     public void read(Handler handler) {
+        ContingencyContext commonContingencyContext = new ContingencyContext(ContingencyContextType.ALL, null);
```
flo-dup, line 49: Just as small improvement, would be better to use 3 static methods to create ContingencyContexts (`ContingencyContext.createAllContingencyContext()`, `ContingencyContext.createNoneContingencyContext()`, `ContingencyContext.createSpecificContingencyContext(String id)`)

flo-dup (changes requested):

geofjamg (approved):

Issue #111, opened 2020-06-05
Reactive power control


annetill:
* **Do you want to request a *feature* or report a *bug*?**

A feature.

* **What is the current behavior?**

Reactive power control is not supported.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Some TSOs (Elia) use reactive power control for generators.

* **What is the expected behavior?**

Support of remote and local reactive power control of generators and static var compensators. 

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

IIDM has to evolve to support that feature.
Details have to be specified.

PR #266, opened 2021-04-08
Remote reactive power control support

obrix:

Issue: #111 

The goal of this PR is to introduce a new regulation mode based on remote reactive power. 
It also use the new regulation mode for generator to be introduced in powsybl-core (Seen [here](https://github.com/powsybl/powsybl-core/pull/1658) )

TBD


src/main/java/com/powsybl/openloadflow/network/ReactivePowerControl.java
```diff
@@ -0,0 +1,49 @@
...
+        this.controlled = controlled;
+        this.targetValue = targetValue;
+        this.controllers = new LinkedHashSet<>();
+    }
+
+    public double getTargetValue() {
+        return targetValue;
+    }
+
+    public LfBus getControlledBus() {
```
annetill, line 32: You should have a `getControlledBranch`.

obrix: Yes this is still basically a copy of VoltageControl class. I will update that with the new information you have given me regarding DiscretePhaseControl class.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -72,6 +74,40 @@ private static void createVoltageControlEquations(VoltageControl voltageControl,
...
+    }
+
+    private static void createReactivePowerControlBranchEquation(RemoteReactivePowerControl reactivePowerControl, LfBranch branch, VariableSet variableSet,
+                                                      EquationSystem equationSystem, AcEquationSystemCreationParameters creationParameters) {
+
+        //Reactive power control on a branch is always distant
+        LfBus controller = reactivePowerControl.getControllerBus();
+        if (controller != null) {
+
+            Equation eqQDist = equationSystem.createEquation(controller.getNum(), EquationType.BRANCH_Q);
```
annetill, line 88: Here I think that you have to follow the same design as:
```
equationSystem.createEquation(controller.getNum(), EquationType.BRANCH_ALPHA1)
                        .addTerm(EquationTerm.createVariableTerm(controller, VariableType.BRANCH_ALPHA1, variableSet))
                        .setActive(false);
```

src/main/java/com/powsybl/openloadflow/util/ParameterConstants.java
```diff
@@ -26,6 +26,9 @@
     public static final String VOLTAGE_REMOTE_CONTROL_PARAM_NAME = "voltageRemoteControl";
     public static final boolean VOLTAGE_REMOTE_CONTROL_DEFAULT_VALUE = true;
 
+    public static final String REACTIVE_POWER_REMOTE_CONTROL_PARAM_NAME = "reactivePowerRemoteControl";
+    public static final boolean REACTIVE_POWER_REMOTE_CONTROL_DEFAULT_VALUE = true;
```
annetill, line 30: This sould be `false` bu default.

obrix: This will disable reactive power remote control by default. This mean that setting a RegulationMode to ReactivePower on a generator will have no effect unless we also enable this parameter explictly.
Are we sure we want this behavior ? If yes I would also add a warning somewhere that warn the user its generator configuration is being overriden by a global parameter default value.

annetill: Yes, I am sure because sometimes it could lead to unfeasible equation system. But, now I wonder about the API on powsybl-core. If a generator has a remote reactive target and an active regulation, if we ignore it, it means that we have to found a `targetQ` that matches locally. @MioRtia this is our issue :-) We have to store from CGMES file both SSH local targetQ and the remote targetQ... what do you think?

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -45,6 +45,9 @@ private static void createBusEquations(LfNetwork network, VariableSet<AcVariable
                 createDiscreteVoltageControlEquation(bus, variableSet, equationSystem);
             }
             Equation<AcVariableType, AcEquationType> v = equationSystem.createEquation(bus.getNum(), AcEquationType.BUS_V);
+            if (creationParameters.isReactivePowerControl()) {
```
geofjamg, line 48: Same concern with this, it it kind of hack that is done after createVoltageControlEquations. We could have a single place to configure control coming from generators.

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
```diff
@@ -35,6 +35,14 @@
 
     protected String controlledBusId;
 
+    protected boolean hasReactivePowerControl = false;
```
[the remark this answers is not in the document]
flo-dup: Replaced with a GeneratorType enum `{NONE, REACTIVE_POWER_CONTROLLED, VOLTAGE_CONTROLLED}`

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
flo-dup: I'm not sure I understood properly what you asked @geofjamg: I put this reactive power control code in a `createGeneratorControlEquations` method, together with the content of `createVoltageControlEquations`, using the fact that it is either voltage control or reactive power control

flo-dup (approved):

geofjamg (approved):

PR #269, opened 2021-04-09
BusVoltagePerTargetV sensitivity factor support

Djazouli:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/SensitivityFactorReaderAdapter.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/SensitivityFactorReaderAdapter.java
@@ -6,16 +6,21 @@
  */
 package com.powsybl.openloadflow.sensi;
 
+import com.powsybl.commons.PowsyblException;
 import com.powsybl.contingency.Contingency;
+import com.powsybl.iidm.network.Bus;
 import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.network.TopologyLevel;
 import com.powsybl.sensitivity.SensitivityFactor;
 import com.powsybl.sensitivity.SensitivityFactorsProvider;
 import com.powsybl.sensitivity.factors.*;
 import com.powsybl.sensitivity.factors.functions.BranchFlow;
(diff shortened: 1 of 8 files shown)


geofjamg (approved):

src/main/java/com/powsybl/openloadflow/sensi/SensitivityFactorReaderAdapter.java
```diff
@@ -71,6 +76,12 @@ private void read(Handler handler, SensitivityFactor factor) {
             PhaseTapChangerAngle phaseTapChangerAngle = ((BranchIntensityPerPSTAngle) factor).getVariable();
             handler.onSimpleFactor(factor, SensitivityFunctionType.BRANCH_CURRENT, branchIntensity.getBranchId(),
                     SensitivityVariableType.TRANSFORMER_PHASE, phaseTapChangerAngle.getPhaseTapChangerHolderId());
+        } else if (factor instanceof BusVoltagePerTargetV) {
+            BusVoltage busVoltage = ((BusVoltagePerTargetV) factor).getFunction();
+            TargetVoltage targetVoltage = ((BusVoltagePerTargetV) factor).getVariable();
+            Bus bus = busVoltage.getBusRef().resolve(network, TopologyLevel.BUS_BRANCH).orElseThrow(() -> new PowsyblException("The bus ref for '" + busVoltage.getId() + "' cannot be resolved."));
```
annetill, line 82: For me, it should be the `BUS_BREAKER` topology level.

geofjamg:
For me, everything seems to be correct except generator remote voltage which is probably not taken into account (to test, or maybe just throws an exception in this version just to be clear it is not supported?).

Regarding topology view.
  - sensitivity calculation has to be done on bus view and without any non impedant branch (we could not compute sensitivity of the non impedant branch, the methodology does not work), that is why we have forced in the parameter breaker to false, and create min impendance. 
  - so for voltage sensitivity input, if we resolve to bus from bus view everything should be ok.

annetill (approved):

PR #272, opened 2021-04-12
DC sensitivity analysis: bug fix in case of contingency losing a participating element

Djazouli:

This is a bugfix

Currently, while running a DC sensitivity analysis, the following bug can occur: 
If there are several contingencies, and one of them changes the slack distribution (by losing generators, for example), then it will change the factorStates, and the basecase sensitivity, for all following contingencies. The one that does not change the participation will then be wrong

We do not overwrite the factorStates matrix, but create a new one.

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
@@ -730,8 +730,7 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis {
                         slackParticipationByBusForThisConnectivity = Collections.singletonMap(lfNetwork.getBusById(lfNetwork.getSlackBus().getId()), -1d);
                     }
 
-                    factorsStates.reset(); // avoid creating a new matrix to avoid buffer allocation time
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -730,8 +730,7 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
                         slackParticipationByBusForThisConnectivity = Collections.singletonMap(lfNetwork.getBusById(lfNetwork.getSlackBus().getId()), -1d);
                     }
 
-                    factorsStates.reset(); // avoid creating a new matrix to avoid buffer allocation time
-                    fillRhsSensitivityVariable(equationSystem, factorGroups, factorStateForThisConnectivity, slackParticipationByBusForThisConnectivity);
+                    factorStateForThisConnectivity = initFactorsRhs(equationSystem, factorGroups, slackParticipationByBusForThisConnectivity);
```
annetill, line 733: I understand you fix. I wonder if there is a way to class the contingencies treatment in order to have in one side contingencies that don't affect the slack participation and in the other side contingencies that affect it.

annetill (approved):

PR #279, opened 2021-04-21
Add slackBusPMaxMismatch parameter

obrix:

Add slackBusPMaxMismatch  parameter with a default value of 0.01

A new parameter is available in .yml files.

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
@@ -11,7 +11,6 @@ import com.powsybl.commons.config.ModuleConfig;
 import com.powsybl.commons.config.PlatformConfig;
 import com.powsybl.commons.extensions.AbstractExtension;
 import com.powsybl.loadflow.LoadFlowParameters;
-import com.powsybl.openloadflow.network.PerUnit;
 import com.powsybl.openloadflow.network.SlackBusSelector;
 import com.powsybl.openloadflow.network.SlackBusSelectorParametersReader;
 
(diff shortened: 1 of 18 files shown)


annetill (changes requested):
How does it work? With your PR, you are able to change a parameter, but where is this parameter used in the code? Then, you have to reach 90% of test coverage in OLF.

obrix:
@annetill This is a new attribute in OpenLoadFlowParameter class, it is used the same way as other parameters like throwsExceptionInCaseOfSlackDistributionFailure.
I just plugged it into DistributedSlackOuterLoop class constructor which is the reference you gave to me. If you think it should also be passed elsewhere, let me know. Coverage is fixed now that the parameter is used.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -45,6 +46,8 @@
 
     private boolean addRatioToLinesWithDifferentNominalVoltageAtBothEnds = ADD_RATIO_TO_LINES_WITH_DIFFERENT_NOMINAL_VOLTAGE_AT_BOTH_ENDS_DEFAULT_VALUE;
 
+    private double slackBusPMaxMismatch = SLACK_BUS_P_MAX_MISMATCH_DEFAULT_VALUE / PerUnit.SB;
```
geofjamg, line 49: You should keep it in MW in the parameters

src/main/java/com/powsybl/openloadflow/ac/DistributedSlackOuterLoop.java
```diff
@@ -49,7 +47,7 @@ public String getType() {
     @Override
     public OuterLoopStatus check(OuterLoopContext context, Reporter reporter) {
         double slackBusActivePowerMismatch = context.getLastNewtonRaphsonResult().getSlackBusActivePowerMismatch();
-        if (Math.abs(slackBusActivePowerMismatch) > SLACK_BUS_P_MAX_MISMATCH) {
+        if (Math.abs(slackBusActivePowerMismatch) > slackBusPMaxMismatch) {
```
geofjamg, line 50: This is here that you should convert to PU

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -116,6 +119,15 @@ public OpenLoadFlowParameters setPlausibleActivePowerLimit(double plausibleActiv
         return this;
     }
 
+    public double getSlackBusPMaxMismatch() {
+        return slackBusPMaxMismatch;
+    }
+
+    public OpenLoadFlowParameters setSlackBusPMaxMismatch(double pSlackBusPMaxMismatch) {
+        this.slackBusPMaxMismatch = pSlackBusPMaxMismatch / PerUnit.SB;
```
annetill, line 127: I get it, so I remove the per unit here.

geofjamg (approved):

PR #283, opened 2021-04-23
Implementation of PROPORTIONAL_TO_GENERATION_P balancetype

obrix:

Implementation of PROPORTIONAL_TO_GENERATION_P balancetype : 

It reuses GenerationActionPowerDistributionStep class as implementation is the same, only the participation factor is computed differently.

Unsupported balance type test has been replaced by a small test of proportional to generation P balance type.

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/util/GenerationActionPowerDistributionStep.java
+++ b/src/main/java/com/powsybl/openloadflow/network/util/GenerationActionPowerDistributionStep.java
@@ -24,6 +24,17 @@ public class GenerationActionPowerDistributionStep implements ActivePowerDistrib
 
     private static final Logger LOGGER = LoggerFactory.getLogger(GenerationActionPowerDistributionStep.class);
 
+    public enum ParticipationType {
+        MAX,
+        TARGET
+    }
+
+    private ParticipationType participationType;
+
+    public GenerationActionPowerDistributionStep(ParticipationType pParticipationType) {
+        this.participationType = pParticipationType;
+    }
+
     @Override
     public String getElementType() {
         return "generation";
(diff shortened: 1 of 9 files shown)


obrix:
@geofjamg Can you confirm GenerationActionPowerDistributionStep should be renamed GenerationActivePowerDistributionStep ?

src/main/java/com/powsybl/openloadflow/network/util/GenerationActionPowerDistributionStep.java
```diff
@@ -92,4 +105,19 @@ public double run(List<ParticipatingElement> participatingElements, int iteratio
...
     }
+
+    private double getParticipationFactor(LfGenerator generator) {
+        double factor;
+        switch (participationType) {
+            case MAX:
+                factor = generator.getMaxP() / generator.getDroop();
+                break;
+            case TARGET:
+                factor = generator.getTargetP();
```
annetill, line 116: Finally, I think that the absolute value is required here.

obrix:
I made the choice of removing getParticipationFactor from LfGenerator and to compute the participationFactor from GenerationActionPowerDistributionStep as it is dependant from the BalanceType configuration. 

This keeps the lfGenerator independant from GenerationActionPowerDistributionStep configuration but made mandatory the implementation of LfGenerator::getDroop.

@geofjamg Any opinion about that ? I think keeping GenerationActionPowerDistributionStep generic and independant is a good idea but I'm not sure about the relevance of exposing the droop parameter.

src/main/java/com/powsybl/openloadflow/network/util/GenerationActionPowerDistributionStep.java
obrix: Okay thanks I will modify the computation.

obrix: Can you confirm it's only for the factor in targetP mode ?

geofjamg:
> I made the choice of removing getParticipationFactor from LfGenerator and to compute the participationFactor from GenerationActionPowerDistributionStep as it is dependant from the BalanceType configuration.
> 
> This keeps the lfGenerator class independant from GenerationActionPowerDistributionStep configuration but made mandatory the implementation of LfGenerator::getDroop.
> 
> @geofjamg Any opinion about that ? I think keeping GenerationActionPowerDistributionStep generic and independant is a good idea but I'm not sure about the relevance of exposing the droop parameter.

Ok, good point.

geofjamg (approved):

PR #299, opened 2021-04-29
Support of BusVoltagePerTargetV relalted to a transformer

annetill:

We don't support it.

We support it. 

We through an exception.

We support it. We first compute the reference voltages like in a classical load flow with the transformer control outer loop active. Then we have to rebuild the equation system, as it was before the transformer steps were rounded and the PV buses deactivated. From that equation system, we compute the sensitivity values.

But:
- The equation system is not updated, but rebuild from scratch ;
- I don't use the equation system listener.
So it could be improved.

**Other information**:


src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -205,6 +213,24 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
...
+                        outerLoop.setActive(false);
+                    }
+                }
+                for (LfBus bus : lfNetwork.getBuses()) {
+                    if (bus.getDiscreteVoltageControl() != null && bus.getDiscreteVoltageControl().getMode().equals(DiscreteVoltageControl.Mode.OFF)) {
+                        // switch on regulating transformers
+                        bus.getDiscreteVoltageControl().setMode(DiscreteVoltageControl.Mode.VOLTAGE);
+                    }
+                }
+                engine.setEquationSystem(null); //FIXME: should be an update only.
```
annetill, line 230: @geofjamg it could be improved here: an update is enough using the `AcEquationSystemUpdater` but I don't know how to do.

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -903,4 +897,18 @@ public SensitivityFactorHolder readAndCheckFactors(Network network, Map<String,
         });
         return factorHolder;
     }
+
+    public boolean hasTransformerBusTargetVoltage(SensitivityFactorHolder factorHolder, Network network) {
+        List<LfSensitivityFactor> factors = factorHolder.getFactorsForBaseNetwork();
+        for (LfSensitivityFactor factor : factors) {
+            if (factor.getVariableType() == SensitivityVariableType.BUS_TARGET_VOLTAGE) {
+                TwoWindingsTransformer t2wt = network.getTwoWindingsTransformer(factor.getVariableId());
```
flo-dup, line 905: You could have used `getIdentifiable` and `instanceof` instead but ok.

flo-dup (approved):

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -210,6 +216,17 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
                 slackParticipationByBus = Collections.singletonMap(lfNetwork.getSlackBus(), -1d);
             }
 
+            // if we have at least one bus target voltage linked to a ratio tap changer, we have to rebuild the AC equation
+            // system obtained just before the transformer steps rounding.
+            if (hasTransformerBusTargetVoltage) {
+                for (LfBus bus : lfNetwork.getBuses()) {
```
sylvlecl, line 222: I am not sure how it will behave on contingencies : 
I guess the mode will be turned `OFF` again ? so maybe it does not work for the post contingency sensitivities ?

src/test/java/com/powsybl/openloadflow/sensi/ac/AcSensitivityAnalysisTest.java
```diff
@@ -347,6 +346,41 @@ void testBusVoltagePerTargetV() {
     @Test
     void testBusVoltagePerTargetVTwt() {
```
sylvlecl, line 347: Nice :)
I think we should add a test with a contingency

src/test/java/com/powsybl/openloadflow/sensi/ac/AcSensitivityAnalysisTest.java
```diff
@@ -362,13 +396,14 @@ void testBusVoltagePerTargetVTwt() {
...
-
-        assertTrue(e.getCause() instanceof NotImplementedException);
+        SensitivityAnalysisResult result = sensiProvider.run(network, VariantManagerConstants.INITIAL_VARIANT_ID, factorsProvider, Collections.emptyList(),
+                sensiParameters, LocalComputationManager.getDefault()).join();
 
-        assertEquals("[T2wT] Bus target voltage on two windings transformer is not managed yet", e.getCause().getMessage());
+        assertEquals(4, result.getSensitivityValues().size());
+        assertEquals(0d, getValue(result, "T2wT", "BUS_1"), LoadFlowAssert.DELTA_V);
+        assertEquals(0.035205d,  getValue(result, "T2wT", "BUS_2"), LoadFlowAssert.DELTA_V);
+        assertEquals(1d,  getValue(result, "T2wT", "BUS_3"), LoadFlowAssert.DELTA_V);
```
sylvlecl, line 405: Just a thought: 
I had a doubt (not obvious when reading the code), that sensitivities are correctly "scaled back" from per-unit.
Finally it seems correct, but it looks like the logic is handled directly in the equation we solve for computing the sensitivity : I think at some point we should make more clear where is the limit between per-united values and "nominal" values (probably having a kind of conversion after having computed the sensitivities in per-unit).

sylvlecl (commented):
Nice!
I think we need to add some test/handling for post contingencies sensitivities though.

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -180,6 +180,12 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
                 .collect(Collectors.toSet());
 
         // create AC engine
+        boolean hasTransformerBusTargetVoltage = hasTransformerBusTargetVoltage(factorHolder, network);
+        if (hasTransformerBusTargetVoltage) {
+            // if we have at least one bus target voltage linked to a ratio tap changer, we activate the transformer
```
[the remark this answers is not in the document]
annetill: So you think that the user should give a bus target voltage related to a transformer AND explicitly says that transformer control is enabled? 
What I don't get in your point is: what is for you the starting point to compute sensitivity values? In AC, it is quite important. In the code, the starting point is after transformer rounding. You prefer having both modes: after the rounding if we have the control enabled in the parameters or just from the initial network?

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
annetill: I will add a test but it will be similar to the generator behaviour.

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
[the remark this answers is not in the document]
sylvlecl: No no I think it's perfectly OK to compute sensitivities after the rounding.

What I mean is that: 
1. if you run a simple loadflow with voltage control disabled, you will get voltage values without tap changes, OK
2. if you run a sensitivity analysis with the same parameters (=voltage control disabled), now you will get different values for voltages (and sensitivities around those different values), because taps will have moved

I think it's weird, if you run the computation with the same parameters, I would expect to obtain the same voltage values.
If voltage control is disabled, I think I would expect to have 0 as sensitivity value (if we do this, maybe we could add a WARN log ?).

src/test/java/com/powsybl/openloadflow/sensi/ac/AcSensitivityAnalysisTest.java
annetill: It is done in `calculateSensitivityValues` but it is not well done for BUS_VOLTAGE and I have some doubt about `sensi += Math.toRadians(factor.getEquationTerm().der(phi1Var));` line but no link with that PR :-)

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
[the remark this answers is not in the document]
annetill: Yes I can do that.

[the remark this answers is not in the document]
annetill: Second thought: for sensitivity calculation linked to a phase shifter, we don’t use the loadflow parameters. We can compute the sensitivity even if the regulation is not active. So it is still coherent to rely on the loadflow parameter on our use case, I think yes but what is our opinion?

sylvlecl (approved):

PR #301, opened 2021-05-04
Sensitivity analysis: fix voltage reference values for BusVoltagePerTargetV factor

sylvlecl:

Bug

Reference values of voltages are not correctly transformed from per-unit back to kV.

We use nominal voltage to correctly transform back the voltage value.

**Other information**:

This is a quick fix, but I guess that those scaling logics could be shared more with other places in the code (loadflow, security analysis).
Also, it's not very safe to have unit tests with nominal voltage all equal to 1 kV, to identify that kind of issues.


annetill (approved):
Thanks a lot @sylvlecl and really sorry to this issue...

PR #302, opened 2021-05-04
Support of slack distribution on loads with negative P0

annetill:

No.

- Loads with negative P0 becomes participating elements ;
- For balance type equals to conform loads, we only works with loads with `LoadDetail` extension. Before, if the extension was no present, we consider `P0` as variable part. Now, we consider `0` ;
- For maintaining power factor constant on loads, our code was working only for buses with loads with positive `P0`. We cannot consider this formula with one negative load connected to the bus:
````
        double constantRatio = bus.getLoadTargetQ() / bus.getLoadTargetP();
        double newLoadTargetQ = newLoadTargetP * constantRatio;
        if (newLoadTargetQ != bus.getLoadTargetQ()) {
            LOGGER.trace("Rescale '{}' reactive power target on load: {} -> {}",
                    bus.getId(), bus.getLoadTargetQ() * PerUnit.SB, newLoadTargetQ * PerUnit.SB);
            bus.setLoadTargetQ(newLoadTargetQ);
        }
```` 

**Other information**:

### CODE REVIEW

File: src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java [hunk 1]
```diff
@@ -420,17 +424,30 @@ void updateGeneratorsState(double generationQ, boolean reactiveLimits) {
...
+            }
             load.getTerminal()
-                    .setP(load.getP0() >= 0 ? factorP * load.getP0() : load.getP0())
-                    .setQ(load.getQ0() >= 0 ? factorQ * load.getQ0() : load.getQ0());
+                    .setP(updatedP0)
+                    .setQ(updateQ0);
+            sumUpdatedQ += updateQ0;
+        }
+        if (Math.abs(sumUpdatedQ - loadTargetQ) > Q_DISPATCH_EPSILON) {
+            LOGGER.error("FIXME"); // it happens when we have a load with negative P0 in the bus.
```
geofjamg, line 450: We cannot let this as it is because when running on some cases, the output is cluttered with FIXME.
Do we know how to fix it?

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
annetill: No, I really don't know how to fix that. Keeping power factor constant for the whole bus works only if all loads have positive set points, otherwise I have to think. For sure, it cannot let a `FIXME` here, but I have no idea of how to handle that.

annetill: I have solve this issue by introducing `LfLoad` new elements. It works and I really don't know how to do this without this new object. Just take a look and tell me your opinion.

src/main/java/com/powsybl/openloadflow/network/LfBus.java
```diff
@@ -99,11 +106,13 @@
 
     List<LfShunt> getShunts();
 
+    List<Load> getLoads();
```
geofjamg, line 109: IIDM elements should not be exposed to OLF core components

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -182,7 +182,8 @@ public static AcLoadFlowParameters createAcParameters(Network network, MatrixFac
                                         parametersExt.isAddRatioToLinesWithDifferentNominalVoltageAtBothEnds(),
                                         branchesWithCurrent,
                                         parameters.getConnectedComponentMode() == LoadFlowParameters.ConnectedComponentMode.MAIN,
-                                        parameters.getCountriesToBalance());
+                                        parameters.getCountriesToBalance(),
+                                        parameters.getBalanceType() == LoadFlowParameters.BalanceType.PROPORTIONAL_TO_CONFORM_LOAD);
```
annetill, line 186: I forgot `parameters.isDistributedSlack() &&`

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -44,7 +45,8 @@ public DcLoadFlowEngine(LfNetwork network, MatrixFactory matrixFactory, boolean
 
     public DcLoadFlowEngine(Object network, DcLoadFlowParameters parameters, Reporter reporter) {
         LfNetworkParameters lfNetworkParameters = new LfNetworkParameters(parameters.getSlackBusSelector(), false, false, false, false,
-                parameters.getPlausibleActivePowerLimit(), false, parameters.isComputeMainConnectedComponentOnly(), parameters.getCountriesToBalance());
+                parameters.getPlausibleActivePowerLimit(), false, parameters.isComputeMainConnectedComponentOnly(), parameters.getCountriesToBalance(),
+                parameters.getBalanceType() == LoadFlowParameters.BalanceType.PROPORTIONAL_TO_CONFORM_LOAD);
```
annetill, line 49: Same remark.

src/main/java/com/powsybl/openloadflow/network/LfBus.java
```diff
@@ -53,15 +55,21 @@
 
     void setLoadTargetP(double loadTargetP);
 
+    double getInitialLoadTargetP();
```
annetill, line 58: all these methos can be hold by `LfLoads` now.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -367,6 +348,11 @@ public void setCalculatedQ(double calculatedQ) {
         return generators;
     }
 
+    @Override
+    public LfLoads getLfLoads() {
```
geofjamg, line 352: I am not sure this LfLoads is more performant than List<LfLoad> but why not. Maybe we should create it lazily to only impact LF with constant power factor load distribution.

src/main/java/com/powsybl/openloadflow/network/impl/LfLoads.java
```diff
@@ -14,69 +14,91 @@
...
+ * @author Anne Tilloy <anne.tilloy at rte-france.com>
+ */
 public class LfLoads extends AbstractElement {
 
     private final List<Load> loads = new ArrayList<>();
 
-    private final List<Double> participationFactors = new ArrayList<>();
+    private double[] participationFactors;
 
-    private final List<Double> powerFactors = new ArrayList<>();
```
annetill: Why have you removed that 2 tables `powerFactors` and `p0s`? As you have done, it is similar as what I have done before and with the comment that we don't have to expose IIDM network in the LfNetwork. The method `getLoadTargetQ` is used in a outerloop.

flo-dup: Because there's no need to store them, they're already stored in the loads, and remain constant. What you did before was exposing IIDM network by adding `getLoads` to `Lfbus` interface, which was returning `Load` IIDM objects, and you were handling these loads in the distribution step itself. Here IIDM is not exposed, it stays within `LfLoads`.

geofjamg (approved):

Issue #289, opened 2021-04-26
Contingency propagation in sensitivity analysis


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Bug.

* **What is the current behavior?**

Sensitivity analysis is performed without keeping switches which are necessary for proper propagation of some contingencies.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

TODO

* **What is the expected behavior?**

Contingencies should correctly propagate in sensitivity analysis, opening all relevant switches. 
An issue seems to be that fast DC sensitivity analysis cannot handle, for now, that kind of contingencies.
Therefore, for DC analysis, we could either:
* throw if such a contingency is defined in inputs
* not propagate that kind of contingencies, but we should have a warning / doc about this

* **Please tell us about your environment:**
  - PowSyBl Version: powsybl-open-loadflow 0.8.0-SNAPSHOT

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

In version 0.7.0, propagated contingencies ended with an exception:
```
The contingency on the branch <SWITCH ID> not found in the network
```
Since commit https://github.com/powsybl/powsybl-open-loadflow/commit/32b3b5fc15589af485d9b970e95dda5b1d2b8a82, there is not more exception, but the behaviour is not right: 
switches are just skipped, they stay closed whereas they should open.

It would make sense, at least for AC sensitivity analysis, to keep breakers in order to be able to properly propagate contingencies, like in AC security analysis.

annetill:
Thanks a lot @sylvlecl for this well described issue. The code to solve this issue could be quite consequent as all the sensitivity analysis are performed on the bus view. We made this choice for performance issue as a first approach. But, you are right this kind of propagation is not well handled. I propose to solve this issue in two steps:
- A warning for these kind of contingency, that will not be propagated at all and a documentation ;
[...]

PR #303, opened 2021-05-05
Detects coupler when propagating contingency in sensitivity analysis

flo-dup:

Yes, first step to solve issue #289. This doesn't fix that issue, but detects the problematic contingency, gives a warning and does not add it to the propagated contingencies to avoid giving wrong results.

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
+++ b/src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
@@ -10,6 +10,8 @@ import com.powsybl.commons.PowsyblException;
 import com.powsybl.contingency.Contingency;
 import com.powsybl.contingency.ContingencyElement;
 import com.powsybl.iidm.network.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -21,6 +23,8 @@ import java.util.Set;
  * @author Gaël Macherel <gael.macherel@artelys.com>
  */
 public class PropagatedContingency {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(PropagatedContingency.class);
 
     private final Contingency contingency;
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
```diff
@@ -46,55 +49,88 @@ public int getIndex() {
...
+                    throw new UnsupportedOperationException("TODO");
             }
-
         }
-        return propagatedContingencies;
+
+        return new PropagatedContingency(contingency, index, branchIdsToOpen, hvdcIdsToOpen, switchesToOpen, terminalsToDisconnect);
+    }
+
+    private static boolean isCoupler(Switch s) {
```
annetill, line 130: `hasBusBarSectionCoulper` could be better here, what do you think?

src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
```diff
@@ -46,55 +49,88 @@ public int getIndex() {
...
-                        //TODO: support all kinds of contingencies
-                        throw new UnsupportedOperationException("TODO");
-                }
+            PropagatedContingency propagatedContingency = PropagatedContingency.create(network, contingency, index);
+            Optional<Switch> coupler = propagatedContingency.switchesToOpen.stream().filter(PropagatedContingency::isCoupler).findFirst();
+            if (coupler.isEmpty()) {
+                propagatedContingencies.add(propagatedContingency);
+            } else {
+                // Sensitivity analysis works in bus view, it cannot deal (yet)  with contingencies whose propagation encounters a coupler
+                LOGGER.warn("Propagated contingency '{}' not processed: coupler '{}' has been encountered while propagating the contingency",
```
annetill, line 81: I don't know if coupler has a meaning in English indeed. But for sure you have to precise that it is between two bus bar sections.

src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
geofjamg: for me coupler already means busbar section coupler

src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
geofjamg: Coupler is a well known term in power system

annetill: Here you skip the contingency, so it means that the result will not contain this contingency. The other solution is to set a predefined result. What is preferable?

Issue #97, opened 2020-05-20
Static var compensator voltage control increasing


annetill:
* **Do you want to request a *feature* or report a *bug*?**

A feature. 

* **What is the current behavior?**

We model the voltage remote control of static var compensator as we do with generators with PQV nodes and equations that link the reactive power of controllers (2 by 2). 

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

It is not a bug but sometimes we prefer to have an simplified approach. 

* **What is the expected behavior?**

When the cim:StaticVarCompensator.slope is positive (and present in an IGM CIM-CGMES file), the reactive output of the SVC is defined by:
Q=(V_bus-V_ref)/slope   
Where:
cim:StaticVarCompensator.slope is positive based on the load sign convention.
Vbus is the voltage at the controlled point determined by cim:RegulatingControl.Terminal which may be local or remote.
Vref is the target value given by cim:RegulatingControl.targetValue.

Extract from the ENTSO-E implementation guide for CGM data modeling 1.0

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

geofjamg:
Such equipment may also have reactive power limits?

annetill:
![image](https://user-images.githubusercontent.com/48123713/98100985-e8d07980-1e91-11eb-9e09-92083c7905ca.png)

This model could be useful. 

The generator model represents the slope by connecting ths SVC to an auxiliary bus separated from the high voltage bus by a reactance equal to the per unit slope. When the SVC is operating the limits, its behaves like a fixed susceptance and its injected reactive power is proportional to the square of the regulated bus voltage.

geofjamg:
Done in #391

PR #304, opened 2021-05-06
Support of static var compensator with VoltagePerReactivePowerControl extension

annetill:

Yes, issue #97

If a unique static var compensator with non-zero slope is connected to a bus, regulating voltage locally, the slope attribute is taken into account. I have to increase the unit testing because the more complexe thing in that PR is to deal with all the configurations of generators connected to that bus: with or without slope, regulating voltage or reactive power. 

The slope is ignored.

If a unique static var compensator with non-zero slope is connected to a bus, regulating voltage locally, the slope attribute is taken into account. 

**Other information**:


annetill:
@annetill Note for that PR: the bus to which the static var compensator is connected is initialized with the nominal voltage. It could be better to start from the targetV value.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -364,6 +366,15 @@ private static void createImpedantBranch(LfBranch branch, LfBus bus1, LfBus bus2
             }
             sq1.addTerm(q1);
             branch.setQ1(q1);
+            // Add term slope * Q for svc
```
annetill, line 369: Can you introduce an utility method that will be called here and just below? It will make the code more comprehensive.

annetill (commented):
Changes requested. In the target vector, you should have only the ideal targetV.

annetill:
@Hadrien-Godard I am going to rebase the branch on master. The work is not finished because our controller bus with slope can switch PV to PQ, but I think we have to avoid switching them from PV to PQ in a first approach.

Hadrien-Godard:
> @Hadrien-Godard I am going to rebase the branch on master. The work is not finished because our controller bus with slope can switch PV to PQ, but I think we have to avoid switching them from PV to PQ in a first approach.

I think we can let them switch PV to PQ in the same manner as classic PV buses.

src/main/java/com/powsybl/openloadflow/ac/ReactiveLimitsOuterLoop.java
```diff
@@ -196,10 +194,12 @@ private void checkPqBus(LfBus controllerCapableBus, List<PqToPvBus> pqToPvBuses)
         double minQ = controllerCapableBus.getMinQ();
         double maxQ = controllerCapableBus.getMaxQ();
         double q = controllerCapableBus.getGenerationTargetQ();
-        if (Math.abs(q - maxQ) < Q_EPS && controllerCapableBus.getV().eval() > getBusTargetV(controllerCapableBus)) { // bus produce too much reactive power
+        double distanceToMaxQ = Math.abs(q - maxQ);
+        double distanceToMinQ = Math.abs(q - minQ);
+        if (distanceToMaxQ < distanceToMinQ  && controllerCapableBus.getV().eval() > getBusTargetV(controllerCapableBus)) { // bus produce too much reactive power
```
annetill, line 199: @geofjamg can you please validate these new lines. For me, it works because we are sure that the bus checked is a PQ bus that where previously PV. Do you agree?

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -61,6 +61,8 @@ private static void createVoltageControls(LfNetwork lfNetwork, List<LfBus> lfBus
             List<LfGenerator> voltageControlGenerators = controllerBus.getGenerators().stream().filter(LfGenerator::hasVoltageControl).collect(Collectors.toList());
```
annetill, line 61: Here we just have the generators linked to a controller bus. But, what happens in the case where we have two controller buses with each one SVC with slope ? We don't support that configuration too.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -92,6 +94,22 @@ private static void createVoltageControls(LfNetwork lfNetwork, List<LfBus> lfBus
...
+            long nbvoltageControlGeneratorsWithSlope = voltageControlGenerators.stream().filter(generator -> generator.getSlope() != 0).count();
+            if (nbvoltageControlGeneratorsWithSlope > 0) {
+                if (nbvoltageControlGeneratorsWithSlope == 1) {
+                    // we don't support a generator controlling voltage with slope and other generators controlling voltage
+                    LOGGER.warn("Non supported: several generators of bus {} control voltage and one controls voltage with slope. Slope is therefore deactivated on those generators.", controllerBus);
+                } else {
+                    // we don't support several generators controlling voltage with slope
+                    LOGGER.warn("Non supported: several generators of bus {} control voltage with slope. Slope is therefore deactivated on those generators.", controllerBus);
+                }
+                voltageControlGenerators.forEach(g -> g.setSlope(0));
```
annetill, line 108: Why we don't keep one finally ?

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -65,30 +67,63 @@ private static void createVoltageControls(LfNetwork lfNetwork, List<LfBus> lfBus
...
                         controllerBus.getId(), controllerTargetV, controllerTargetV * controllerBus.getNominalV() / controlledBus.getNominalV());
-                    controlledBus = controllerBus;
+                    createVoltageControl(controllerBus, controllerBus, controllerTargetV, voltageControls);
                 }
+            }
+        }
 
-                VoltageControl voltageControl = controlledBus.getVoltageControl().orElse(new VoltageControl(controlledBus, controllerTargetV));
-                voltageControl.addControllerBus(controllerBus);
+        voltageControls.forEach(LfNetworkLoaderImpl::checkGeneratorsWithSlope);
```
annetill, line 93: You can not do that if the parameter is not active, it will be a loss in performances.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -65,30 +67,63 @@ private static void createVoltageControls(LfNetwork lfNetwork, List<LfBus> lfBus
...
+    }
+
+    private static void updateVoltageControl(VoltageControl voltageControl, LfBus controllerBus, double controllerTargetV) {
+        voltageControl.addControllerBus(controllerBus);
+        checkUniqueTargetVControlledBus(controllerTargetV, controllerBus, voltageControl);
+    }
+
+    private static void checkGeneratorsWithSlope(VoltageControl voltageControl) {
+
+        List<LfGenerator> generatorsWithSlope = voltageControl.getControllerBuses().stream()
```
annetill, line 110: Ti be checked with Geo for performance.

src/main/java/com/powsybl/openloadflow/network/impl/LfStaticVarCompensatorImpl.java
```diff
@@ -60,6 +63,9 @@ public double getMaxQ(double p) {
         if (svc.getRegulationMode() == StaticVarCompensator.RegulationMode.VOLTAGE) {
             setVoltageControl(svc.getVoltageSetpoint(), svc.getRegulatingTerminal(), breakers, report);
         }
+        if (svc.getExtension(VoltagePerReactivePowerControl.class) != null) {
```
annetill, line 66: Indeed, as we have the parameter in `LfNetworkParameters`, why we don't use it here? If it is false, can we set the slope to zero?

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemUpdater.java
```diff
@@ -47,7 +47,18 @@ private void updateControlledBus(VoltageControl voltageControl, EquationSystem e
...
             .filter(LfBus::isVoltageControllerEnabled)
             .collect(Collectors.toList());
-        equationSystem.createEquation(controlledBus.getNum(), EquationType.BUS_V).setActive(!controllerBusesWithVoltageControlOn.isEmpty());
+
+        if (creationParameters.isVoltagePerReactivePowerControl() && voltageControl.isVoltageControlLocal()
+                && !controllerBuses.iterator().next().getGeneratorsControllingVoltageWithSlope().isEmpty()) {
+            LfBus controllerBus = controllerBuses.iterator().next();
+            if (!controllerBus.isVoltageControllerEnabled()) {
+                // we only support one controlling static var compensator without any other controlling generators
+                equationSystem.createEquation(controlledBus.getNum(), EquationType.BUS_V_SLOPE).setActive(false);
```
geofjamg, line 56: There is no criteria to switch it on?

geofjamg (approved):

flo-dup (approved):

PR #295, opened 2021-04-29
Handle all connected components of a network

obrix:
Try handling all the connected components in a network when building a LfNetwork and executing a loadflow.

**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [X] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*

Introduce a boolean value in the network loader to chose to load only the main connected component of a network, or all connected components.

**What is the current behavior?** *(You can also link to an open issue here)*

Only the main connected component was loaded.

**What is the new behavior (if this is a feature change)?**

All connected components can now be loaded.

**Does this PR introduce a breaking change or deprecate an API?** *If yes, check the following:*
- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

PR #308, opened 2021-05-20
Use new parameters isDcUseTransformerRatio and countriesToBalance

obrix:

Use the two new parameters available in powsybl-core isDcUseTransformerRatio and countriesToBalance. The branch for this PR is based on the branch from the PR https://github.com/powsybl/powsybl-open-loadflow/pull/295 to use the compatibility work already done for 4.2.0 release candidate.
Thus it is to be rebased and merged after this PR.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
@@ -265,9 +265,10 @@ public class OpenLoadFlowProvider implements LoadFlowProvider {
                                                                      parametersExt.getPlausibleActivePowerLimit(),
                                                                      parametersExt.isAddRatioToLinesWithDifferentNominalVoltageAtBothEnds(),
                                                                      true,
-                                                                     parameters.getConnectedComponentMode() == LoadFlowParameters.ConnectedComponentMode.MAIN);
+                                                                     parameters.getConnectedComponentMode() == LoadFlowParameters.ConnectedComponentMode.MAIN,
+                                                                      parameters.getCountriesToBalance());
 
-        List<DcLoadFlowResult> results = new DcLoadFlowEngine(network, dcParameters, reporter, parameters.getCountriesToBalance())
+        List<DcLoadFlowResult> results = new DcLoadFlowEngine(network, dcParameters, reporter)
                 .run(reporter);
 
         Networks.resetState(network);
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
@@ -181,7 +181,8 @@ public class OpenLoadFlowProvider implements LoadFlowProvider {
                                         forceA1Var,
                                         parametersExt.isAddRatioToLinesWithDifferentNominalVoltageAtBothEnds(),
                                         branchesWithCurrent,
-                                        parameters.getConnectedComponentMode() == LoadFlowParameters.ConnectedComponentMode.MAIN);
+                                        parameters.getConnectedComponentMode() == LoadFlowParameters.ConnectedComponentMode.MAIN,
+                                        parameters.getCountriesToBalance());
     }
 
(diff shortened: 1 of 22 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -566,4 +566,11 @@ private static LfNetwork create(int numCC, int numSC, List<Bus> buses, List<Swit
 
         return Optional.empty();
     }
+
+    static boolean participateToBalance(LfNetworkParameters parameters, Bus b) {
```
annetill, line 570: Can you change the method name to `participateToSlackDistribution`? It will be more coherent with the rest of the code.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcLoadFlowParameters.java
```diff
@@ -145,4 +150,8 @@ public boolean isComputeMainConnectedComponentOnly() {
         return computeMainConnectedComponentOnly;
     }
 
+    public Set<Country> getCountriesToBalance() {
+        return countriesToBalance;
```
flo-dup, line 154: Maybe add `Collections.unmodifiableSet()` to avoid unwanted changes to parameters.

src/main/java/com/powsybl/openloadflow/network/LfNetworkParameters.java
```diff
@@ -80,4 +87,8 @@ public boolean isAddRatioToLinesWithDifferentNominalVoltageAtBothEnds() {
     public boolean isComputeMainConnectedComponentOnly() {
         return computeMainConnectedComponentOnly;
     }
+
+    public Set<Country> getCountriesToBalance() {
+        return countriesToBalance;
```
flo-dup, line 92: same remark

flo-dup (commented):
In the end, that's not at all because of this PR, but these dc/ac/olf parameters classes are a bit messy. Maybe we should try to clean that up in another PR as you already suggested. By putting apart the load part and using the OpenLoadFlowParameters for it?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -266,7 +267,7 @@ private LoadFlowResult runDc(Network network, LoadFlowParameters parameters, Ope
                                                                      true,
                                                                      parameters.getConnectedComponentMode() == LoadFlowParameters.ConnectedComponentMode.MAIN);
 
-        List<DcLoadFlowResult> results = new DcLoadFlowEngine(network, dcParameters, reporter)
+        List<DcLoadFlowResult> results = new DcLoadFlowEngine(network, dcParameters, reporter, parameters.getCountriesToBalance())
```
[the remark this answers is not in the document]
obrix: Because I think it should neither be in AcParameters nor in DcParameters. I modified it to have the symmetry but I think the parameters class need to be refactorized to avoid duplication.

obrix:
@flo-dup  I aggree, there is a a lot of unnecessary duplications in the parameters class, we should discuss it soon.

src/main/java/com/powsybl/openloadflow/network/util/GenerationActivePowerDistributionStep.java
```diff
@@ -43,7 +43,7 @@ public String getElementType() {
     @Override
     public List<ParticipatingElement> getParticipatingElements(Collection<LfBus> buses) {
         return buses.stream()
-                .filter(bus -> !(bus.isDisabled() || bus.isFictitious()))
+                .filter(bus -> !(bus.isDisabled() || bus.isFictitious() || !bus.isParticipating()))
```
annetill, line 46: The double negative is not well understandable...

obrix: The alternative is bus.isParticipating() && !bus.isDisabled() && !bus.isFictitious(), I think its better, do you agree ?

annetill: Yes :-)

annetill (approved):

PR #313, opened 2021-05-27
DC security analysis

obrix:

WIP implementation of dc security analysis.

**Other information**:

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -11,6 +11,12 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
+      - name: Checkout powsybl-core sources
+            uses: actions/checkout@v1
+            with:
+              repository: powsybl/powsybl-core
+              ref: refs/heads/new_limit_violation_types
+
       - name: Checkout sources
         uses: actions/checkout@v1
 
@@ -18,6 +24,9 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 11
+
+      - name: Build and install powsybl-core with Maven
+        run: mvn --batch-mode -DskipTests=true --file ../powsybl-core/pom.xml install
 
       - name: Build with Maven
         run: mvn --batch-mode -Pjacoco install
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,7 +15,7 @@ jobs:
             uses: actions/checkout@v1
             with:
               repository: powsybl/powsybl-core
-              ref: refs/heads/new_limit_violation_types
+              ref: refs/heads/master
 
       - name: Checkout sources
         uses: actions/checkout@v1
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -12,10 +12,10 @@ jobs:
 
     steps:
       - name: Checkout powsybl-core sources
-            uses: actions/checkout@v1
-            with:
-              repository: powsybl/powsybl-core
-              ref: refs/heads/master
+        uses: actions/checkout@v1
+        with:
+          repository: powsybl/powsybl-core
+          ref: refs/heads/master
 
       - name: Checkout sources
         uses: actions/checkout@v1
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -11,12 +11,6 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
-      - name: Checkout powsybl-core sources
-        uses: actions/checkout@v1
-        with:
-          repository: powsybl/powsybl-core
-          ref: refs/heads/master
-
       - name: Checkout sources
         uses: actions/checkout@v1
 
@@ -24,9 +18,6 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 11
-
-      - name: Build and install powsybl-core with Maven
-        run: mvn --batch-mode -DskipTests=true --file ../powsybl-core/pom.xml install
 
       - name: Build with Maven
         run: mvn --batch-mode -Pjacoco install
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,6 @@
         <jafama.version>2.3.1</jafama.version>
         <junit-jupiter.version>5.5.2</junit-jupiter.version>
         <slf4jtoys.version>1.6.2</slf4jtoys.version>
-
         <powsybl-core.version>4.3.0-alpha-1</powsybl-core.version>
     </properties>
 
--- a/pom.xml
+++ b/pom.xml
@@ -47,6 +47,7 @@
         <jafama.version>2.3.1</jafama.version>
         <junit-jupiter.version>5.5.2</junit-jupiter.version>
         <slf4jtoys.version>1.6.2</slf4jtoys.version>
+
         <powsybl-core.version>4.3.0-alpha-1</powsybl-core.version>
     </properties>
 
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
@@ -112,7 +112,7 @@ public class OpenLoadFlowProvider implements LoadFlowProvider {
         return parametersExt;
     }
 
-    private static SlackBusSelector getSlackBusSelector(Network network, LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
+    public static SlackBusSelector getSlackBusSelector(Network network, LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
         SlackBusSelector slackBusSelector = SlackBusSelector.fromMode(parametersExt.getSlackBusSelectionMode(), parametersExt.getSlackBusId());
         return parameters.isReadSlackBus() ? new NetworkSlackBusSelector(network, slackBusSelector)
                                            : slackBusSelector;
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
@@ -112,7 +112,7 @@ public class OpenLoadFlowProvider implements LoadFlowProvider {
         return parametersExt;
     }
 
-    public static SlackBusSelector getSlackBusSelector(Network network, LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
+    private static SlackBusSelector getSlackBusSelector(Network network, LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
         SlackBusSelector slackBusSelector = SlackBusSelector.fromMode(parametersExt.getSlackBusSelectionMode(), parametersExt.getSlackBusId());
         return parameters.isReadSlackBus() ? new NetworkSlackBusSelector(network, slackBusSelector)
                                            : slackBusSelector;
--- a/src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
+++ b/src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
@@ -55,9 +55,9 @@ public abstract class AbstractLfBranch extends AbstractElement implements LfBran
 
     private final LfBus bus2;
 
-    private HashMap<LimitType, List<LfLimit>> limits1 = new HashMap<>();
+    private Map<LimitType, List<LfLimit>> limits1 = new HashMap<>();
 
-    private HashMap<LimitType, List<LfLimit>> limits2 = new HashMap<>();
+    private Map<LimitType, List<LfLimit>> limits2 = new HashMap<>();
 
     private final PiModel piModel;
 
--- a/src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
+++ b/src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
@@ -6,10 +6,7 @@
  */
 package com.powsybl.openloadflow.network;
 
-import com.powsybl.iidm.network.CurrentLimits;
-import com.powsybl.iidm.network.LoadingLimits;
-import com.powsybl.iidm.network.PhaseTapChanger;
-import com.powsybl.iidm.network.RatioTapChanger;
+import com.powsybl.iidm.network.*;
 import com.powsybl.openloadflow.network.impl.Transformers;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -58,9 +55,9 @@ public abstract class AbstractLfBranch extends AbstractElement implements LfBran
 
     private final LfBus bus2;
 
-    private List<LfLimit> limits1;
+    private HashMap<LimitType, List<LfLimit>> limits1 = new HashMap<>();
 
-    private List<LfLimit> limits2;
+    private HashMap<LimitType, List<LfLimit>> limits2 = new HashMap<>();
 
     private final PiModel piModel;
 
@@ -77,11 +74,11 @@ public abstract class AbstractLfBranch extends AbstractElement implements LfBran
         this.piModel = Objects.requireNonNull(piModel);
     }
 
-    protected static List<LfLimit> createSortedLimitsList(CurrentLimits currentLimits, LfBus bus) {
+    protected static List<LfLimit> createSortedLimitsList(LoadingLimits loadingLimits, LfBus bus) {
         LinkedList<LfLimit> sortedLimits = new LinkedList<>();
-        if (currentLimits != null) {
+        if (loadingLimits != null) {
             double toPerUnit = bus.getNominalV() / PerUnit.SB;
-            for (LoadingLimits.TemporaryLimit temporaryLimit : currentLimits.getTemporaryLimits()) {
+            for (LoadingLimits.TemporaryLimit temporaryLimit : loadingLimits.getTemporaryLimits()) {
                 if (temporaryLimit.getAcceptableDuration() != 0) {
                     // it is not useful to add a limit with acceptable duration equal to zero as the only value plausible
                     // for this limit is infinity.
@@ -90,7 +87,7 @@ public abstract class AbstractLfBranch extends AbstractElement implements LfBran
                     sortedLimits.addFirst(LfLimit.createTemporaryLimit(temporaryLimit.getAcceptableDuration(), valuePerUnit));
                 }
             }
-            sortedLimits.addLast(LfLimit.createPermanentLimit(currentLimits.getPermanentLimit() * toPerUnit));
+            sortedLimits.addLast(LfLimit.createPermanentLimit(loadingLimits.getPermanentLimit() * toPerUnit));
         }
         if (sortedLimits.size() > 1) {
             // we only make that fix if there is more than a permanent limit attached to the branch.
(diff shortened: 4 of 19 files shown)


src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
```diff
@@ -117,17 +114,19 @@ public LfBus getBus2() {
     }
 
     protected List<LfLimit> getLimits1(CurrentLimits currentLimits) {
```
flo-dup, line 116: Shouldn't we deprecate those 2 getLimits?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+        return false;
+    }
+
+    List<LfContingency> createContingencies(List<PropagatedContingency> propagatedContingencies, LfNetwork network) {
+        return LfContingency.createContingencies(propagatedContingencies, network, network.createDecrementalConnectivity(connectivityProvider), true);
+    }
+
+    double computeApparentPower(double p, double q) {
+        return Math.sqrt(p * p + q * q);
```
flo-dup, line 221: FastMath is used instead of Math in open-loadflow

src/main/java/com/powsybl/openloadflow/network/LfBranch.java
```diff
@@ -49,6 +50,10 @@
 
     List<AbstractLfBranch.LfLimit> getLimits2();
```
flo-dup, line 51: Same here?

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
```diff
@@ -88,7 +85,7 @@ protected AbstractLfBranch(LfNetwork network, LfBus bus1, LfBus bus2, PiModel pi
                     sortedLimits.addFirst(LfLimit.createTemporaryLimit(temporaryLimit.getAcceptableDuration(), valuePerUnit));
                 }
             }
-            sortedLimits.addLast(LfLimit.createPermanentLimit(currentLimits.getPermanentLimit() * toPerUnit));
+            sortedLimits.addLast(LfLimit.createPermanentLimit(loadingLimits.getPermanentLimit() * toPerUnit));
```
annetill, line 88: This is not going to work: current limits have a per-unit factor equal to `bus.getNominalV() / PerUnit.SB`, but active power limits and I guess apparent power limits have a per-unit factor only equal to `1 / PerUnit.SB`. Do you have a unit test directly on limits to validate all of that?

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
annetill: What do you mean by deprecate? These methods have to be removed.

src/main/java/com/powsybl/openloadflow/network/LfBranch.java
annetill: You mean remove?

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBranch.java
```diff
@@ -61,6 +62,19 @@ public boolean hasPhaseControlCapability() {
         return Collections.emptyList();
     }
 
+    @Override
+    public List<LfLimit> getLimits1(final LimitType type) {
+        if (type != LimitType.CURRENT) {
+            throw new UnsupportedOperationException(String.format("Getting %s limits is not supported on dangling line branch.", type.name()));
+        }
+        return getLimits1(type, danglingLine.getCurrentLimits());
```
annetill, line 70: You are not able to get active power and apparent power limits on the side 1 of the dangling line?

src/main/java/com/powsybl/openloadflow/network/impl/LfLegBranch.java
```diff
@@ -112,6 +113,16 @@ public boolean hasPhaseControlCapability() {
         return Collections.emptyList();
     }
 
+    @Override
+    public List<LfLimit> getLimits1(final LimitType type) {
```
annetill, line 117: You are not able to get active power and apparent power limits on network side of a three windings transformer?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -0,0 +1,223 @@
...
+                .map(temporaryLimit1 -> createLimitViolation1(branch, temporaryLimit1, LimitViolationType.CURRENT, PerUnit.SB / branch.getBus1().getNominalV(), branch.getI1().eval()))
+                .ifPresent(limitViolation -> violations.put(getSubjectSideId(limitViolation), limitViolation));
+
+            branch.getLimits1(LimitType.ACTIVE_POWER).stream()
+                  .filter(temporaryLimit1 -> branch.getP1().eval() > temporaryLimit1.getValue())
+                  .findFirst()
+                  .map(temporaryLimit1 -> createLimitViolation1(branch, temporaryLimit1, LimitViolationType.ACTIVE_POWER, PerUnit.SB, branch.getP1().eval()))
+                  .ifPresent(limitViolation -> violations.put(getSubjectSideId(limitViolation), limitViolation));
+
+            double apparentPower1 = computeApparentPower(branch.getP1().eval(), branch.getQ1().eval());
```
annetill, line 120: This could not be a method inside the branch?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -0,0 +1,223 @@
...
+                .map(temporaryLimit2 -> createLimitViolation2(branch, temporaryLimit2, LimitViolationType.CURRENT, PerUnit.SB / branch.getBus2().getNominalV(), branch.getI2().eval()))
+                .ifPresent(limitViolation -> violations.put(getSubjectSideId(limitViolation), limitViolation));
+
+            branch.getLimits2().stream()
+                  .filter(temporaryLimit2 -> branch.getP2().eval() > temporaryLimit2.getValue())
+                  .findFirst()
+                  .map(temporaryLimit2 -> createLimitViolation2(branch, temporaryLimit2, LimitViolationType.ACTIVE_POWER, PerUnit.SB, branch.getP2().eval()))
+                  .ifPresent(limitViolation -> violations.put(getSubjectSideId(limitViolation), limitViolation));
+
+            double apparentPower2 = computeApparentPower(branch.getP2().eval(), branch.getQ2().eval());
```
annetill, line 141: Same remark.

annetill (changes requested):

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
obrix: If we want this change to not be a breaking change, I think we must keep them as deprecated for a while.

src/main/java/com/powsybl/openloadflow/network/LfBranch.java
obrix: Same remark.

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
annetill: A breaking change for who? This is the internal network of the open loadlfow engine, nobody use that `LfNetwork` except the engine.

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
obrix: Yes, in an another place I adapted the scale value used in LimitViolation object building and used getNominalV only in the case of a CURRENT related limit violation. Did not catch this one.
Regarding this function, what you say is simply to adapt toPerUnit variable computation to the real LimitType that's it ? Keep bus.getNominalV() / PerUnit.SB for CURRENT and use 1 / PerUnit.SB for the rest ?

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBranch.java
obrix: Actually I think I can, must have been confused by DanglingLine interface that only has a getCurrentLimits but inherits the rest from FlowsLimitsHolder.

annetill: You think that we miss something in `powsybl-core`? User-friendy methods are missing?

src/main/java/com/powsybl/openloadflow/network/impl/LfLegBranch.java
obrix: Same as DanglingLine, I will match the implementation to the one of LfBranchImpl.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
obrix: You see this method as a part of LfBranch interface ?

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
obrix: I you say it is safe to break, I'll remove them.

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBranch.java
obrix: No, DanglineLineImpl gives the requested implementation for current, activepower and apparentpower. Dangling line interface just has a specific re declaration of getCurrentLimits which is curious. But all is there to support the three on openloadflow side.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
annetill: In general, I put some methods in the `LfBranch` interface to deal with special branches such as switch, dangling lines and legs. If you leave this method here, because it is more efficient, don't forget to think about how the special branches behave.

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBranch.java
annetill: @MioRtia Do you have an opinion on that?

MioRtia: Yes, sorry, it is a mistake because these methods existed before the introduction of the `FlowsLimitHolder` class... I will make a PR to delete them.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -0,0 +1,225 @@
...
+        buses.forEach(bus -> detectBusViolations(bus, violations));
+    }
+
+    /**
+     * Detect violation limits on one branch and add them to the given list
+     * @param branch branch of interest
+     * @param violations list on which the violation limits encountered are added
+     */
+    protected void detectBranchViolations(LfBranch branch, Map<Pair<String, Branch.Side>, LimitViolation> violations) {
+        // detect violation limits on a branch
```
flo-dup, line 106: Maybe add here the comment that we always take only the most serious violation, and as the limits are sorted in descending gravity that's why we take the first one

flo-dup (approved):

pom.xml
```diff
@@ -47,7 +47,6 @@
         <jafama.version>2.3.1</jafama.version>
         <junit-jupiter.version>5.5.2</junit-jupiter.version>
         <slf4jtoys.version>1.6.2</slf4jtoys.version>
-
```
annetill, line 50: I think you have to leave the space here (it is a detail I know, it is for readability).

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -112,7 +112,7 @@ public static OpenLoadFlowParameters getParametersExt(LoadFlowParameters paramet
         return parametersExt;
     }
 
-    private static SlackBusSelector getSlackBusSelector(Network network, LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
+    public static SlackBusSelector getSlackBusSelector(Network network, LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
```
annetill, line 115: Why do you need that change?

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
```diff
@@ -193,6 +186,19 @@ protected void checkTargetDeadband(RatioTapChanger rtc) {
...
+    protected static double getScaleForLimitType(LimitType type, LfBus bus) {
+        switch (type) {
+            case ACTIVE_POWER:
+            case APPARENT_POWER:
+                return 1.0 / PerUnit.SB;
+            case CURRENT:
+                return bus.getNominalV() / PerUnit.SB;
+            case VOLTAGE:
+            default:
+                throw new UnsupportedOperationException(String.format("Getting %s limits is not supported on dangling line branch.", type));
```
annetill, line 198: Why this message? What is the link with dangling line?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
obrix: I used it temporarily elsewhere and probably forgot to put it back to private when it was not needed any more. It can remain private.

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
obrix: Probably copy pasted the exception throwing too fast.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -237,7 +324,7 @@ void testFourSubstations() {
             .map(b -> new Contingency(b.getId(), new BranchContingency(b.getId())))
             .collect(Collectors.toList());
 
-        OpenSecurityAnalysis securityAnalysis = new OpenSecurityAnalysis(network, new DefaultLimitViolationDetector(),
```
annetill, line 240: To remain coherent with the rest of the code, we have to use the provider.

obrix: Yes I think there is still some work to adapt the provider to be more generic. I am on it.

flo-dup: Indeed, the provider should be used instead as it it part of the security analysis API, unlike the OpenSecurityAnalysis itself. I also did some unit tests with the OpenSecurityAnalysis in another PR, but after some thinking I think it is misleading: people tend to copy the code from unit tests to use powsybl. You should do something similar to what was done for sensitivity analysis, with the `AbstractSensitivityAnalysisTest` class.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisProvider.java
```diff
@@ -47,7 +47,7 @@ public OpenSecurityAnalysisProvider() {
                                                          LimitViolationFilter limitViolationFilter, ComputationManager computationManager,
                                                          SecurityAnalysisParameters securityAnalysisParameters, ContingenciesProvider contingenciesProvider,
                                                          List<SecurityAnalysisInterceptor> interceptors) {
-        OpenSecurityAnalysis osa = new OpenSecurityAnalysis(network, limitViolationDetector, limitViolationFilter, matrixFactory, connectivityProvider);
+        AbstractSecurityAnalysis osa = new AcSecurityAnalysis(network, limitViolationDetector, limitViolationFilter, matrixFactory, connectivityProvider);
```
flo-dup, line 50: You shoud run either Ac or Dc, based on `lfParameters.isDc()`, like it is done in `OpenSensitivityAnalysisProvider`

.github/workflows/maven.yml
```diff
@@ -11,6 +11,12 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
+      - name: Checkout powsybl-core sources
```
annetill, line 14: I imagine that you have to clean that now. `Maven.yml` must be the one from master branch.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisProvider.java
obrix: Yes this is what I did.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -0,0 +1,271 @@
...
+        // Only detect the most serious one (findFirst) : limit violations are ordered by severity
+        if (branch.getBus1() != null) {
+            branch.getLimits1(LimitType.CURRENT).stream()
+                .filter(temporaryLimit1 -> branch.getI1().eval() > temporaryLimit1.getValue())
+                .findFirst()
+                .map(temporaryLimit1 -> createLimitViolation1(branch, temporaryLimit1, LimitViolationType.CURRENT, PerUnit.SB / branch.getBus1().getNominalV(), branch.getI1().eval()))
```
annetill, line 136: Maybe an `getUnScaleForLimitType` should be great here?

src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
```diff
@@ -0,0 +1,98 @@
...
+        List<PostContingencyResult> postContingencyResults = new ArrayList<>();
+        for (Contingency contingency : contingencies) {
+            List<SensitivityValue2> values = res.getValues(contingency.getId());
+            List<LimitViolation> violations = new ArrayList<>();
+
+            for (SensitivityValue2 v : values) {
+                SensitivityFactor2 factor = (SensitivityFactor2) v.getFactorContext();
+                String branchId = factor.getFunctionId();
+                Branch branch = network.getBranch(branchId);
+                detector.checkActivePower(branch, Branch.Side.ONE, Math.abs(v.getFunctionReference()), violations::add);
```
annetill, line 90: Just a question: in the detector, do we have the rule as we have implemented here with `POST_CONTINGENCY_INCREASING_FACTOR`? If not, maybe we have to think at the implementation on core side.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
obrix: In AbstractLfBranch API ?

annetill: Yes? As you need with scale.

src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
obrix: POST_CONTINGENCY_INCREASING_FACTOR seems to be used as a post process to modify postContingencyLimitViolations list in relation with pre contingency list. As far as I know, there is nothing equivalent implemented yet in DC mode.

annetill (approved):

annetill:
Maybe some code smells car be resolved?

obrix:
The non generic ones have been fixed.

annetill (approved):

PR #315, opened 2021-06-01
Monitor security analysis

EtienneLt:
Signed-off-by: Etienne Lesot <etienne.lesot@rte-france.com>

#1699

**Other information**:


src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -225,10 +252,18 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
...
         // create a contingency list that impact the network
         List<LfContingency> contingencies = createContingencies(propagatedContingencies, network);
+        List<BranchResult> preContingencyBranchResults = new ArrayList<>();
+        List<BusResults> preContingencyBusResults = new ArrayList<>();
 
         // run pre-contingency simulation
         try (AcloadFlowEngine engine = new AcloadFlowEngine(network, acParameters)) {
             AcLoadFlowResult preContingencyLoadFlowResult = engine.run(Reporter.NO_OP);
+            monitors.values().forEach(stateMonitor -> {
+                if (stateMonitor.getContingencyContext().getContextType() == ContingencyContextType.ALL ||
```
annetill, line 262: @sylvlecl Reading again this, I am not sure that our choice on `powsybl-core` side is the good one...

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -302,6 +340,10 @@ private PostContingencyResult runPostContingencySimulation(LfNetwork network, Ac
                 network.getBranches().stream().filter(b -> !lfContingency.getBranches().contains(b)),
                 network.getBuses().stream().filter(b -> !lfContingency.getBuses().contains(b)),
                 postContingencyLimitViolations);
+            StateMonitor stateMonitor = monitors.get(lfContingency.getContingency().getId());
+            if (stateMonitor != null && stateMonitor.getContingencyContext().getContextType() == ContingencyContextType.SPECIFIC) {
```
annetill, line 344: Maybe you have to implement the `ALL` mode in a different way to be more efficient, but I am not sure about that.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -342,4 +386,22 @@ private static boolean violationWeakenedOrEquivalent(LimitViolation violation1,
...
+
+    private void addMonitorInfos(LfNetwork lfNetwork, StateMonitor monitor, Consumer<BranchResult> branchResultConsumer,
+                                 Consumer<BusResults> busResultsConsumer) {
+        monitor.getBranchIds().forEach(branchId -> branchResultConsumer.accept(createBranchResult(lfNetwork.getBranchById(branchId))));
+        lfNetwork.getBuses().stream().filter(lfBus -> monitor.getVoltageLevelIds().contains(lfBus.getVoltageLevelId())).forEach(lfBus ->
+            busResultsConsumer.accept(createBusResult(lfBus, lfBus.getVoltageLevelId())));
+    }
+
+    private BranchResult createBranchResult(LfBranch branch) {
+        double currentScale = PerUnit.SB / branch.getBus1().getNominalV();
```
annetill, line 398: Here I am not sure that it is robust to get the nominal voltage through the bus. I think you should try to put this method in the LfBranch directly and use attributes.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -342,4 +386,22 @@ private static boolean violationWeakenedOrEquivalent(LimitViolation violation1,
...
+        return new BranchResult(branch.getId(), branch.getP1().eval() * PerUnit.SB, branch.getQ1().eval(), branch.getI1().eval() * currentScale,
+            branch.getP2().eval() * PerUnit.SB, branch.getQ2().eval() * PerUnit.SB, branch.getI2().eval() * currentScale);
+    }
+
+    private BusResults createBusResult(LfBus bus, String voltageLevelId) {
```
annetill, line 403: Same here: please try to put that inside the LfBus.

annetill (changes requested):
Some changes are needed but globally it is correct.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
annetill: Add a unit test with open branches, dangline line and three windings transformer.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -259,9 +259,13 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, Loa
             LfBus lfBus1 = getLfBus(t3wt.getLeg1().getTerminal(), lfNetwork, parameters.isBreakers());
             LfBus lfBus2 = getLfBus(t3wt.getLeg2().getTerminal(), lfNetwork, parameters.isBreakers());
             LfBus lfBus3 = getLfBus(t3wt.getLeg3().getTerminal(), lfNetwork, parameters.isBreakers());
-            addBranch(lfNetwork, LfLegBranch.create(lfNetwork, lfBus1, lfBus0, t3wt, t3wt.getLeg1(), parameters.isTwtSplitShuntAdmittance()), report);
-            addBranch(lfNetwork, LfLegBranch.create(lfNetwork, lfBus2, lfBus0, t3wt, t3wt.getLeg2(), parameters.isTwtSplitShuntAdmittance()), report);
-            addBranch(lfNetwork, LfLegBranch.create(lfNetwork, lfBus3, lfBus0, t3wt, t3wt.getLeg3(), parameters.isTwtSplitShuntAdmittance()), report);
+            LfLegBranch lfBranch1 = LfLegBranch.create(lfNetwork, lfBus1, lfBus0, t3wt, t3wt.getLeg1(), parameters.isTwtSplitShuntAdmittance());
```
sylvlecl, line 262: The last argument (report) has been removed in those calls : I think it's not what you wanted to do !

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
```diff
@@ -205,4 +206,22 @@ public boolean isVoltageController() {
...
     }
+
+    @Override
+    public BranchResult createBranchResult() {
+        if (getBus1() == null && getBus2() == null) {
+            return new BranchResult(getId(), 0, 0, 0, 0, 0, 0);
+        } else if (getBus1() == null) {
+            double currentScale = PerUnit.SB / getBus2().getNominalV();
+            return new BranchResult(getId(), 0, 0, 0,
+                getP2().eval() * PerUnit.SB, getQ2().eval() * PerUnit.SB, getI2().eval() * currentScale);
```
sylvlecl, line 217: Probably to do in another PR, but I think it would make sense to share the scaling methods in the `PerUnit` class, what do you think @annetill @geofjamg ?
Like
```java
class PerUnit {
    double scaleP(double p) { ... }
    double unscaleP(double p)  { ... }
    double scaleI(double i, double nominalV)  { ... }
    double unscaleI(double i, double nominalV)  { ... }
    ...
```

It would reduce a little the boiler plate code and possibilities for mistakes.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -60,15 +68,32 @@
...
+    public OpenSecurityAnalysis(Network network, LimitViolationDetector detector, LimitViolationFilter filter,
+                                MatrixFactory matrixFactory, Supplier<GraphDecrementalConnectivity<LfBus>> connectivityProvider, List<StateMonitor> stateMonitors) {
         this.network = Objects.requireNonNull(network);
         this.detector = Objects.requireNonNull(detector);
         this.filter = Objects.requireNonNull(filter);
         this.matrixFactory = Objects.requireNonNull(matrixFactory);
         this.connectivityProvider = Objects.requireNonNull(connectivityProvider);
+        Objects.requireNonNull(stateMonitors);
+        this.monitors = new HashMap<>();
+        stateMonitors.forEach(monitor -> {
```
sylvlecl, line 89: let's not forget to re-use what you have done in core once it's done :)

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
sylvlecl: You mean to include the "ALL" into the N situation ?

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -269,7 +307,11 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
             }
 
             LimitViolationsResult preContingencyResult = new LimitViolationsResult(preContingencyComputationOk, new ArrayList<>(preContingencyLimitViolations.values()));
-            return new SecurityAnalysisResult(preContingencyResult, postContingencyResults);
+            return new SecurityAnalysisResult(preContingencyResult, postContingencyResults,
```
sylvlecl, line 310: Not a must-do, but maybe it would be easier to do building of the maps inside the constructor, so that we do not have to do it ourselves here

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -312,16 +361,20 @@ private PostContingencyResult runPostContingencySimulation(LfNetwork network, Ac
...
 
         LfContingency.reactivateEquations(deactivatedEquations, deactivatedEquationTerms);
-
         stopwatch.stop();
         LOGGER.info("Post contingency '{}' simulation done in {} ms", lfContingency.getContingency().getId(),
                 stopwatch.elapsed(TimeUnit.MILLISECONDS));
 
-        return new PostContingencyResult(lfContingency.getContingency(), postContingencyComputationOk, new ArrayList<>(postContingencyLimitViolations.values()));
+        return new PostContingencyResult(lfContingency.getContingency(), postContingencyComputationOk, new ArrayList<>(postContingencyLimitViolations.values()),
+            branchResults.stream().collect(Collectors.toMap(BranchResult::getBranchId, Function.identity())),
```
sylvlecl, line 369: Same here, maybe it would be simpler to change the constructor so that we just provide the lists

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -342,4 +395,13 @@ private static boolean violationWeakenedOrEquivalent(LimitViolation violation1,
     List<LfContingency> createContingencies(List<PropagatedContingency> propagatedContingencies, LfNetwork network) {
         return LfContingency.createContingencies(propagatedContingencies, network, network.createDecrementalConnectivity(connectivityProvider), true);
     }
+
+    private void addMonitorInfos(LfNetwork lfNetwork, StateMonitor monitor, Consumer<BranchResult> branchResultConsumer,
+                                 Consumer<BusResults> busResultsConsumer, Consumer<ThreeWindingsTransformerResult> threeWindingsTransformerResultConsumer) {
+        monitor.getBranchIds().forEach(branchId -> branchResultConsumer.accept(lfNetwork.getBranchById(branchId).createBranchResult()));
```
sylvlecl, line 401: I wonder what we should do in case of an invalid branch ID (one that does not existe in the network.

I guess that with this code, it will throw a null pointer exception. I would be in favor of silently ignore that branch instead.
In any case we must add a unit test to cover that case! (same for buses and T3)

src/main/java/com/powsybl/openloadflow/network/impl/LfThreeWindingsTransformer.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public LfThreeWindingsTransformer(String lfThreeWindingsTransformerId, LfLegBranch branch1, LfLegBranch branch2, LfLegBranch branch3) {
+        this.lfThreeWindingsTransformerId = Objects.requireNonNull(lfThreeWindingsTransformerId);
+        this.branch1 = Objects.requireNonNull(branch1);
+        this.branch2 = Objects.requireNonNull(branch2);
+        this.branch3 = Objects.requireNonNull(branch3);
+    }
+
+    public ThreeWindingsTransformerResult createThreeWindingsTransformerResult() {
+        return new ThreeWindingsTransformerResult(lfThreeWindingsTransformerId, branch1.getP1().eval(), branch1.getQ1().eval(),
```
sylvlecl, line 31: we probably need to handle the disconnected cases here, like for the branches (checking that buses are not null) : we'll need a unit test for this too.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -302,6 +346,11 @@ private PostContingencyResult runPostContingencySimulation(LfNetwork network, Ac
                 network.getBranches().stream().filter(b -> !lfContingency.getBranches().contains(b)),
                 network.getBuses().stream().filter(b -> !lfContingency.getBuses().contains(b)),
                 postContingencyLimitViolations);
+            StateMonitor stateMonitor = monitors.get(lfContingency.getContingency().getId());
+            if (stateMonitor != null && (stateMonitor.getContingencyContext().getContextType() == ContingencyContextType.ALL
+                || stateMonitor.getContingencyContext().getContextType() == ContingencyContextType.SPECIFIC)) {
+                addMonitorInfos(network, stateMonitor, branchResults::add, busResults::add, threeWindingsTransformerResults::add);
```
sylvlecl, line 352: Maybe there is something to do here to filter out buses and branches which are disconnected by the contingency, see filters above in the identification of violations.

Not 100% sure about that : @annetill and @geofjamg do you know if there is already a test case covering that ?

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
EtienneLt: look at the lines below report is there

sylvlecl: you're right, my mistake!

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
annetill: Yes, great idea, per unit system must be hidden. I will make it myself I think.

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
annetill: yes, use `isDisabled` for buses. For branches, the method has to be created.

src/main/java/com/powsybl/openloadflow/network/AbstractLfBranch.java
sylvlecl: An afterthought : actually it could even be an interface, that way we could use different per-unit systems and compare their effect  !! :)  But I admit, it's not really useful in practice ...

src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysis.java
```diff
@@ -225,10 +245,17 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
...
         // create a contingency list that impact the network
         List<LfContingency> contingencies = createContingencies(propagatedContingencies, network);
+        List<BranchResult> preContingencyBranchResults = new ArrayList<>();
+        List<BusResults> preContingencyBusResults = new ArrayList<>();
+        List<ThreeWindingsTransformerResult> preContingencyThreeWindingsTransformerResults = new ArrayList<>();
 
         // run pre-contingency simulation
         try (AcloadFlowEngine engine = new AcloadFlowEngine(network, acParameters)) {
             AcLoadFlowResult preContingencyLoadFlowResult = engine.run(Reporter.NO_OP);
+            addMonitorInfos(network, monitorIndex.getNoneStateMonitor(), preContingencyBranchResults::add, preContingencyBusResults::add,
```
annetill, line 255: I understand why it works but it could be more efficient to check the contingency context here, no?

geofjamg (approved):

annetill (approved):

PR #316, opened 2021-06-01
[WIP] Improve equation system listener 

annetill:

No. Indeed we have an issue when we have, after a contingency, a voltage discrete control that is modified. It could be modified by the number of controllers, or by the lost of the controlled bus. This is not supported. In the tests, I have also some `MAX_ITERATION_REACHED` that could be investigated.

**Other information**:


annetill:
After analyzing:
- If we lost a bus where we have generators, the listener does not change the voltage control. We have to put an equation system update maybe in the `setDisabled` method ?
- If we lost a branch where we have a controlling ratio tap changer or a controlling phase tap changer, discrete voltage control and phase control are not changed and the equation system does not change too. 
- If we lost a bus that is the controlled bus, we have to find something to update the equation system. Remove all the voltage control ?

PR #320, opened 2021-06-16
Don't check non null on debug method toString()

yichen88:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
@@ -141,7 +141,7 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
     public String toString() {
         return "OpenLoadFlowParameters(" +
                 "slackBusSelectionMode=" + slackBusSelectionMode +
-                ", slackBusId='" + Objects.requireNonNull(slackBusId, "") + "'" +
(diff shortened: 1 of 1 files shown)


flo-dup (approved):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -141,7 +141,7 @@ public static OpenLoadFlowParameters load() {
     public String toString() {
         return "OpenLoadFlowParameters(" +
                 "slackBusSelectionMode=" + slackBusSelectionMode +
-                ", slackBusId='" + Objects.requireNonNull(slackBusId, "") + "'" +
+                ", slackBusId='" + slackBusId + "'" +
```
geofjamg, line 144: The mistake is that Objects.requireNonNull has been used instead of Objects.toString

geofjamg (changes requested):

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
flo-dup: well, `", slackBusId='" + slackBusId + "'"` with a null `slackbusId` gives `, slackBusId='null'`, while your fix gives `, slackBusId=''`

PR #321, opened 2021-06-17
Fix bus/breaker LF network and non impedant branches

geofjamg:

**Other information**:


annetill:
@flo-dup I see in the code where we fix the discrete voltage controls with method `fixDiscreteVoltageControls`, but why we don't have any similar treatment for voltage controls? Furthermore, I think that discrete controls creation must depends on the loadflow parameters for performance issue.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -114,6 +114,10 @@ public boolean hasVoltageControllerCapability() {
         return Optional.ofNullable(voltageControl);
     }
 
+    public void removeVoltageControl() {
```
annetill, line 117: @flo-dup I imagine that it is not the good fix, but it works :-)

src/test/java/com/powsybl/openloadflow/NonImpedantBranchTest.java
```diff
@@ -171,8 +169,8 @@ void inconsistentTargetVoltagesTest() {
         createLine(network, b2, b3, "l23", 0.05);
         createLine(network, b3, b4, "l34", 0.05);
 
-        CompletionException exception = assertThrows(CompletionException.class, () -> loadFlowRunner.run(network, parameters));
-        assertEquals("Non impedant branch 'l12' is connected to PV buses 'b1_vl_0' and 'b2_vl_0' with inconsistent target voltages: 1.0 and 1.01", exception.getCause().getMessage());
+        // CompletionException exception = assertThrows(CompletionException.class, () -> loadFlowRunner.run(network, parameters));
```
annetill, line 172: I have an issue here because I have "removed" this check. It is in the code but never reached. It has to be moved but it could be nice to see that with you @flo-dup. Thanks for help !

src/test/java/com/powsybl/openloadflow/ac/NonImpedantBranchWithBreakerIssueTest.java
```diff
@@ -0,0 +1,73 @@
...
+                                                                             Collections.emptyList(), new DenseMatrixFactory(), false, false, false, false, false, breakers, ParameterConstants.PLAUSIBLE_ACTIVE_POWER_LIMIT_DEFAULT_VALUE,
+                                                                             false, true, Collections.emptySet(), true, Collections.emptySet(), false);
+        new AcloadFlowEngine(lfNetwork, acLoadFlowParameters)
+                .run();
+        lfNetwork.updateState(false, false, false, false, false, false);
+        for (Bus bus : network.getBusView().getBuses()) {
+            assertEquals(400, bus.getV(), 0);
+            assertEquals(0, bus.getAngle(), 0);
+        }
+        assertEquals(-200, network.getGenerator("G1").getTerminal().getQ(), 0);
```
annetill, line 51: @flo-dup my code is not deterministic aïe...

annetill (approved):

PR #323, opened 2021-06-22
Update open loadflow from sensitivity analysis api

tadam50:

No

Refactor

New sensitivity API is not used

Using new Sensitivity API (#1741)

**Other information**:


obrix:
@tadam50 Do you have any idea about why all the tests are failling ?

tadam50:
> @tadam50 Do you have any idea about why all the tests are failling ?

I'm not sure, but it's seems the list of SensitivityFactor provided to run method into class OpenSensitivityAnalysisProvider
is not well built
Otherwise some bus id are not found, i don't know why

obrix:
@tadam50 Regarding the bus id not found : for example in testEurostagFactory the BusBreakerView is used to build the SensitivityFactorsProvider while the checkBus function that is raising an exception use the BusView. I think the "Bus not found" exception is coming from here. (Tagging @flo-dup who is also looking at this PR)

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -628,7 +629,7 @@ protected boolean rescaleGlsk(List<SensitivityFactorGroup<V, E>> factorGroups, S
     protected void writeSkippedFactors(Collection<LfSensitivityFactor<V, E>> lfFactors, SensitivityValueWriter valueWriter) {
         List<LfSensitivityFactor<V, E>> skippedFactors = lfFactors.stream().filter(factor -> factor.getStatus() == LfSensitivityFactor.Status.SKIP).collect(Collectors.toList());
 
-        skippedFactors.forEach(factor -> valueWriter.write(factor.getContext(), null, -1, 0, Double.NaN));
+        skippedFactors.forEach(factor -> valueWriter.write((SensitivityFactor) factor.getContext(), null, -1, 0, Double.NaN));
```
annetill, line 632: Just for my learning: why do you need this cast?

obrix: Because for some reason AbstractLfSensitivityFactor has a generic non typed Object as context and not explicitly a SensitivityFactor. There may have been a reason for this at some point but I am not sur it is still relevant.

obrix: In fact SensitivityValueWriter did take a generic object as context parameter before it was moved to the core. This is why the cast was not needed before.

PR #325, opened 2021-06-29
Security analysis: disconnect terminals of isolated end nodes

flo-dup:

No

Feature

In security analysis, switches are retained even if they are encountered just before encountering a line / a 2wt while propagating the contingency.

Those switches are not retained anymore, leading to a simplified bus breaker view.

No


flo-dup:
What happens if while propagating a contingency, we encountered an openable switch which is just before a 3WT? The question being, is the 3WT still working when one of its leg is disconnected?

annetill (approved):

PR #326, opened 2021-07-01
Security analysis: support new branch result API

obrix:

No

Go further in DC mode security analyis : support new branch result elements in analysis result, also maybe support bus and bus results.

Branch result are not used and the analysis only support branch contingencies.

TBD

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
@@ -77,7 +77,6 @@ public class DcSecurityAnalysis extends AbstractSecurityAnalysis {
             List<SensitivityValue2> values = res.getValues(contingency.getId());
             List<LimitViolation> violations = new ArrayList<>();
             BranchResult preContRefBR = preContingencyBranchResults.get(contingency.getId());
-            double preContRefFlow = preContRefBR.getP1();
 
             for (SensitivityValue2 v : values) {
                 SensitivityFactor2 factor = (SensitivityFactor2) v.getFactorContext();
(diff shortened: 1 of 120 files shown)


src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
```diff
@@ -55,31 +57,48 @@ SecurityAnalysisReport runSync(final SecurityAnalysisParameters securityAnalysis
...
             for (SensitivityValue2 v : values) {
                 SensitivityFactor2 factor = (SensitivityFactor2) v.getFactorContext();
                 String branchId = factor.getFunctionId();
                 Branch<?> branch = network.getBranch(branchId);
+
+                if (monitor.getBranchIds().contains(branchId)) {
+                    BranchResult preContBR = preContingencyBranchResults.get(branchId);
+                    double preContingencyFlow = preContBR.getP1();
+                    double postContingencyFlow = v.getFunctionReference();
+                    double flowDelta = preContingencyFlow >= 0.0 ? postContingencyFlow - preContingencyFlow : preContingencyFlow - postContingencyFlow;
```
annetill, line 91: My understanding of the definition of a flow transfer on side 1 is that you have:
flow transfer =  (postContingencyFlow  - preContingencyFlow) / preContRefFlow

obrix: Yes that's what is computed when preContingencyFlow is positive. @Hadrien-Godard suggested inversing it when it is < 0.0, I am not sure about the convention here. @Hadrien-Godard do you have any idea ?

Hadrien-Godard: If preContingencyFlow is negative, for example it equals -100MW, it means, in DC, that on side 2, there is a flow of 100MW.

If the flow on this branch increases after the contingency, then the postContingencyFlow will equal, for example, -110MW. Therefore, postContingencyFlow - preContingencyFlow will equal -10MW.

Hence, if flow transfer = (postContingencyFlow - preContingencyFlow) / preContRefFlow, a negative value of a flow transfer does not mean that the flow on the line descreases after the contingency.

I have proposed a computation that ensures that the flow transfer is positive if and only if more power are flowing throught the line.

However, it may not be the correct definition of a flow transfer on side 1. So I suggest @obrix that you roll back to the previous computation with flow transfer = (postContingencyFlow - preContingencyFlow) / preContRefFlow, waiting further instructions from @annetill , althought results are harder to interpret with this definition.

annetill: Yes I understand your point and I was aware of that. I imagine that you way to compute is more user-friendly. Have you found some references of the definition of flow transfers in papers ? @murgeyseb do you have an opinion on that small computation ?

PR #340, opened 2021-07-23
Current limiter implementation

Hadrien-Godard:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
@@ -76,7 +76,7 @@ public class PhaseControlOuterLoop implements OuterLoop {
                 .filter(Optional::isPresent)
                 .map(Optional::get)
                 .filter(dpc -> dpc.getMode() == DiscretePhaseControl.Mode.LIMITER)
-                .filter(dpc -> detectUnstability(context, dpc) == OuterLoopStatus.UNSTABLE)
+                .filter(dpc -> changeTapPositions(context, dpc) == OuterLoopStatus.UNSTABLE)
                 .collect(Collectors.toList());
 
         return phaseControlsUnstable.isEmpty() ? OuterLoopStatus.STABLE : OuterLoopStatus.UNSTABLE;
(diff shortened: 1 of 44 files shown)


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemUpdater.java
```diff
@@ -130,4 +127,16 @@ public void onVoltageControlModeChange(DiscreteVoltageControl voltageControl, Di
             }
         }
     }
+
+    @Override
+    public void onPhaseControlTapChange(DiscretePhaseControl phaseControl) {
```
geofjamg, line 132: This should be handled in the `TargetVector` to just update the alpha1 target value. There is no need to change the equation system. => performance will also be better by just updating the target (no need to rebuild the jacobian matrix)

src/main/java/com/powsybl/openloadflow/network/PiModelArray.java
```diff
@@ -146,4 +146,32 @@ public void roundR1ToClosestTap() {
         }
         r1 = Double.NaN;
     }
+
+    @Override
+    public boolean getNewTapPosition(Direction direction) {
```
geofjamg, line 151: This should be `updateTapPosition`. A getter is not suppose to change internal state.
This is here that the network notification should be fired regading tap position change.
=> so it means that `PiModel` should be part of Network listenable data.

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
```diff
@@ -87,4 +91,38 @@ private void switchOffPhaseControl(DiscretePhaseControl phaseControl) {
...
+        boolean success = false;
+        if (phaseControl.getControlledSide() == DiscretePhaseControl.ControlledSide.ONE && currentLimit < controllerBranch.getI1().eval()) {
+            isSensibilityPositive = isSensitivityCurrentPerA1Positive(context.getVariableSet(), controllerBranch, DiscretePhaseControl.ControlledSide.ONE);
+            success = isSensibilityPositive ? piModel.getNewTapPosition(PiModel.Direction.DECREASE) : piModel.getNewTapPosition(PiModel.Direction.INCREASE);
+        } else if (phaseControl.getControlledSide() == DiscretePhaseControl.ControlledSide.TWO && currentLimit < controllerBranch.getI2().eval()) {
+            isSensibilityPositive = isSensitivityCurrentPerA1Positive(context.getVariableSet(), controllerBranch, DiscretePhaseControl.ControlledSide.TWO);
+            success = isSensibilityPositive ? piModel.getNewTapPosition(PiModel.Direction.DECREASE) : piModel.getNewTapPosition(PiModel.Direction.INCREASE);
+        }
+        if (success) {
+            for (LfNetworkListener listener : controllerBranch.getNetwork().getListeners()) {
```
geofjamg, line 110: Once notification will be managed inside `PiModel` we won't need this hack anymore because the call to `getNewTapPosition` will trigger the notif.

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemUpdater.java
annetill: I have made a fix and I think it is correct. I knew that it was not the good solution when updating the equation system. So I have used the listeners you have implemented and I am really impress by what you have implemented in the open-loadflow. It is really clever !

src/main/java/com/powsybl/openloadflow/network/PiModelArray.java
annetill: Fixed.

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
annetill: Done.

annetill (commented):
Hi! Thanks for your work @Hadrien-Godard, I have changed many lines of code, because it was not the perfect way to implement the current limiter, but functionally it was correct. Many we have now to go deeper into the details and increase tests. Thanks for the current derivatives ! I am going to test this feature on real case.

geofjamg (approved):

PR #342, opened 2021-07-30
[WIP] Handle deadband in PhaseControlOuterLoop

obrix:
Handle deadband in PhaseControlOuterLoop check process. Deadband is supported only when controlled branch is the same as controller branch.

Handle deadband parameter provided in DiscretePhaseControl parameter for switchOffPhaseControl function.

Deadband paramater is ignored.

Deadband parameter is taken into account but only if controllerBranch is equal to the controlledBranch.

**Other information**:


annetill:
I close this PR because nothing has been done on it since months. I think that if we really want to deal with target deadband for phase control, we have to run a first resolution without any phase control and check after if it is worth controlling or not, depending on the deadband value.

PR #349, opened 2021-08-30
Fix reference flow that were not computed and set as NaN when variable element is not in the main component but function element is

Hadrien-Godard:

**Other information**:


annetill (changes requested):
Can you please check my two last commits and increase the test coverage ?

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -50,6 +50,8 @@ private void calculateSensitivityValues(List<LfSensitivityFactor<AcVariableType,
                                             String contingencyId, int contingencyIndex, SensitivityValueWriter valueWriter) {
         Set<LfSensitivityFactor<AcVariableType, AcEquationType>> lfFactorsSet = new HashSet<>(lfFactors);
         lfFactors.stream().filter(factor -> factor.getStatus() == LfSensitivityFactor.Status.ZERO).forEach(factor -> valueWriter.write(factor.getContext(), contingencyId, contingencyIndex, 0, Double.NaN));
+        lfFactors.stream().filter(factor -> factor.getStatus() == LfSensitivityFactor.Status.SKIP_ONLY_VARIABLE).forEach(factor -> valueWriter.write(factor.getContext(), contingencyId, contingencyIndex, 0, unscaleFunction(factor, factor.getFunctionReference())));
```
annetill, line 53: This is not consistent with the `Double.NaN` chosen as value for `SKIP` factor.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -244,6 +244,27 @@ protected void setBaseCaseSensitivityValues(List<SensitivityFactorGroup<DcVariab
...
+            flowValue += contingencyElement.getAlphaForFunctionReference() * contingencySensitivity;
+            if (contingencyElement.getElement().getId().equals(functionBranchId)) {
+                // the monitored branch is in contingency, its post-contingency flow equals to zero.
+                flowValue = 0d;
+                break;
+            }
+        }
+
+        valueWriter.write(factor.getContext(), contingency != null ? contingency.getContingency().getId() : null, contingency != null ? contingency.getIndex() : -1,
+                0d, unscaleFunction(factor, flowValue));
```
annetill, line 265: Here we choose `0` for sensitivity value, it is what you want? Or is it better to choose `Double.NaN`?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -805,14 +858,15 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
...
 
-                // !!! we need to reset predefined values for factor that need to be calculated for this set of contingency
+                // we need to reset predefined values for factor that need to be calculated for this set of contingency
                 // predefined may have be set by a previous set of contingency with loss of connectivity
-                lfFactorsForContingencies.forEach(factor -> factor.setPredefinedResult(null));
-
-                connectivityAnalysisResult.setPredefinedResults();
+                lfFactorsForContingencies.forEach(factor -> factor.setSensitivityValuePredefinedResult(null));
+                lfFactorsForContingencies.forEach(factor -> factor.setFunctionPredefinedResult(null));
+                connectivityAnalysisResult.setSensitivityValuePredefinedResults();
```
annetill, line 868: Why do you need these two lines?

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -604,11 +633,33 @@ public void cutConnectivity(LfNetwork lfNetwork, GraphDecrementalConnectivity<Lf
...
+            if (factor.getStatus() == LfSensitivityFactor.Status.VALID) {
+                // after a contingency, we check if the factor function and the variable are in different connected components
+                boolean variableConnected = factor.isVariableConnectedToSlackComponent(connectedComponent);
+                boolean functionConnected = factor.isFunctionConnectedToSlackComponent(connectedComponent);
+                if (!variableConnected && functionConnected) {
+                    // VALID_ONLY_FOR_FUNCTION status
+                    factor.setSensitivityValuePredefinedResult(0d);
+                }
+                if (!variableConnected && !functionConnected) {
+                    // SKIP status
```
annetill, line 645: Here I think that this level of details is not needed: in case of factor with SKIP status, in pre-contingency state, you write `NaN` for sensi and `NaN` for function. But you introduce a new level of detail after a contingency. It is not coherent and I think that if it is more right, it is less perfomant.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -129,7 +134,9 @@ protected static Terminal getEquipmentRegulatingTerminal(Network network, String
 
         boolean areVariableAndFunctionDisconnected(GraphDecrementalConnectivity<LfBus> connectivity);
 
-        boolean isConnectedToComponent(Set<LfBus> connectedComponent);
+        boolean isVariableConnectedToSlackComponent(Set<LfBus> connectedComponent);
```
annetill, line 137: What is for me a bit hard to understand if that these two functions are not used for pre-contingency state. It is an incoherent behavior, no?

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -628,7 +658,9 @@ protected boolean rescaleGlsk(List<SensitivityFactorGroup<V, E>> factorGroups, S
     protected void writeSkippedFactors(Collection<LfSensitivityFactor<V, E>> lfFactors, SensitivityValueWriter valueWriter) {
         List<LfSensitivityFactor<V, E>> skippedFactors = lfFactors.stream().filter(factor -> factor.getStatus() == LfSensitivityFactor.Status.SKIP).collect(Collectors.toList());
 
-        skippedFactors.forEach(factor -> valueWriter.write(factor.getContext(), null, -1, 0, Double.NaN));
+        // SKIP factors are for factors where both variable and function elements are not in the main connected componant.
+        // Therefore, their sensitivity and reference values are set to NaN.
+        skippedFactors.forEach(factor -> valueWriter.write(factor.getContext(), null, -1, Double.NaN, Double.NaN));
```
annetill, line 663: See, for base case, SKIP is always NaN and NaN.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
Hadrien-Godard: I agree with you.
Indeed, in precontingency network, skip status is used when both variable and function are disconnected from slack and the sensitivity is set to NaN.
But in case of a contingency breaking connectivity we treat the case where both function and element are disconnected from slack and are in different connected components with a 0 value for the sensitivity.

This behavior is not coherent.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
Hadrien-Godard: Those two functions are not used for pre-contingency state indeed and we only check if variable and function element are in slack connected component. This may result to an incoherent behavior regarding SKIP status as discussed in another comment.

phiedw (approved):

PR #365, opened 2021-10-03
Fix cyclic dependencies

geofjamg:

**Other information**:


src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -138,9 +137,16 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
                 }
 
                 // start a simulation for each of the contingency
-                Iterator<LfContingency> contingencyIt = contingencies.iterator();
+                Iterator<PropagatedContingency> contingencyIt = propagatedContingencies.iterator();
```
flo-dup, line 140: Aside this PR, but I'm using it to ask: do we need a iterator? Shouldn't we restore the pre-contingency state at the end?

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public double getActivePowerLoss() {
+        return activePowerLoss;
+    }
+
+    public static Optional<LfContingency> create(PropagatedContingency propagatedContingency, LfNetwork network,
+                                                 GraphDecrementalConnectivity<LfBus> connectivity, boolean useSmallComponents) {
+        // find contingency branches that are part of this network
+        Set<LfBranch> branches = new HashSet<>(1);
+        for (String branchId : propagatedContingency.getBranchIdsToOpen()) {
```
flo-dup, line 74: Oh, good not to remove the processed branches from the list nor the PropagatedContingency! didn't really like it before

flo-dup (approved):

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -138,9 +137,16 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
...
                 while (contingencyIt.hasNext()) {
-                    LfContingency lfContingency = contingencyIt.next();
+                    PropagatedContingency propagatedContingency = contingencyIt.next();
+                    LfContingency lfContingency = LfContingency.create(propagatedContingency, network, network.createDecrementalConnectivity(connectivityProvider), true)
+                            .orElse(null);
+
+                    // only process contingencies that impact the network
+                    if (lfContingency == null) {
```
flo-dup, line 147: A bit strange to test the nullity instead of the `Optional` itself.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
geofjamg: This is an optimization because after last contingency simulation, we don't need to restore base case state.
No idea if it worse it or not...

flo-dup (approved):

PR #366, opened 2021-10-05
Adapt code for a reactive limits private outer loop

annetill:

Some small adaptations are needed to design a private reactive outerloops. 

**Other information**:


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemUpdater.java
```diff
@@ -58,6 +58,7 @@ private void updateControlledBus(VoltageControl voltageControl, EquationSystem<A
             // clean reactive power distribution equations
             controllerBuses.forEach(b -> equationSystem.removeEquation(b.getNum(), AcEquationType.ZERO_Q));
             // is there one static var compensator with a non-zero slope
+            // EquationTerm<AcVariableType, AcEquationType> vTerm = EquationTerm.createVariableTerm(controlledBus, AcVariableType.BUS_V, variableSet, controlledBus.getV().eval());
```
annetill, line 61: This has to be checked. Why do I need this line in my first implementation of stand by automaton?

src/main/java/com/powsybl/openloadflow/network/LfGenerator.java
```diff
@@ -76,4 +76,14 @@ default void setSlope(double slope) {
     Object getUserObject();
 
     void setUserObject(Object userObject);
+
+    enum GeneratorControlType {
```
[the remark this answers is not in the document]
annetill: I have removed the generator type `MONITORING_VOLTAGE`.

src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
```diff
@@ -60,6 +75,6 @@ public Evaluable getQ() {
 
     @Override
     public void updateState() {
-        shuntCompensator.getTerminal().setQ(q.eval() * PerUnit.SB);
+        terminal.setQ(q.eval() * PerUnit.SB);
```
annetill, line 78: @geofjamg I have some doubts about this line because in case of a static var compensator in control mode `VOLTAGE` the update of reactive power should be the one of the generator and not this one. But if the static var compensator becomes PQ, it could remain a shunt and the update should be this one.

src/test/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImplTest.java
```diff
@@ -142,14 +142,18 @@ void defaultMethodsTest() {
...
     @Test
-    void defaultMethodsTest2() {
-        network = DanglingLineFactory.create();
+    void defaultMethodsTest4() {
+        network = EurostagTutorialExample1Factory.create();
         List<LfNetwork> lfNetworks = Networks.load(network, new FirstSlackBusSelector());
         assertEquals(1, lfNetworks.size());
 
         LfNetwork mainNetwork = lfNetworks.get(0);
-        LfBus lfDanglingLineBus = mainNetwork.getBusById("dl1_BUS");
```
annetill, line 151: Indeed I need this test `defaultMethodsTest2` for droop coverage.

src/main/java/com/powsybl/openloadflow/network/impl/LfStaticVarCompensatorImpl.java
```diff
@@ -102,9 +107,10 @@ public double getMaxP() {
 
     @Override
     public void updateState() {
+        double vSquare = bus.getV() * bus.getV() * bus.getNominalV() * bus.getNominalV();
         svc.getTerminal()
                 .setP(0)
-                .setQ(Double.isNaN(calculatedQ) ? svc.getReactivePowerSetpoint() : -calculatedQ);
+                .setQ((Double.isNaN(calculatedQ) ? -targetQ : -calculatedQ) - optionalB * vSquare);
```
annetill, line 113: Here: in case of calculatedQ equals to NaN, it means that the static var compensator does not regulate voltage and so, the optional B should be equal to zero.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -711,6 +713,8 @@ private static LfNetwork create(int numCC, int numSC, List<Bus> buses, List<Swit
             }
         }
 
+        postProcessors.forEach(pp -> pp.onNetworkAdded(lfNetwork));
```
annetill, line 716: Indeed it could be improved through a post processing on voltage controls only.

PR #390, opened 2021-11-30
Fix DcValueVoltageInitializer

geofjamg:

No

Bug fix

There is too much hard coded parameters for the DC LF use for voltage angle init.

Only relevant parameters remains hard coded

**Other information**:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
         <junit-jupiter.version>5.5.2</junit-jupiter.version>
         <slf4jtoys.version>1.6.2</slf4jtoys.version>
 
-        <powsybl-core.version>4.6.0-SNAPSHOT</powsybl-core.version>
+        <powsybl-core.version>4.5.0</powsybl-core.version>
     </properties>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
         <junit-jupiter.version>5.5.2</junit-jupiter.version>
         <slf4jtoys.version>1.6.2</slf4jtoys.version>
 
-        <powsybl-core.version>4.5.0</powsybl-core.version>
+        <powsybl-core.version>4.6.0-SNAPSHOT</powsybl-core.version>
     </properties>
 
     <build>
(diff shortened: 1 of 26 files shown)


geofjamg:
@annetill last commit fix the case_SyntheticUSA.mat divergence. It was because of generator active power target, not saved/restored by first DC LF

annetill:
> @annetill last commit fix the case_SyntheticUSA.mat divergence. It was because of generator active power target, not saved/restored by first DC LF

You mean that we have changed the targetP of generators without restoring it ? So it means that we have the same problem with loads in case of slack distribution on loads. So, I think that maybe the best fix is to use `BusState` as we do in security analysis (there is a dedicate method for active variables).

geofjamg:
> > @annetill last commit fix the case_SyntheticUSA.mat divergence. It was because of generator active power target, not saved/restored by first DC LF
> 
> You mean that we have changed the targetP of generators without restoring it ? So it means that we have the same problem with loads in case of slack distribution on loads. So, I think that maybe the best fix is to use `BusState` as we do in security analysis (there is a dedicate method for active variables).

We cannot because BusState also save/restore voltage angle.

annetill:
> > > @annetill last commit fix the case_SyntheticUSA.mat divergence. It was because of generator active power target, not saved/restored by first DC LF
> > 
> > 
> > You mean that we have changed the targetP of generators without restoring it ? So it means that we have the same problem with loads in case of slack distribution on loads. So, I think that maybe the best fix is to use `BusState` as we do in security analysis (there is a dedicate method for active variables).
> 
> We cannot because BusState also save/restore voltage angle.

Yes I know, but we can maybe refactor that BusState for this prupose?

geofjamg:
> > > > @annetill last commit fix the case_SyntheticUSA.mat divergence. It was because of generator active power target, not saved/restored by first DC LF
> > > 
> > > 
> > > You mean that we have changed the targetP of generators without restoring it ? So it means that we have the same problem with loads in case of slack distribution on loads. So, I think that maybe the best fix is to use `BusState` as we do in security analysis (there is a dedicate method for active variables).
> > 
> > 
> > We cannot because BusState also save/restore voltage angle.
> 
> Yes I know, but we can maybe refactor that BusState for this prupose?

Done

annetill (approved):
Nice fix !

Issue #107, opened 2020-06-05
Modeling of shunt compensators voltage control


annetill:
* **Do you want to request a *feature* or report a *bug*?**

A feature.

* **What is the current behavior?**

Shunts have no regulating capabilities.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

Shunts with multiple sections could have local or remote regulating capabilities. Shunts with just one section can be automatically "connected" or "disconnected" if needed to respect a target.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

Details of the algorithm have to be specified before. 
IIDM does not support non linear shunt compensator for the moment.

PR #391, opened 2021-12-01
Shunt compensator voltage control

annetill:

**Does this PR already have an issue describing the problem ? If so, link to this issue using '#XXX' and skip the rest

Yes, an issue: #107

**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)

For the moment, this PR is a first minimal feature that supports:

- Voltage control for a single shunt connected to a bus: local and remote control ;
- Voltage control for several shunts connected to a bus: local and remote control. But, the algorithm to dispatch B has to be improved (it is instead more an optimization that has to be implemented) ;
- But we don't support shared remote control between several controller buses. 
- Then, in case of a bus controlled by generators or transformers and a shunt compensator, the shunt regulation is disabled. It is the first approach.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
@@ -53,6 +53,10 @@ public class LfShuntImpl extends AbstractElement implements LfShunt {
             this.bMagnitude = Math.abs(bMax - bMin);
         }
 
+        public String getId() {
+            return id;
+        }
+
         public List<Double> getSections() {
             return sections;
         }
@@ -71,10 +75,6 @@ public class LfShuntImpl extends AbstractElement implements LfShunt {
 
         public double getBMagnitude() {
             return bMagnitude;
-        }
-
-        public String getId() {
-            return id;
         }
     }
 


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -108,10 +108,28 @@ private static void createReactivePowerControlBranchEquation(LfBranch branch, Re
...
+        List<LfShunt> controllerShunts = bus.getControllerShunts();
+        if (controllerShunts.isEmpty()) {
+            for (LfShunt shunt : bus.getShunts()) {
+                ShuntCompensatorReactiveFlowEquationTerm q = new ShuntCompensatorReactiveFlowEquationTerm(shunt, bus, variableSet, false);
+                equationSystem.createEquation(bus.getNum(), AcEquationType.BUS_Q).addTerm(q);
+                shunt.setQ(q);
+            }
+        } else {
+            Optional<DiscreteVoltageControl> dvc = bus.getDiscreteVoltageControl();
+            if (dvc.isPresent() && dvc.get().getControllerBuses().contains(bus)) {
```
annetill, line 120: This part is the most hard to understand and probably the implementation is not the good one. In case of several shunts, instead of creating several B variables, I have decided to create just one variable B that takes all we need to control voltage and to redispatch the amount of B on the physical shunts after the resolution.

geofjamg: You have to use only one b variable, but maybe you could improve your implementation by spreading the b into each shunt by just adding a scaler to the term.

src/main/java/com/powsybl/openloadflow/network/DiscreteVoltageControl.java
```diff
@@ -17,15 +17,18 @@
 public class DiscreteVoltageControl {
 
     public enum Mode {
-        VOLTAGE,
+        VOLTAGE_TRANSFORMER,
+        VOLTAGE_SHUNT,
```
geofjamg, line 21: What was the motivation to share the same object control with transformer one?
This look to me 2 different things.
The implementation looks now very confusing to me as we have 2 modes and 2 controller lists and that both have to be consistent.
What about renaming existing `DiscreteVoltageControl` to `TransformerVoltageControl` and adding a new `ShuntCompensatorVoltageControl` ?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -162,6 +162,7 @@ public static AcLoadFlowParameters createAcParameters(Network network, MatrixFac
         LOGGER.info("Connected component mode: {}", parameters.getConnectedComponentMode());
         LOGGER.info("Voltage per reactive power control: {}", parametersExt.isVoltagePerReactivePowerControl());
         LOGGER.info("Reactive Power Remote control: {}", parametersExt.hasReactivePowerRemoteControl());
+        LOGGER.info("Shunt voltage control: {}", parameters.isSimulShunt());
```
geofjamg, line 165: A comment for the loadflow api: "simulShunt" is poorly named...
For a user point of view, what does "simulate shunts" means? It does not appear clearly that we are talking about shunt voltage control.

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
annetill: Can you explain with more details what you have in mind ? In my implementation, I use only one `b` variable on a bus, but if I have several shunts, I have to find the best combination of the sections of the shunts that is close to the solved `b`. How can we do that through a scalar ?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
annetill: Yes, you are right. I have created an issue : https://github.com/powsybl/powsybl-core/issues/1929

src/main/java/com/powsybl/openloadflow/network/DiscreteVoltageControl.java
annetill: Okay, no problem to do that.

annetill:
@geofjamg With your last merged PRs, it is my third huge rebase for that PR: I guess that it is my Christmas present!

geofjamg:
> @geofjamg With your last merged PRs, it is my third huge rebase for that PR: I guess that it is my Christmas present!

I will do it

annetill:
> > @geofjamg With your last merged PRs, it is my third huge rebase for that PR: I guess that it is my Christmas present!
> 
> I will do it

No that is okay, indeed I have time to do it and it will be a good way to understand your recent changes.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
geofjamg: So we cannot just have a proportional distribution of b across all shunts directly in the equation system? 
I don't remember why we create as many LfShunt by bus as in the IIDM network. Maybe it would be simpler to have only one LfShunt by LfBus that would aggregate IIDM shunt compensator. In the case of non regulating shunt compensators that would only required in the updateState method to distribute the q proportionally to individual b and in your regulating shunt case that would be simpler to model.

src/main/java/com/powsybl/openloadflow/ac/ShuntVoltageControlOuterLoop.java
```diff
@@ -0,0 +1,51 @@
...
+import com.powsybl.openloadflow.network.AbstractDiscreteVoltageControl;
+import com.powsybl.openloadflow.network.ShuntVoltageControl;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * @author Anne Tilloy <anne.tilloy at rte-france.com>
+ */
+public class ShuntVoltageControlOuterLoop implements OuterLoop {
```
geofjamg, line 22: Are you sure we need an outerloop? I do not see like for transformer a first step with continuous variable for susceptance then rounding and then we re-run with rounded values?

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/ac/ShuntVoltageControlOuterLoop.java
annetill: Yes I do, but the global design of this PR was bad, so I have made some fixes to fit with what we do for transformers. It looks better, even if I think that we can improve the way we model shunts, maybe in two equations max by bus. I have made some tests on real test cases (Elia and Mavir) and for both I have an issue that says that I have too many variables for the number of equations directly in the first iteration:
For Elia:
`
09:05:37.442 [ForkJoinPool.commonPool-worker-3] ERROR  c.p.openloadflow.ac.nr.NewtonRaphson - com.powsybl.commons.PowsyblException: Expected to have same number of equations (1630) and variables (1631)
com.powsybl.commons.PowsyblException: Expected to have same number of equations (1630) and variables (1631)
`
Any idea?

annetill: Fixed, it was a serious bug in the network loading. A bus was both controlled by a generator and shunts, but it was not detected during the loading. Now, it works on both real networks.

src/main/java/com/powsybl/openloadflow/ac/ShuntVoltageControlOuterLoop.java
```diff
@@ -0,0 +1,76 @@
...
+            return shuntVoltageControls.isEmpty() ? OuterLoopStatus.STABLE : OuterLoopStatus.UNSTABLE;
+        }
+        return status;
+    }
+
+    private void switchOffVoltageControl(ShuntVoltageControl vc) {
+
+        for (LfBus controllerBus : vc.getControllers()) {
+            // round the rho shift to the closest tap
+            List<LfShunt> controllerShunts = controllerBus.getControllerShunts();
```
annetill, line 55: I am obliged to OFF the VC after this dispatching.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
annetill: The distribution could not be proportional to be optimal because each shunt have sections (in a different number too), that could be non linear too. So I have to fill the shunt one by one to find the optimal combinaison that is the closest to the continuous value. But for sure, it would be easier to have a single LfShunt by a bus, with one or two shunt equations: one for the constant B shunts and the other one for the regulating ones. But it is a huge change... and maybe the refactoring should be done before or after this PR.

src/main/java/com/powsybl/openloadflow/network/DiscreteVoltageControl.java
annetill: Done.

annetill: @flo-dup I think that we can simplify the code now by removing the notion of "discrete" for phase controls.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -192,7 +196,11 @@ static void updateRemoteVoltageControlEquations(VoltageControl voltageControl, E
             terms.add(q);
         }
         controllerBus.getShunt().ifPresent(shunt -> {
-            ShuntCompensatorReactiveFlowEquationTerm q = new ShuntCompensatorReactiveFlowEquationTerm(shunt, controllerBus, variableSet);
+            ShuntCompensatorReactiveFlowEquationTerm q = new ShuntCompensatorReactiveFlowEquationTerm(shunt, controllerBus, variableSet, false);
+            terms.add(q);
+        });
+        controllerBus.getControllerShunt().ifPresent(shunt -> {
```
annetill, line 202: I am not totally sure of this term.

src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
```diff
@@ -6,35 +6,119 @@
...
+
+    private double b;
 
     protected LfBus bus;
 
+    private final List<ControllerLfShunt> controllerLfShunts = new ArrayList<>();
+
+    private double zb;
+
+    private static class ControllerLfShunt {
```
geofjamg, line 37: Not clear to me why you need to create this intermediate data model

src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
```diff
@@ -6,35 +6,119 @@
...
+        }
+    }
+
     public LfShuntImpl(List<ShuntCompensator> shuntCompensators, LfNetwork network) {
         super(network);
         if (shuntCompensators.isEmpty()) {
             throw new IllegalArgumentException("Empty shunt compensator list");
         }
-        this.shuntCompensators = Objects.requireNonNull(shuntCompensators);
+        for (ShuntCompensator sc : shuntCompensators) {
```
geofjamg, line 86: But looking at how you build ShuntImpl, there either only shunts controlling voltage or only shunts not controlling voltage?

geofjamg (commented):
- Maybe there is room for simplying the `LfShuntImpl`code
- I suspect the shared remote control is not working. Like for any other shared control you need to add extra equation for b distribution => as there is no real use case and as it will complexity the code, I would not support it now but throws an exception instead and only support local and non shared remote control.
- My main concern is that there is exactly the same functional issue as for transformer voltage control. There is no management of min/max susceptance. Like for transformers at a given loop, if some shunt reach the min or max b, they should be switched to fixed value (min ou max value) and the other ones should continue to control voltage at next loop. Everything stop when no other shunt reach min or max or if all shunts have reached min or max b. Also control mode should not be managed at control object level but at each of the controller bus so that control mode status is the result of individual controller status. This will allow for a given control to have at some moment, part of the controllers controlling and other ones not (if at min/max).  But maybe we should let it like this and fix both transformer and shunt voltage control afterward in another PR.

src/main/java/com/powsybl/openloadflow/network/AbstractLfNetworkListener.java
```diff
@@ -22,7 +22,7 @@ public void onDiscretePhaseControlModeChange(DiscretePhaseControl phaseControl,
     }
 
     @Override
-    public void onDiscreteVoltageControlModeChange(DiscreteVoltageControl voltageControl, DiscreteVoltageControl.Mode oldMode, DiscreteVoltageControl.Mode newMode) {
+    public void onDiscreteVoltageControlModeChange(DiscreteVoltageControl voltageControl, DiscreteVoltageControl.Mode newMode) {
```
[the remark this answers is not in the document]
annetill: Just because it is not used in the code.

src/main/java/com/powsybl/openloadflow/ac/equations/ShuntCompensatorReactiveFlowEquationTerm.java
```diff
@@ -56,12 +59,20 @@ private double v() {
         return stateVector.get(vVar.getRow());
     }
 
+    private double b() {
+        return bVar != null ? stateVector.get(bVar.getRow()) : shunt.getB(); // FIXME
```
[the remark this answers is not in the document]
annetill: No, I am going to remove it, thanks!

src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
annetill: Yes, so I think that you are right I can simply the `LfShuntImpl` knowing that.

annetill:
> * Maybe there is room for simplying the `LfShuntImpl`code

I am going to try to have only one list of ShuntCompensator but it am not sure I can do better without loosing in readability. 

> * I suspect the shared remote control is not working. Like for any other shared control you need to add extra equation for b distribution => as there is no real use case and as it will complexity the code, I would not support it now but throws an exception instead and only support local and non shared remote control.

No it is not supported, I have to commit my work: it is "easy" to support it as we do for transformers but it has to be refactor later.

> * My main concern is that there is exactly the same functional issue as for transformer voltage control. There is no management of min/max susceptance. Like for transformers at a given loop, if some shunt reach the min or max b, they should be switched to fixed value (min ou max value) and the other ones should continue to control voltage at next loop. Everything stop when no other shunt reach min or max or if all shunts have reached min or max b. Also control mode should not be managed at control object level but at each of the controller bus so that control mode status is the result of individual controller status. This will allow for a given control to have at some moment, part of the controllers controlling and other ones not (if at min/max).  But maybe we should let it like this and fix both transformer and shunt voltage control afterward in another PR.

Yes, I know because you explain me that for transfromers. But I really prefer that we merge this PR first and deal with that later (just after this one). The rebases are really really time consuming for me. For the control mode, I understand that you want me to add a `hasVoltageControl` boolean in `LfShunt` and in a transformer, as we do for generators?

geofjamg:
> No it is not supported, I have to commit my work: it is "easy" to support it as we do for transformers but it has to be refactor later.
> 
> > *

If you can, try to do it using just activation and not equation removal.

geofjamg:
> Yes, I know because you explain me that for transfromers. But I really prefer that we merge this PR first and deal with that later (just after this one). The rebases are really really time consuming for me. For the control mode, I understand that you want me to add a `hasVoltageControl` boolean in `LfShunt` and in a transformer, as we do for generators?

ok

src/main/java/com/powsybl/openloadflow/network/AbstractLfNetworkListener.java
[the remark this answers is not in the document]
geofjamg: I would keep it as it can be useful later

geofjamg (approved):

annetill:
:-)

PR #408, opened 2021-12-26
Refactor shared voltage control modelling

geofjamg:

No

Refactoring

Shared voltage control is modeled using one PQV bus at controlled bus and (Ngen - 1) reactive distribution equations at all controller buses except one. These equations allows to define reactive power distribution across all voltage controller buses.  
The problem with such a modeling is that when one controller bus of the shared control switch PQ (reactive limit bound reached), as reactive power equations have been defined 2 by 2 using an arbitrary reference controller bus we have to remove all reactive power distribution equations and re-create all of them using potentially a new reference controller bus.
This is a heavy operation and it should be simpler if we can just active/deactivate some predefined equations. Furthermore for later performance optimization we will try to define alternative equations (i.e couple of equation that when one is activate the other is deactivate automatically and vice versa) so that we only have to update Jacobian matrix when a bus is switched PV/PQ ou PQ/PV and not fully rebuild it.

TODO: same for transformer voltage control has to be done in this PR

We only need to deactivate/activate right predefined equations when switch a PV to PQ or PQ to PV bus. There is no more equation removal. 

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
@@ -11,10 +11,7 @@ import com.powsybl.openloadflow.equations.*;
 import com.powsybl.openloadflow.network.*;
 import com.powsybl.openloadflow.network.DiscretePhaseControl.Mode;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import java.util.stream.Collectors;
 
 /**
@@ -124,7 +121,7 @@ public final class AcEquationSystem {
         controlledBus.setCalculatedV(vTerm);
 
         // create reactive power distribution equations at voltage controller buses
-        createReactivePowerDistributionEquations(new ArrayList<>(voltageControl.getControllerBuses()), networkParameters, equationSystem, creationParameters);
+        createReactivePowerDistributionEquations(voltageControl.getControllerBuses(), networkParameters, equationSystem, creationParameters);
 
         updateRemoteVoltageControlEquations(voltageControl, equationSystem);
     }
@@ -154,7 +151,8 @@ public final class AcEquationSystem {
         // activate reactive power distribution equation at all enabled controller buses except one
         for (int i = 0; i < enabledControllerBuses.size(); i++) {
             LfBus controllerBus = enabledControllerBuses.get(i);
-            var qDistrEq = equationSystem.getEquation(controllerBus.getNum(), AcEquationType.DISTR_Q).orElseThrow();
+            var qDistrEq = equationSystem.getEquation(controllerBus.getNum(), AcEquationType.DISTR_Q)
+                    .orElseThrow();
             qDistrEq.setActive(i != 0);
         }
     }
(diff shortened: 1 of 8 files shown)


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -72,8 +65,7 @@ private static void createGeneratorControlEquations(LfBus bus, LfNetworkParamete
             if (voltageControl.isVoltageControlLocal()) {
                 createLocalVoltageControlEquation(bus, networkParameters, equationSystem, creationParameters);
             } else if (bus.isVoltageControlled()) {
-                // remote controlled: set voltage equation on this controlled bus
-                createVoltageControlledBusEquations(voltageControl, networkParameters, equationSystem, creationParameters);
+                createRemoteVoltageControlEquations(voltageControl, networkParameters, equationSystem, creationParameters);
```
annetill, line 68: If I remember well we have also a case where we create distribution equations for remote and local controllers. We fall in that second case. You can keep `remote` in the name of the method, but maybe I can add it in the comment.

src/main/java/com/powsybl/openloadflow/network/LfBus.java
```diff
@@ -152,4 +152,8 @@ default boolean isParticipating() {
      * Find bus + parallel branches neighbors.
      */
     Map<LfBus, List<LfBranch>> findNeighbors();
+
+    double getRemoteControlReactivePercent();
```
annetill, line 156: Can you add `voltage` in the name of the method because we also have shunt remote voltage control and reactive remote control.

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
```diff
@@ -59,4 +64,72 @@ public boolean isVoltageControlLocal() {
...
+        for (int i = 0; i < controllerBuses.size(); i++) {
+            LfBus controllerBus = controllerBuses.get(i);
+            if (!controllerBus.isVoltageControllerEnabled()) {
+                reactiveKeys[i] = 0d;
+            }
+        }
+
+        // update bus reactive keys
+        double reactiveKeysSum = Arrays.stream(reactiveKeys).sum();
+        for (int i = 0; i < controllerBuses.size(); i++) {
```
annetill, line 83: How do you know here that the sum is never equal to zero after updating the keys?

annetill (changes requested):
Maybe just some changes/comments, but the idea is good and looks right. I have run the LF on all my real test cases (IOP, RGCE, matpower cases) and all is okay. But during my test on a TYNDP file, I fall in an exception `Caused by: com.powsybl.commons.PowsyblException: Generators [_9064117a-2da4-46db-b1ee-f60810974a76, _ee520ab7-786a-499b-a3ae-d96fb5b12596] connected to bus '_55207717-1417-458c-9b5f-e67a3eedc46a_0' must control the voltage of the same bus`. Can we do something for that and just log a warn or do we assume the exception? I don't remember this issue with this file indeed.

geofjamg:
Do you also have it on main branch?

annetill:
> Do you also have it on main branch?

Yes. I have planned to work on stand by automaton of SVC tonight and tomorrow. And I think it will be better to have this PR merged before... even if it means a new rebase for the PR of shunt voltage control ! Do you think it is possible?

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
geofjamg: Looking at the code there is nothing specific to local controller even if it works with hybride remote/local controller. The only local mode is handled in the previous if.

src/main/java/com/powsybl/openloadflow/network/LfBus.java
geofjamg: Done

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
geofjamg: Yes it might happen, when all generators of the control, go PQ. 
=> I added a check

geofjamg:
> Maybe just some changes/comments, but the idea is good and looks right. I have run the LF on all my real test cases (IOP, RGCE, matpower cases) and all is okay. But during my test on a TYNDP file, I fall in an exception `Caused by: com.powsybl.commons.PowsyblException: Generators [_9064117a-2da4-46db-b1ee-f60810974a76, _ee520ab7-786a-499b-a3ae-d96fb5b12596] connected to bus '_55207717-1417-458c-9b5f-e67a3eedc46a_0' must control the voltage of the same bus`. Can we do something for that and just log a warn or do we assume the exception? I don't remember this issue with this file indeed.

This is manageable but we have probably to rethink our controller modeling. A controller bus is supposed in our design, to be part of only one voltage control (only one voltageController attribute in `AbstractLfBus` for instance). So there is quite a lot of rework to support it. Let's create an issue and do it in another PR?

What I do not understand is how we didn't had this case before. Is it something new in our test cases? Or do we just not skip it before?

geofjamg:
> > Do you also have it on main branch?
> 
> Yes. I have planned to work on stand by automaton of SVC tonight and tomorrow. And I think it will be better to have this PR merged before... even if it means a new rebase for the PR of shunt voltage control ! Do you think it is possible?

I'm not 100% confident with this PR. Is it really linked to SVC devs? Maybe you could just start from this branch to easily rebase when merged.

geofjamg:
> > > Do you also have it on main branch?
> > 
> > 
> > Yes. I have planned to work on stand by automaton of SVC tonight and tomorrow. And I think it will be better to have this PR merged before... even if it means a new rebase for the PR of shunt voltage control ! Do you think it is possible?
> 
> I'm not 100% confident with this PR. Is it really linked to SVC devs? Maybe you could just start from this branch to easily rebase when merged.

Looks good in fact.

annetill (approved):
It looks good indeed. Do you want me to run all the tests again?

geofjamg:
> It looks good indeed. Do you want me to run all the tests again? 

If you already did it, no.

PR #410, opened 2022-01-01
Shunt compensator refactoring

geofjamg:

No

Refactoring

We create one equation per IIDM shunt compensator.

We now create one equation for all IIDM shunt compensators connected to same bus. 

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfShunt.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfShunt.java
@@ -19,5 +19,7 @@ public interface LfShunt extends LfElement {
 
     Evaluable getQ();
 
+    void setBus(LfBus bus);
+
     void updateState();
 }
--- a/src/main/java/com/powsybl/openloadflow/network/LfShunt.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfShunt.java
@@ -6,18 +6,12 @@
  */
 package com.powsybl.openloadflow.network;
 
-import com.powsybl.openloadflow.util.Evaluable;
-
 /**
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 public interface LfShunt extends LfElement {
 
     double getB();
-
-    void setQ(Evaluable q);
-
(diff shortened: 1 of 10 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
```diff
@@ -60,6 +67,13 @@ public Evaluable getQ() {
 
     @Override
     public void updateState() {
-        shuntCompensator.getTerminal().setQ(q.eval() * PerUnit.SB);
+        double qCalc = q.eval();
+        double bSum = shuntCompensators.stream()
+                .mapToDouble(ShuntCompensator::getB)
+                .sum();
```
annetill, line 73: Be careful here because even if we have one or several shunts connected to a bus, the sum could be equal to zero.

src/main/java/com/powsybl/openloadflow/network/LfShunt.java
```diff
@@ -19,5 +19,7 @@
```
annetill, line 19: If we have access to the bus voltage in the `LfShunt`, we don't need `q` anymore. @geofjamg what do you think ?

src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
```diff
@@ -8,29 +8,33 @@
...
-    private Evaluable q = NAN;
+    protected LfBus bus;
 
-    public LfShuntImpl(ShuntCompensator shuntCompensator, LfNetwork network) {
+    public LfShuntImpl(List<ShuntCompensator> shuntCompensators, LfNetwork network) {
         super(network);
-        this.shuntCompensator = Objects.requireNonNull(shuntCompensator);
-        double nominalV = shuntCompensator.getTerminal().getVoltageLevel().getNominalV();
+        if (shuntCompensators.isEmpty()) {
+            throw new IllegalArgumentException("Empty shunt compensator list");
```
annetill, line 29: This could not happened indeed, I cannot unit test this line.

PR #415, opened 2022-01-08
Network should have at least one PV bus

geofjamg:

No

Bug fix

The check that is supposed to discard networks when there is not at least one bus controlling voltage is buggy. It is done before discarding generators from voltage control in case of narrow reactive range.
I have seen this bug, running load flow on CGMES RealGrid (Divergence using classical flat start, because of no PV bus).

2 fixes:
 - Check is done after network loading, so that we have the definitive controlled buses.
 - When reactive limits are disabled in `LoadFlowParameters`, we don't check anything on reactive limits.

**Other information**:


geofjamg:
@annetill I suspect, CGMES import has a bug, because with RealGrid, we only have minQ/maxQ generators limits with [0, 0] which means in IIDM that we have a valid "zero only" reactive range and load flow fails on it.

annetill (approved):

PR #419, opened 2022-01-11
Fails on a security analysis with phase control

annetill:

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -250,12 +250,9 @@ public void updateState(boolean phaseShifterRegulationOn, boolean isTransformerV
         branch.getTerminal2().setP(p2.eval() * PerUnit.SB);
         branch.getTerminal2().setQ(q2.eval() * PerUnit.SB);
 
-        if (phaseShifterRegulationOn && isPhaseController()  && discretePhaseControl.getMode() != DiscretePhaseControl.Mode.CONTROLLER) {
+        if (phaseShifterRegulationOn && isPhaseController()) {
```
geofjamg, line 253: I removed last part of the check, but I am not sure it is a good idea.

geofjamg:
@annetill looks good now!

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -96,13 +96,17 @@ private static void createLocalVoltageControlEquation(LfBus bus, LfNetworkParame
...
-    private static void createReactivePowerControlBranchEquation(LfBranch branch, ReactivePowerControl.ControlledSide controlledSide,
-                                                                 EquationSystem<AcVariableType, AcEquationType> equationSystem, EquationTerm<AcVariableType, AcEquationType> q) {
-        branch.getReactivePowerControl().ifPresent(reactivePowerControl -> {
-            if (reactivePowerControl.getControlledSide() == controlledSide) {
-                equationSystem.createEquation(branch.getNum(), AcEquationType.BRANCH_TARGET_Q).addTerm(q);
-            }
-        });
+    private static void createReactivePowerControlBranchEquation(LfBranch branch, LfBus bus1, LfBus bus2, EquationSystem<AcVariableType, AcEquationType> equationSystem,
+                                                                 boolean deriveA1, boolean deriveR1) {
+        if (bus1 != null && bus2 != null) {
```
annetill, line 101: I understand your fix, but I wonder if our building or `ReactivePowerControl` is safe enough...

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
```diff
@@ -324,26 +328,33 @@ private static void createNonImpedantBranch(LfBranch branch, LfBus bus1, LfBus b
...
-                // this equation is now inactive
-                LfBranch controller = dpc.getController();
-                EquationTerm<AcVariableType, AcEquationType> a1 = equationSystem.getVariable(controller.getNum(), AcVariableType.BRANCH_ALPHA1)
-                        .createTerm();
-                branch.setA1(a1);
-                equationSystem.createEquation(controller.getNum(), AcEquationType.BRANCH_TARGET_ALPHA1)
-                        .addTerm(a1)
-                        .setActive(false);
-            });
+    private static void createBranchActivePowerTargetEquation(LfBranch branch, LfBus bus1, LfBus bus2, EquationSystem<AcVariableType, AcEquationType> equationSystem,
```
annetill, line 331: Why does it solves our problem indeed because we don't go into this method for an update. What was the issue ?

geofjamg: The bug was because of an `EquationTerm` reused 2 times in 2 different equations. This was not obvious in the `Equation`design and a check was missing (I added it, see `Equation.java`fix) so this not surprising that at 2 places in the code an `EquationTerm` was associated to 2 `Equation` (for phase shifter flow control and also generator reactive flow control). An `EquationTerm` has a reverse link to its `Equation` which is set when term is added. When the term was added to the second `Equation`, the first link was overwritten and so on lost. The `EquationSystem` indexing rely on this `EquationTerm` to `Equation` link and as the link was incorrect, the index was not updated correctly.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystem.java
geofjamg: Yes, we should add more unit test to test different network conditions. Here is we should at least add a test with an open branch and maybe also test the lost of controlled branch in the security analysis.

annetill: In this PR?

geofjamg: Non let's keep it as it is in this PR and create another one later.

geofjamg (approved):

annetill (commented):
Approved!

PR #422, opened 2022-01-14
Support of shunt compensator contingency in security analysis

annetill:

**Other information**:


src/main/java/com/powsybl/openloadflow/network/LfContingency.java
```diff
@@ -99,10 +107,18 @@ public double getActivePowerLoss() {
         }
         buses.forEach(b -> branches.addAll(b.getBranches()));
 
+        Set<Pair<LfShunt, Double>> shunts = new HashSet<>(1);
+        for (Pair<String, Double> shuntAndB : propagatedContingency.getShuntIdsToLose()) {
+            LfShunt shunt = network.getShuntById(shuntAndB.getKey());
+            if (shunt != null) {
+                shunts.add(Pair.of(shunt, shuntAndB.getValue()));
```
geofjamg, line 114: The B value seems to be the global one (bus level) and not the one the targeting shunt?

geofjamg: In this case I think there is something tricky to implement in the LfShuntImpl::updateState to find the right section taking into account triggered IIDM shunt

annetill: Indeed, in this PR, it is not obvious but I don't support shunt with voltage control on, so it means that I will always have fixed shunt and a global B. That is why it is working... I am going to add a unit test.

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
```diff
@@ -79,7 +87,7 @@ public double getActivePowerLoss() {
         }
 
         // check if contingency split this network into multiple components
-        if (branches.isEmpty()) {
+        if (branches.isEmpty() && propagatedContingency.getShuntIdsToLose().isEmpty()) {
```
annetill, line 90: @geofjamg here maybe my check is not so clean... it was designed for branch contingencies only. Could you maybe improve it ?

src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
```diff
@@ -86,14 +96,19 @@ public PropagatedContingency(Contingency contingency, int index, Set<String> bra
...
+            }
         }
     }
 
     public static List<PropagatedContingency> createListForSensitivityAnalysis(Network network, List<Contingency> contingencies) {
         List<PropagatedContingency> propagatedContingencies = new ArrayList<>();
         for (int index = 0; index < contingencies.size(); index++) {
             Contingency contingency = contingencies.get(index);
-            PropagatedContingency propagatedContingency = PropagatedContingency.create(network, contingency, index);
+            PropagatedContingency propagatedContingency = PropagatedContingency.create(network, contingency, index, null);
```
annetill, line 111: @geofjamg do you agree with null here?

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
geofjamg: Looks good, but probably an issue with other injection contingency

geofjamg (approved):

Issue #241, opened 2021-03-12
Support more contingency types in security analysis


annetill:
* **Do you want to request a *feature* or report a *bug*?**

* **What is the current behavior?**

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

Increase the support of contingencies. We support Branch, Dangling Line, Line, Two Windings Transformer. We need to support Injection contingencies like Generator, Load, HVDC Line (treated as 2 loads for the moment) and maybe SVC. A bus bar section contingency will be a mixt of branches and buses contingencies. A shunt contingency must changed the equation system by removing the shunt equation.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #431, opened 2022-01-20
Security analysis: support of load and generator contingencies

annetill:

Yes, #241 .

For the moment, we only support branch and shunt contingencies in security analysis. With this PR, we will support also load and generator contingencies. I think that I will update the AC sensitivity analysis after this PR. 

Nothing.

It works except for a load contingency in a distributed slack on conform load.  The extension load detail is not taken into account at this stage. The parameter that ensure power factor to be constant is not taken into account also.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/LfGenerator.java
```diff
@@ -21,6 +25,10 @@
 
     boolean hasVoltageControl();
```
annetill, line 26: I wonder if `hasVoltageControl` and `hasReactivePowerControl` (indeed it means remote control here) are still useful because it could be replaced by gettint the generator control type.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -633,13 +625,12 @@ private void applyInjectionContingencies(Network network, LfNetwork lfNetwork, P
...
 
-        for (String loadId : contingency.getLoadIdsToLose()) {
-            Load load = network.getLoad(loadId);
-            Bus bus = load.getTerminal().getBusView().getBus();
-            if (bus != null) {
-                LfBus lfBus = lfNetwork.getBusById(bus.getId());
-                lfBus.setLoadTargetP(lfBus.getLoadTargetP() - load.getP0() / PerUnit.SB);
-                lfBus.getLfLoads().setAbsVariableLoadTargetP(lfBus.getLfLoads().getAbsVariableLoadTargetP() - load.getP0());
+        for (Triple<String, Double, Double> loadInfo : contingency.getLoadIdsToLose()) {
+            if (loadInfo.getLeft() != null) {
```
annetill, line 629: We have an issue here, I am going to fix it in next commits.

src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
```diff
@@ -90,28 +91,39 @@ public PropagatedContingency(Contingency contingency, int index, Set<String> bra
                 branchIdsToOpen.add(terminal.getConnectable().getId());
             }
             if (terminal.getConnectable() instanceof Generator) {
```
annetill, line 93: @flo-dup I am not sure that the following lines are relevant and useful. Can you explain me why you introduce this loop on terminals to disconnect and if my code is relevant when we loose a load, a generator or a shunt ? Thanks a lot !

flo-dup: Well, I didn't introduce that loop, but these lines are probably quite relevant and useful as they fill up the lists of disconnected equipments with the equipments traversed during propagation. I think those lists are then useful in the analysis - at least they are used a lot. But I feel like I misunderstood your question?

About the code correctness, I don't really get what you want me to check? indeed, you didn't really change that loop

flo-dup: A few things after trying to understand this code though:
- I don't get why you store the targetP of the generator? it seems that it's not used in the analysis, you only use the id
- The Triple object you store for loads makes the code unclear, why didn't you introduce a specific object instead?
- Do we need to store parameters of disconnected shuntCompensators with this Pair list? whereas it looks like the LfShuntCompensator already holds that info
- and strangely `getShuntIdsToLoose` is not called in the SensitivityAnalysis, unlike all other lists, so there's maybe something missing? Not current PR's topic though.

annetill: It is just the unit tests never go into that lines except the first if statement. So... I was wondering if it was relevant as it was introduced by non graph experts!

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -591,24 +567,39 @@ private void applyInjectionContingencies(Network network, LfNetwork lfNetwork, P
...
-            Generator generator = network.getGenerator(generatorId);
-            processGenerator(lfNetwork, generator, generators);
+        for (Map.Entry<String, Double> e : contingency.getGeneratorIdsToLose().entrySet()) {
+            generators.add((LfGeneratorImpl) lfNetwork.getGeneratorById(e.getKey()));
         }
 
-        for (String loadId : contingency.getLoadIdsToLose()) {
-            Load load = network.getLoad(loadId);
-            Bus bus = load.getTerminal().getBusView().getBus();
-            if (bus != null) {
```
geofjamg, line 608: What was the reason for null checking here? If load is part of the calculation, it is connected to a bus?

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -0,0 +1,332 @@
...
+        this.index = index;
+        this.branchIdsToOpen = Objects.requireNonNull(branchIdsToOpen);
+        this.hvdcIdsToOpen = Objects.requireNonNull(hvdcIdsToOpen);
+        this.switchesToOpen = Objects.requireNonNull(switchesToOpen);
+        this.generatorIdsToLose = Objects.requireNonNull(generatorIdsToLose);
+        this.loadIdsToShift = Objects.requireNonNull(loadIdsToShift);
+        this.shuntIdsToShift = Objects.requireNonNull(shuntIdsToShift);
+
+        // WTF?????
+        for (Switch sw : switchesToOpen) {
```
geofjamg, line 88: We will do it later, we there is something to fix here...

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -0,0 +1,332 @@
...
+            buses = connectivity.getSmallComponents().stream().flatMap(Set::stream).collect(Collectors.toSet());
+        } else {
+            int slackBusComponent = connectivity.getComponentNumber(network.getSlackBus());
+            buses = network.getBuses().stream().filter(b -> connectivity.getComponentNumber(b) != slackBusComponent).collect(Collectors.toSet());
+        }
+        buses.forEach(b -> branches.addAll(b.getBranches()));
+
+        Map<LfShunt, Double> shunts = new HashMap<>(1);
+        for (var e : shuntIdsToShift.entrySet()) {
+            // FIXME does not work when multiple shunts connected to a bus because IDs won't be found !!!!
```
geofjamg, line 302: This was not the purpose of this PR, but shunt contingency during SA cannot work as we do not search with the right ID since we have aggregated shunts by buses with a deviated ID. To fix later?

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -0,0 +1,332 @@
...
+
+        // process elements of the contingency
+        for (ContingencyElement element : contingency.getElements()) {
+            switch (element.getType()) {
+                case BRANCH:
+                case LINE:
+                case TWO_WINDINGS_TRANSFORMER:
+                    // branch check is done inside tripping
+                    ContingencyTripping.createBranchTripping(network, element.getId())
+                        .traverse(switchesToOpen, terminalsToDisconnect);
```
geofjamg, line 156: I removed here the adding to branchIdsToOpen, because it seems to me redundant with the post propagation processing as terminalsToDisconnect contains to initial starting equipment. Seems ok to you?

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -0,0 +1,332 @@
...
+                    if (hvdcLine == null) {
+                        throw new PowsyblException("HVDC line '" + element.getId() + "' not found in the network");
+                    }
+                    hvdcIdsToOpen.add(element.getId());
+                    break;
+                case DANGLING_LINE:
+                    // dangling line check is done inside tripping
+                    ContingencyTripping.createDanglingLineTripping(network, element.getId())
+                        .traverse(switchesToOpen, terminalsToDisconnect);
+                    break;
```
geofjamg, line 169: Same

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -0,0 +1,332 @@
...
+
+    private final Set<String> branchIdsToOpen;
+
+    private final Set<Switch> switchesToOpen;
+
+    private final Set<String> hvdcIdsToOpen;
+
+    private final Map<String, Double> generatorIdsToLose;
+
+    private final Map<String, PowerShift> loadIdsToShift;
```
geofjamg, line 43: I used shift because is only part of an aggregate

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -0,0 +1,332 @@
...
+
+    private final Map<String, Double> shuntIdsToShift;
```
geofjamg, line 45: I used shift because is only part of an aggregate

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -0,0 +1,332 @@
...
+        for (ShuntCompensator shunt : shuntsToLose) {
+            double nominalV = shunt.getTerminal().getVoltageLevel().getNominalV();
+            shuntIdsToShift.put(shunt.getId(), shunt.getB() * nominalV * nominalV / PerUnit.SB);
+        }
+
+        return new PropagatedContingency(contingency, index, branchIdsToOpen, hvdcIdsToOpen, switchesToOpen,
+                                         generatorIdsToLose, loadIdsToShift, shuntIdsToShift);
+    }
+
+    private static boolean isCoupler(Switch s) {
```
geofjamg, line 253: Outch... to redesign later

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/util/PropagatedContingency.java
geofjamg: > * Do we need to store parameters of disconnected shuntCompensators with this Pair list? whereas it looks like the LfShuntCompensator already holds that info

Shunts are aggregated by buses in the LF network

geofjamg: > * and strangely `getShuntIdsToLoose` is not called in the SensitivityAnalysis, unlike all other lists, so there's maybe something missing? Not current PR's topic though.

Shunt contingencies don't work at ALL in security and sensitivity analysis

geofjamg: > * I don't get why you store the targetP of the generator? it seems that it's not used in the analysis, you only use the id

Right, removed

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -0,0 +1,337 @@
...
+        Set<LfBus> buses;
+        if (useSmallComponents) {
+            buses = connectivity.getSmallComponents().stream().flatMap(Set::stream).collect(Collectors.toSet());
+        } else {
+            int slackBusComponent = connectivity.getComponentNumber(network.getSlackBus());
+            buses = network.getBuses().stream().filter(b -> connectivity.getComponentNumber(b) != slackBusComponent).collect(Collectors.toSet());
+        }
+        buses.forEach(b -> branches.addAll(b.getBranches()));
+
+        for (String hvdcId : hvdcIdsToOpen) {
```
geofjamg, line 301: I added an exception to warm that HVDC line contingency is not yet implemented in SA and sense analysis

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
annetill: Yes, better.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
annetill: Yes, better.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
annetill: Indeed I really don't know because it was my question to Florian but it tells me that it is not from him.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
annetill: I don't understand the problem. For each IIDM shunt id (`shunt.getIds()` gives a list of IIDM shunt ids) we can get the aggregated shunt. So for me if we loose 2 shunts connected to the same bus, the same aggregated LfShunt will be shift twice. The issue is maybe to use a HashMap or not to check if the LfShunt is already in the map ?

geofjamg: Right, I removed my comment

PR #435, opened 2022-01-24
Equation system index optimization

geofjamg:

No

feature

We do not precisely detect from equation system listener, if the resulting matrix structure has changed (row or column dimension change, or non zero element change).

We added an equation system index listener which is responsible for maintaining a precise few of an equation system change on matrix structure.

=> 10% performance improvement on security analysis

**Other information**:


geofjamg:
@annetill it should be better now

annetill (approved):

PR #443, opened 2022-02-03
Fixes: batteries and slack distribution on generators 

annetill:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfBatteryImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfBatteryImpl.java
@@ -21,19 +21,19 @@ public final class LfBatteryImpl extends AbstractLfGenerator {
 
     private static final double DEFAULT_DROOP = 1; // why not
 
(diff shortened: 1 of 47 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfBatteryImpl.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public final class LfBatteryImpl extends AbstractLfGenerator {
+
+    private static final double DEFAULT_DROOP = 1; // why not
```
geofjamg, line 22: Is there a reason to not take the same value as generators?

geofjamg:
Coverage is a little bit too low because, there is no test with action power control extension on battery

geofjamg (approved):

PR #447, opened 2022-02-07
Remove EquationUtil

geofjamg:

No

feature

The `EquationUtil` allows to save and restore equations state change because of a contingency.

We don't use `EquationUtil` anymore but only relies on bus/branch disabling. It means that there is no more direct access to equation system from LF/AS/Sensi. We always interact with the `LfNetwork` which automatically  reflects changes to the equation system. 

**Other information**:


src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -112,6 +109,9 @@ private void calculatePostContingencySensitivityValues(List<LfSensitivityFactor<
                                                            LoadFlowParameters lfParameters, OpenLoadFlowParameters lfParametersExt,
                                                            String contingencyId, int contingencyIndex, SensitivityValueWriter valueWriter,
                                                            Reporter reporter, boolean hasTransformerBusTargetVoltage) {
+        for (LfBranch branch : lfContingency.getBranches()) {
```
geofjamg, line 112: @annetill we can safely replace by LfContingency.apply ?

annetill: It is on my TODO list, I think it should work but I need to add some tests and to perform real tests. My question is: in an AC sensitivity analysis, do you need to support transformers control such as voltage or active power control ? Do we need to support shunt voltage control ?

geofjamg: For a sensitivity analysis, everything regarding control does not make really sense.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemUpdater.java
```diff
@@ -84,22 +44,48 @@ public void onShuntVoltageControlChange(LfShunt controllerShunt, boolean newVolt
...
-                bus.getVoltageControl().ifPresent(this::updateVoltageControl);
+                equationSystem.getEquation(bus.getNum(), AcEquationType.BUS_TARGET_PHI)
+                        .ifPresent(eq -> eq.setActive(!bus.isDisabled()));
+                equationSystem.getEquation(bus.getNum(), AcEquationType.BUS_TARGET_P)
+                        .ifPresent(eq -> eq.setActive(!bus.isDisabled() && !bus.isSlack()));
+                // set voltage target equation inactive, various voltage control will set next to the correct value
+                equationSystem.getEquation(bus.getNum(), AcEquationType.BUS_TARGET_V)
+                        .orElseThrow()
+                        .setActive(false);
+                bus.getVoltageControl().ifPresent(voltageControl -> AcEquationSystem.updateGeneratorVoltageControl(voltageControl, equationSystem));
```
annetill, line 80: Do you miss `controllerBus.getReactivePowerControl().ifPresent` here?

geofjamg: Yes, fixed

annetill (approved):

PR #450, opened 2022-02-17
Fix transformer control outerloop `AFTER_GENERATOR_VOLTAGE_CONTROL`

annetill:
- Add target deadband management.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/ShuntVoltageControl.java
```diff
@@ -19,7 +19,7 @@
     private final List<LfShunt> controllers = new ArrayList<>();
 
     public ShuntVoltageControl(LfBus controlled, double targetValue) {
-        super(controlled, targetValue);
+        super(controlled, targetValue, 0.0);
```
AnkurArohi, line 22: Here we set by  the value of dead band as zero(is a fixed value, should ideally be used as a parameter or as a property that could be set by the user), is that correct? Furthermore, in TransformerVoltagecontrolouterloop , this value is used to check if voltage lies beyond this point, in order to set the voltageControl as true

AnkurArohi (changes requested):

src/main/java/com/powsybl/openloadflow/ac/TransformerVoltageControlOuterLoop.java
```diff
@@ -14,13 +14,18 @@
...
+import java.util.List;
+
 /**
  * @author Anne Tilloy <anne.tilloy at rte-france.com>
  */
 public class TransformerVoltageControlOuterLoop extends AbstractTransformerVoltageControlOuterLoop {
 
     private double maxControlledNominalVoltage = Double.MIN_VALUE;
 
+    private List<LfBus> disabledControllerBuses = new ArrayList<>();
```
geofjamg, line 27: Disabled but never restored?
Also, we have to think about state management in outer loops, we should keep them stateless if possible to be able later to parallelize the security analysis more easily.

geofjamg (changes requested):

annetill:
Nothing is really important in that PR, I prefer to close it and reopen subjects later such as shunts' deadband.

PR #455, opened 2022-02-22
Disable switch without removing variables

geofjamg:

No

feature

When we disable a switch, we also disable the 2 dummy variables in AC (dummy P and dummy Q).

When we disable a switch, we keep dummy P and Q and set it to zero. As a consequence, the Jacobian dimension is constant, allowing further performance improvements.

**Other information**:


src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -1234,4 +1235,18 @@ void testViolationsWeakenedOrEquivalent() {
 
         assertFalse(AbstractSecurityAnalysis.violationWeakenedOrEquivalent(violation1, violation4, violationsParameters));
     }
+
+    @Test
+    void testWithNonImpedendantLineConnectedToSlackBus() {
```
annetill, line 1240: Impedendant?

annetill (approved):

PR #460, opened 2022-02-23
Fix KLU issue with regulating phase shifter necessary for connectivity

geofjamg:

No

Bug fix

We have a KLU factorization issue when a regulating phase shifter is necessary for connectivity: i.e if we lost the phase shifter network would be broken into 2 components.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
@@ -41,13 +41,13 @@ public class PhaseControlOuterLoop implements OuterLoop {
     @Override
     public void initialize(LfNetwork network) {
         List<LfBranch> controllerBranches = new ArrayList<>(1);
-        List<LfBranch> disableBranches = new ArrayList<>(1);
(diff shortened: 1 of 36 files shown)


annetill (approved):

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
```diff
@@ -34,6 +38,47 @@ public String getType() {
...
+                var connectivity = network.getConnectivity();
+
+                // apply contingency (in case we are inside a security analysis)
+                for (LfBranch disabledBranch : disabledBranches) {
+                    connectivity.cut(disabledBranch.getBus1(), disabledBranch.getBus2());
+                }
+                int smallComponentsCountBeforePhaseShifterLoss = connectivity.getSmallComponents().size();
+
+                // then the phase shifter controlled branch
+                if (!disabledBranches.contains(controlledBranch)) {
```
geofjamg, line 66: I think it was not the correct fix, as when a phase shifter is disable it is already deactivated (see ac equation system), so we can just just skip the connectivity check.

PR #461, opened 2022-02-25
refactor loadflow provider from core

EtienneLt:

it implements the modifications of the loadflow provider made in https://github.com/powsybl/powsybl-core/pull/1987

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
@@ -274,7 +274,10 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
     }
 
     public static OpenLoadFlowParameters load() {
-        return new OpenLoadFlowProvider().loadSpecificParameters(PlatformConfig.defaultConfig()).orElseThrow(IllegalStateException::new);
+        return new OpenLoadFlowProvider().loadSpecificParameters(PlatformConfig.defaultConfig())
+                .filter(OpenLoadFlowParameters.class::isInstance)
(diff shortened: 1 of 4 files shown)


src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -300,85 +302,6 @@ public String toString() {
                 ')';
     }
 
-    @AutoService(LoadFlowParameters.ConfigLoader.class)
-    public static class OpenLoadFlowConfigLoader implements LoadFlowParameters.ConfigLoader<OpenLoadFlowParameters> {
```
flo-dup, line 304: Here the class is removed and the loading is moved within `OpenLoadFlowProvider::loadSpecificParameters`, I'm ok to put this class in the provider but why not keep the class? Simply replacing 
`LoadFlowParameters.ConfigLoader<OpenLoadFlowParameters>` by `ExtensionConfigLoader<LoadFlowParameters, OpenLoadFlowParameters>`

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -275,7 +274,10 @@ public OpenLoadFlowParameters setHvdcAcEmulation(boolean hvdcAcEmulation) {
     }
 
     public static OpenLoadFlowParameters load() {
-        return new OpenLoadFlowConfigLoader().load(PlatformConfig.defaultConfig());
```
flo-dup, line 278: And then this change could be reversed

flo-dup (changes requested):

flo-dup (changes requested):

PR #482, opened 2022-03-18
Remove transformer voltage controls when no PV buses at non controlled side

### LINKED ISSUE #481: Voltage controller tap changer leading to KLU issues
* **Do you want to request a *feature* or report a *bug*?**

A feature, because for me, the solution is to deactivate a control. 

* **What is the current behavior?**

In case of transformer voltage control on, when a tap changer is controlling voltage (in general, at its secondary), if at primary we don't have any close generator controlling voltage too, it leads to a KLU issue.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

For local control only:

We have to deactivate this kind of tap changer. The question is how to detect this kind of tap changers ? During the network loading in `WITH_GENERATOR_CONTROL` mode or just before the second outerloop in `AFTER_VOLTAGE_CONTROL` mode, we have to :
- Making a connectivity analysis with the remove of all tap changers in voltage control ;
- Take for each the component that contains the terminal that is not the regulating terminal. In that component we have to find a controlled bus (meaning a bus controlled by a generator). If not, we deactivate tap changer control.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

annetill:

Yes, fixes #481.


annetill:
AC sensitivities support is missing.

src/main/java/com/powsybl/openloadflow/ac/AbstractTransformerVoltageControlOuterLoop.java
```diff
@@ -48,4 +50,43 @@ public void cleanup(LfNetwork network) {
...
+            Set<LfBus> componentOnNotControlledSide = null;
+            if (controlledBus.equals(branch.getBus1())) {
+                componentOnNotControlledSide = connectivity.getConnectedComponent(branch.getBus2());
+            } else if (controlledBus.equals(branch.getBus2())) {
+                componentOnNotControlledSide = connectivity.getConnectedComponent(branch.getBus1());
+            } else {
+                // I don't know.
+            }
+            if (componentOnNotControlledSide != null) {
+                Optional<LfBus> generatorControlledBus = componentOnNotControlledSide.stream().filter(LfBus::isVoltageControlled).findAny();
```
sylvlecl, line 83: May be worth it to cache the result for this component, in order to not go over the same buses multiple times.

sylvlecl (approved):

src/main/java/com/powsybl/openloadflow/ac/AbstractTransformerVoltageControlOuterLoop.java
```diff
@@ -48,4 +50,43 @@ public void cleanup(LfNetwork network) {
...
+                if (!generatorControlledBus.isPresent()) {
+                    branch.setVoltageControlEnabled(false);
```
geofjamg, line 85: Do not forget to restore it in the outer loop cleanup!

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/ac/AbstractTransformerVoltageControlOuterLoop.java
annetill: The component depends on the contingency?

src/main/java/com/powsybl/openloadflow/ac/AbstractTransformerVoltageControlOuterLoop.java
annetill: It is done as the condition is only to have a voltage control, no?

src/main/java/com/powsybl/openloadflow/ac/AbstractTransformerVoltageControlOuterLoop.java
sylvlecl: Yes, I mean for this call of the function, not for successive calls:
if you have multiple transformers connected to the same component, we don't need to go over the buses of the component for each transformer.

sylvlecl: Actually there is a more sever issue than this very minor perf consideration:
we must not perform the connectivity cut and the component check in the same for loop!

We must :
1. cut all the transformers
2. check the component for each transformer

Otherwise, the check may find that first transformers are connected to a "voltage controlled" component, just because we have not yet disconnected other transformers.

src/test/java/com/powsybl/openloadflow/sensi/ac/AcSensitivityAnalysisTest.java
```diff
@@ -428,6 +428,22 @@ void testBusVoltagePerTargetVTwt() {
...
+                .setTargetDeadband(0)
+                .setRegulating(true)
+                .setTapPosition(0)
+                .setRegulationTerminal(t2wt.getTerminal1()) // control will be disabled.
+                .setTargetV(135.0);
+
+        SensitivityAnalysisResult result2 = sensiRunner.run(network, factors, Collections.emptyList(), Collections.emptyList(), sensiParameters);
+
+        // FIXME, should be zero.
+        assertEquals(4, result2.getValues().size());
```
annetill, line 442: @Hadrien-Godard maybe you can help me on that unit test: the sensitivity value are so high...

annetill:
For me: all is handled except in an AC sensitivity analysis, after a contingency, if a transformer becomes enable to control, the predefined results are not filled.

annetill:
For me: all is handled except in an AC sensitivity analysis, after a contingency, if a transformer becomes enable to control, the predefined results are not filled.

geofjamg:
We should close this PR as when #522 will be done, we will only keep new the way the simulate tap changer (which is the only safe and correct way)

sylvlecl:
> We should close this PR as when #522 will be done, we will only keep new the way the simulate tap changer (which is the only safe and correct way)

Won't you have KLU issues in sensitivities, in incremental mode, if you don't merge this one ?

Also, if the goal is to leave all modes, not only incremental, this one will be necessary. I am not sure it's a good idea to remove other modes, since I suspect the incremental mode will be very very slow ?

annetill:
> > We should close this PR as when #522 will be done, we will only keep new the way the simulate tap changer (which is the only safe and correct way)
> 
> Won't you have KLU issues in sensitivities, in incremental mode, if you don't merge this one ?
> 
> Also, if the goal is to leave all modes, not only incremental, this one will be necessary. I am not sure it's a good idea to remove other modes, since I suspect the incremental mode will be very very slow ?

I have tested the incremental mode on CGMES IOP files, and it is clearly a beta feature. I really need to spend more time on it, but I don't have this time for the moment. In that PR, @Hadrien-Godard and me are also working on voltage sensitivities. So we will continue.

geofjamg:
> > > We should close this PR as when #522 will be done, we will only keep new the way the simulate tap changer (which is the only safe and correct way)
> > 
> > 
> > Won't you have KLU issues in sensitivities, in incremental mode, if you don't merge this one ?
> > Also, if the goal is to leave all modes, not only incremental, this one will be necessary. I am not sure it's a good idea to remove other modes, since I suspect the incremental mode will be very very slow ?
> 
> I have tested the incremental mode on CGMES IOP files, and it is clearly a beta feature. I really need to spend more time on it, but I don't have this time for the moment. In that PR, @Hadrien-Godard and me are also working on voltage sensitivities. So we will continue.

Every unrelated changes on voltage sensi have to move in another PR

geofjamg:
> > We should close this PR as when #522 will be done, we will only keep new the way the simulate tap changer (which is the only safe and correct way)
> 
> Won't you have KLU issues in sensitivities, in incremental mode, if you don't merge this one ?
> 
> Also, if the goal is to leave all modes, not only incremental, this one will be necessary. I am not sure it's a good idea to remove other modes, since I suspect the incremental mode will be very very slow ?

Not significantly slower that with this mode.
Only one or two additional LU factorizations, in case we target directly the final tap.

sylvlecl:
> Not significantly slower than with this mode.
> Only one or two additional LU factorizations, in case we target directly the final tap.

Ok, from the text of the PR I thought it was only +/-1 increments, but I understand it's not the case.

Hadrien-Godard (approved):

sylvlecl:
The previous discussion made me think that instead of the topological analysis, we could rely on a numerical analysis for the identification of those transfos.
Those would be the ones for which voltage sensitivity to the ratio is zero (epsilon).

src/main/java/com/powsybl/openloadflow/ac/AbstractTransformerVoltageControlOuterLoop.java
```diff
@@ -47,4 +51,41 @@ protected OuterLoopStatus roundVoltageRatios(OuterLoopContext context) {
         }
         return status;
     }
+
+    public static void checkControl(LfNetwork network) {
+        var connectivity = network.getConnectivity();
```
flo-dup, line 56: I'm wondering if using the DecrementalConnectivity here is consistent, as the init is costly. 

If we launch a security analysis, that's not a question as the init will be reused.

But if we only launch a loadflow, that's not really appropriate, and the naive approach is 4 times faster.

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/ac/AbstractTransformerVoltageControlOuterLoop.java
flo-dup: If we keep it this way, the performance issue has been fixed by #537. It was because we get the main connected component many times. And EvenShiloach algorithm recomputes the main connected component each time it is asked (it was first implemented for use cases where only the small components were needed).

geofjamg: > I'm wondering if using the EvenShiloachGraphDecrementalConnectivity here is consistent, as the init is costly.
> 
> If we launch a security analysis, that's not a question as the init will be reused.
> 
> But if we only launch a loadflow, that's not really appropriate, and the naive approach is 4 times faster.

This piece of code is ran both for a simple LF and a SA. We probably should use naive implementation for the LF and EvenShiloach for the SA but in both cases it has to be fast so it's a good thing that you improved performance for EvenShiloach.

geofjamg (approved):

annetill:
Performances are great now!

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -621,13 +621,11 @@ public void fixTransformerVoltageControls() {
...
                 // I don't know.
             }
-            if (componentOnNotControlledSide != null) {
-                if (componentOnNotControlledSide.stream().noneMatch(LfBus::isVoltageControlled)) {
-                    branch.setVoltageControlEnabled(false);
-                    LOGGER.trace("Transformer {} voltage control has been disabled because no PV buses on not controlled side connected component",
-                            branch.getId());
-                    disabledTransformerCount++;
-                }
+            if (componentOnNotControlledSide != null && componentOnNotControlledSide.stream().noneMatch(LfBus::isVoltageControlled)) {
```
flo-dup, line 624: Is there any chance we're looping many times over a big set of LfBus? Then we should save the results for each component.
But it would mean that there is a big (main) connected component with many controller branches with the corresponding controlled buses outside that component. And no other controlled bus. Not very likely I suppose.

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -621,13 +621,11 @@ public void fixTransformerVoltageControls() {
             } else {
                 // I don't know.
```
flo-dup, line 622: Shouldn't we add a TODO?

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
sylvlecl: Yes, it's likely:
typically you will have a very large component for EHV (400-225 kV), on which will be connected most transformers that regulate on the lower level (90-63 kV in France).

flo-dup: I added a commit to save the result for each component, it's supposed to be a bit faster... or at least more robust. It seems that in practice it's not faster as in big components, the `noneMatch` encounters always quickly a controlled bus.

flo-dup (approved):

PR #486, opened 2022-03-25
New sensitivity function types

obrix:

No

Adapt code to new function types available in the core.

On side one of branches was supported implicitely for reference flow.

Both sides of branches are now supported for reference flow.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
@@ -230,9 +230,15 @@ public abstract class AbstractSensitivityAnalysis<V extends Enum<V> & Quantity, 
         public EquationTerm<V, E> getFunctionEquationTerm() {
             switch (functionType) {
                 case BRANCH_ACTIVE_POWER:
+                case BRANCH_ACTIVE_POWER_1:
                     return (EquationTerm<V, E>) ((LfBranch) functionElement).getP1();
+                case BRANCH_ACTIVE_POWER_2:
+                    return (EquationTerm<V, E>) ((LfBranch) functionElement).getP2();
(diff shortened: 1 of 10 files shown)


src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -230,9 +230,15 @@ public ContingencyContext getContingencyContext() {
         public EquationTerm<V, E> getFunctionEquationTerm() {
             switch (functionType) {
                 case BRANCH_ACTIVE_POWER:
```
annetill, line 232: Do we have to keep case `BRANCH_ACTIVE_POWER`, I though the compatibility was done at a upper level?

annetill, line 232: Do we have to keep case `BRANCH_ACTIVE_POWER`, I though the compatibility was done at a upper level?

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -230,9 +230,15 @@ public ContingencyContext getContingencyContext() {
...
+                case BRANCH_ACTIVE_POWER_1:
                     return (EquationTerm<V, E>) ((LfBranch) functionElement).getP1();
+                case BRANCH_ACTIVE_POWER_2:
+                    return (EquationTerm<V, E>) ((LfBranch) functionElement).getP2();
                 case BRANCH_CURRENT:
```
annetill, line 237: Same question for the current.

annetill, line 237: Same question for the current.

PR #487, opened 2022-03-27
Fix transformer voltage control with multiple components

geofjamg:

No

Bug fix

When transformer voltage control is activated and multiple connected components is activated, the max nominal voltage stored inside outer loop is shared between the different component calculation => consequence is inconsistency/crash for other component calculation. This is because outer loop are not supposed to be stateful as there is only one instance for all calculation. So no data should be stored as a attribute of outer loops. 

State is store in LfNetwork as a user object.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/TransformerVoltageControlOuterLoop.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/TransformerVoltageControlOuterLoop.java
@@ -19,17 +19,20 @@ import com.powsybl.openloadflow.network.TransformerVoltageControl;
  */
 public class TransformerVoltageControlOuterLoop extends AbstractTransformerVoltageControlOuterLoop {
 
-    private double maxControlledNominalVoltage = Double.MIN_VALUE;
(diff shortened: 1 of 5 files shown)


src/main/java/com/powsybl/openloadflow/ac/TransformerVoltageControlOuterLoop.java
```diff
@@ -19,17 +19,20 @@
  */
 public class TransformerVoltageControlOuterLoop extends AbstractTransformerVoltageControlOuterLoop {
 
-    private double maxControlledNominalVoltage = Double.MIN_VALUE;
+    public static final String MAX_CONTROLLED_NOMINAL_VOLTAGE = "maxControlledNominalVoltage";
 
     @Override
     public void initialize(LfNetwork network) {
         // All transformer voltage control are disabled for the first equation system resolution.
+        double[] maxControlledNominalVoltage = new double[1];
```
annetill, line 27: Why to you need an array here and not a simple double?

geofjamg: To acces it from the lambda

annetill (approved):

PR #493, opened 2022-04-05
Contingency refactoring

annetill:

No, mainly a refactoring.

A refactoring for DC sensitvity analysis and the support of HVDC contingency for security analysis. Be careful that HVDC in AC emulation are not support at this stage.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -108,7 +109,7 @@ private static PowerShift getLoadPowerShift(Load load, boolean slackDistribution
         List<PropagatedContingency> propagatedContingencies = new ArrayList<>();
         for (int index = 0; index < contingencies.size(); index++) {
             Contingency contingency = contingencies.get(index);
-            PropagatedContingency propagatedContingency = PropagatedContingency.create(network, contingency, index, false, false, slackDistributionOnConformLoad);
+            PropagatedContingency propagatedContingency = PropagatedContingency.create(network, contingency, index, false, false, slackDistributionOnConformLoad, false);
```
annetill, line 112: For me: for an HVDC line in AC emulation, in AC sensitivity analysis and in AC security analysis, we have to:
- Handle it;
- Throw an exception (it is what we do for AC security analysis);
- Not do what is done for AC sensitivity analysis: saying nothing and try to handle it as set point.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -160,7 +163,25 @@ private static PropagatedContingency create(Network network, Contingency conting
                     if (hvdcLine == null) {
                         throw new PowsyblException("HVDC line '" + element.getId() + "' not found in the network");
                     }
+                    HvdcAngleDroopActivePowerControl control = hvdcLine.getExtension(HvdcAngleDroopActivePowerControl.class);
+                    if (control != null && control.isEnabled() && hvdcAcEmulation) {
+                        throw new PowsyblException("Contingency on HVDC line '" + element.getId() + "' operated in AC emulation: not supported yet.");
+                    }
                     hvdcIdsToOpen.add(element.getId());
```
annetill, line 170: For me:
- Use `hvdcIdsToOpen` only for HVDC in AC emulation, through a disabling of the line.
- Use converter station changes only for HVDC in active power setpoint.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -646,50 +563,41 @@ public DenseMatrix calculateStates(JacobianMatrix<DcVariableType, DcEquationType
...
+        if (contingency.getGeneratorIdsToLose().isEmpty() && contingency.getLoadIdsToShift().isEmpty()) {
             calculateSensitivityValues(factors, factorStates, contingenciesStates, flowStates, contingencyElements,
                     contingency, valueWriter);
         } else {
-            // if we have a contingency including the loss of a DC line or a generator
-            // if we have a contingency including the loss of a generator
             // if we have a contingency including the loss of a DC line or a generator or a load
-
-            List<BusState> busStates = new ArrayList<>();
+            // save base state for later restoration after each contingency
```
annetill, line 576: @geofjamg big change here: I save all the network state and not the buses affected by the contingency. But if you want, I can think about it and try.

Hadrien-Godard (approved):

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -164,7 +167,26 @@ private static PropagatedContingency create(Network network, Contingency conting
                     if (hvdcLine == null) {
                         throw new PowsyblException("HVDC line '" + element.getId() + "' not found in the network");
                     }
-                    hvdcIdsToOpen.add(element.getId());
+                    HvdcAngleDroopActivePowerControl control = hvdcLine.getExtension(HvdcAngleDroopActivePowerControl.class);
+                    if (control != null && control.isEnabled() && hvdcAcEmulation) {
+                        // Not depending on AC or DC computation.
+                        throw new PowsyblException("Contingency on HVDC line '" + element.getId() + "' with AC emulation control: not supported yet.");
```
annetill, line 173: @Hadrien-Godard and @geofjamg I am about to support this feature indeed for AC sensitivity analysis and for AC security analysis, but in another PR...

phiedw:
This seems to have created an issue in AC loadflows for sensitivity factors that have as function a network element that is in the contingency of the factor.

Hadrien-Godard:
> This seems to have created an issue in AC loadflows for sensitivity factors that have as function a network element that is in the contingency of the factor. (I get the following exception in a test for Farao: com.powsybl.commons.PowsyblException: Found an inactive equation for a factor that has no predefined result)

It is strange. A NaN pre defined result should be provided in this case (sensitivity function element which is in the contingency). Is the network element a line in your example ?

phiedw:
> It is indeed a line.
> 
> > > This seems to have created an issue in AC loadflows for sensitivity factors that have as function a network element that is in the contingency of the factor. (I get the following exception in a test for Farao: com.powsybl.commons.PowsyblException: Found an inactive equation for a factor that has no predefined result)
> > 
> > 
> > It is strange. A NaN pre defined result should be provided in this case (sensitivity function element which is in the contingency). Is the network element a line in your example ?

Your last commit fixed the issue! :smiley:

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
```diff
@@ -32,20 +29,25 @@ public class LfContingency {
...
 
+    private final Set<LfHvdc> disabledHvdcs;
+
     private final Map<LfShunt, Double> shuntsShift;
 
     private final Map<LfBus, PowerShift> busesLoadShift;
 
     private final Set<LfGenerator> generators;
 
+    private final Set<LfGenerator> participatingGeneratorsToBeRemoved = new HashSet<>();
```
geofjamg, line 40: lostParticipatingGenerators?

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -136,6 +129,12 @@ private void calculatePostContingencySensitivityValues(List<LfSensitivityFactor<
             lfNetwork.fixTransformerVoltageControls();
         }
 
+        if (hasMultiVariables && (!lfContingency.getBusesLoadShift().isEmpty() || !lfContingency.getGenerators().isEmpty())) {
+            // FIXME. It does not work with a contingency that breaks connectivity and loose an isolate injection.
+            Set<LfBus> affectedBuses = lfContingency.getLoadAndGeneratorBuses();
```
geofjamg, line 134: affected does not means what it is expected here. impacted?

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
```diff
@@ -118,6 +123,7 @@ public void apply(LoadFlowParameters parameters) {
             generator.setParticipating(false);
             if (generator.getGeneratorControlType() != LfGenerator.GeneratorControlType.OFF) {
                 generator.setGeneratorControlType(LfGenerator.GeneratorControlType.OFF);
+                participatingGeneratorsToBeRemoved.add(generator);
```
geofjamg, line 126: apply method is not supposed to change the state of LfContingency but LfNetwork one. If apply is called 2 times, participatingGeneratorsToBeRemoved will be filled 2 times. Is it really something that has to be stored in LfContingency? Maybe it could be just created on the fly by filtering generators of the LfContingency?

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -745,25 +742,55 @@ private static Injection<?> getInjection(Network network, String injectionId) {
         if (injection == null) {
             injection = network.getVscConverterStation(injectionId);
         }
-
-        if (injection == null) {
```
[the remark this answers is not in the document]
annetill: I think it is needed for your last PR that introduces the configured bus sensitivity computation, no?

geofjamg (approved):

PR #494, opened 2022-04-05
Get original ID

geofjamg:

No

feature

There is no strict equivalence between IIDM elements and Lf Network ones. Some elements like loads and shunts are aggregated and some other one a split (3 windings transformers and dangling lines).

We can get list of original IDs for each of the element and for branches a type to get original part of the IIDM element.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfBus.java
@@ -6,7 +6,6 @@
  */
 package com.powsybl.openloadflow.network;
 
-import com.powsybl.openloadflow.network.impl.LfLoads;
 import com.powsybl.openloadflow.util.Evaluable;
 import com.powsybl.security.results.BusResults;
 
--- a/src/main/java/com/powsybl/openloadflow/network/LfBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfBus.java
@@ -117,7 +117,7 @@ public interface LfBus extends LfElement {
 
     Optional<LfShunt> getControllerShunt();
 
-    LfLoads getLfLoads();
+    LfLoads getLoads();
 
     List<LfBranch> getBranches();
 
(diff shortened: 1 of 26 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
```diff
@@ -140,6 +140,11 @@ public String getId() {
         return shuntCompensators.stream().map(ShuntCompensator::getId).collect(Collectors.toList());
```
annetill, line 140: Can we delete this `getIds` method then?

annetill, line 140: Can we delete this `getIds` method then?

annetill:
And what about loads?

annetill:
And what about loads?

src/main/java/com/powsybl/openloadflow/network/impl/LfShuntImpl.java
geofjamg: Done

geofjamg:
> And what about loads?

Added

geofjamg:
For myself: add unit tests

src/main/java/com/powsybl/openloadflow/network/LfBus.java
```diff
@@ -118,7 +117,7 @@ default double getHighVoltageLimit() {
 
     Optional<LfShunt> getControllerShunt();
 
-    LfLoads getLfLoads();
+    LfLoads getLoads();
```
annetill, line 120: You really prefer that name? It is a bit confusing with the `Load` of IIDM network, no?

geofjamg: For any other element accessors we did not use lf

annetill (approved):

PR #496, opened 2022-04-05
DC sensitivity analysis: fix if contingency element is in GLSK

annetill:

No. In case of a contingency that involves an element of a GLSK, the computation of sensitivity value and reference flows is wrong. The GLSK has to be updated.

The GLSK is updated and the bus that contains the element is removed. But it leads to an approximation (or an error indeed) in case of two elements connected to the same bus and both part of the GLSK. This is illustrated in a unit test that fails (`assertNotEquals)`.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
@@ -697,8 +697,6 @@ public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis<DcVariabl
                 } else { // slack distribution on loads
                     newParticipatingElements = getParticipatingElements(lfNetwork.getBuses(), lfParameters.getBalanceType(), lfParametersExt);
                 }
-            } else {
-                newParticipatingElements = participatingElements;
             }
             if (participatingElementsChanged || rhsChanged) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -667,6 +667,24 @@ public void calculateContingencySensitivityValues(PropagatedContingency continge
...
+                        affectedBuses.add(bus);
+                    }
+                }
+                for (String id : contingency.getGeneratorIdsToLose()) {
+                    LfBus bus = lfNetwork.getGeneratorById(id).getBus();
+                    if (bus != null) {
+                        affectedBuses.add(bus);
+                    }
+                }
+                rhsChanged = rescaleGlsk(factorGroups, affectedBuses);
```
geofjamg, line 686: Are you sure, in case of multiple contingencies that we don't need to reset the weights as for a particular contingency part of the GLSK we update it but for next contingency we should restart from initial weights?

annetill: Yes, but it is what is done, no? In the factor, we keep them.

geofjamg (approved):

PR #499, opened 2022-04-06
Fix Even Shiloach algorithm

phiedw:

No

Bug fix

The Even Shiloach algorithm is run even if we do not disconnect all the edges between v1 and v2. This seems to cause the algorithm to sometimes never end.

The algorithm is now run only if we disconnect all edges between two vertices.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/graph/EvenShiloachGraphDecrementalConnectivity.java
+++ b/src/main/java/com/powsybl/openloadflow/graph/EvenShiloachGraphDecrementalConnectivity.java
@@ -180,19 +180,21 @@ public class EvenShiloachGraphDecrementalConnectivity<V> implements GraphDecreme
             V vertex2 = cutEdge.getRight();
             graph.removeEdge(vertex1, vertex2);
 
-            GraphProcess processA = new GraphProcessA(vertex1, vertex2);
-            GraphProcessB processB = new GraphProcessB(vertex1, vertex2);
-            while (!processA.isHalted() && !processB.isHalted()) {
-                processA.next();
-                if (!processA.isHalted()) {
-                    processB.next();
+            if (graph.getAllEdges(vertex1, vertex2).isEmpty()) {
+                GraphProcess processA = new GraphProcessA(vertex1, vertex2);
+                GraphProcessB processB = new GraphProcessB(vertex1, vertex2);
+                while (!processA.isHalted() && !processB.isHalted()) {
+                    processA.next();
+                    if (!processA.isHalted()) {
+                        processB.next();
+                    }
                 }
-            }
 
-            if (processA.isHalted()) {
-                processB.undoChanges();
-            } else { // processB halted
-                allSavedChangedLevels.add(processB.savedChangedLevels);
+                if (processA.isHalted()) {
+                    processB.undoChanges();
+                } else { // processB halted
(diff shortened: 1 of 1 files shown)


geofjamg:
> Are we able to reproduce this issue in a small unit test? What is the code of OLF that you are using and that makes you meet this issue? An AC/DC sensitivity analysis?
> @geofjamg and @flo-dup I was just wondering: even if you have some doubts about this fix, it will increase the performances, so I am for it. No?

We need a unit test to show the issue

flo-dup:
Three reasons for the observed bug:
- The non-deterministic behaviour is due to the `PropagatedContingency.branchIdsToOpen` being a `HashSet` and the `propagatedContingencyMap` using a `HashSet`.
- The `connectivity.reset()` is done too late in the sensitivity analysis, leading to the edges being cut twice for the contingencies breaking the connectivity (and the reset done twice in a row).
- The EvenShiloach algorithm changes the graph if a edge (with no parallel edges) is cut twice, as all the cut edges are added in the reset. And then the save changed levels are not consistent with the graph itself, leading to the unwanted behaviour.
 
These three reasons can be easily fixed - the last one should probably raise an exception I'd say? 

And we should rename this PR as this is not really a fix of the Even Shiloach but rather a misuse! :stuck_out_tongue:

annetill:
Good news ! Do you think you can do the fixes today?

geofjamg:
> * The EvenShiloach algorithm changes the graph if a edge (with no parallel edges) is cut twice, as all the cut edges are added in the reset. And then the save changed levels are not consistent with the graph itself, leading to the unwanted behaviour.

We already seen this bug, a few monthes ago but just workaround by no calling it twice (for phase shifter connectivity). I think we should either throw an exception or support multiple cut to void re-having the same issue.

geofjamg:
> Three reasons for the observed bug:
> 
> * The non-deterministic behaviour is due to the `PropagatedContingency.branchIdsToOpen` being a `HashSet` and the `propagatedContingencyMap` using a `HashSet`.
> * The `connectivity.reset()` is done too late in the sensitivity analysis, leading to the edges being cut twice for the contingencies breaking the connectivity (and the reset done twice in a row).
> * The EvenShiloach algorithm changes the graph if a edge (with no parallel edges) is cut twice, as all the cut edges are added in the reset. And then the save changed levels are not consistent with the graph itself, leading to the unwanted behaviour.
> 
> These three reasons can be easily fixed - the last one should probably raise an exception I'd say?
> 
> And we should rename this PR as this is not really a fix of the Even Shiloach but rather a misuse! 😛

To fix in another PR.

geofjamg:
> * The `connectivity.reset()` is done too late in the sensitivity analysis, leading to the edges being cut twice for the contingencies breaking the connectivity (and the reset done twice in a row).

@annetill has it been fixed in your PR?

geofjamg:
> And we should rename this PR as this is not really a fix of the Even Shiloach but rather a misuse! 😛

We have to fix all connectivity implementations to ensure that cutting edge multiple is like cutting edge just one time. I agree that it is a bad usage of the algorithm but this is the second time has such a mistake and probably not the last one and probably best option is to be robust to this use case.

annetill:
Partially resolved by #516 and #510

flo-dup:
Replaced by #517 as the fix is changing the GraphDecrementalConnectivity API

PR #501, opened 2022-04-08
Switch contingency support

geofjamg:

**Other information**:


annetill (commented):
What about the impact on sensitivity analysis? As we are in bus view for sensitivity analysis (right?), we will fall in the exception?

geofjamg:
> What about the impact on sensitivity analysis? As we are in bus view for sensitivity analysis (right?), we will fall in the exception?

I have to test it. For DC sensitivity analysis it is same issue as for contingency propagation: so it is an ongoing topic.
For AC sensitivity analysis it should work without any problem as we rebuild the jacobian matrix for each contingency => to test

geofjamg:
> > What about the impact on sensitivity analysis? As we are in bus view for sensitivity analysis (right?), we will fall in the exception?
> 
> I have to test it. For DC sensitivity analysis it is same issue as for contingency propagation: so it is an ongoing topic. For AC sensitivity analysis it should work without any problem as we rebuild the jacobian matrix for each contingency => to test

I just threw an exception in case of sensitivity analysis. As calculation is done on bus/view (breakers = false in DC and in AC), there is a LOT a work to make it working for sensitivity analysis and not sure somebody needs it. So let's see that later.

annetill (approved):
Yes, later for sensitivity analysis !

PR #506, opened 2022-04-11
Open branches sensitivity

flo-dup:

No

Bug fix

Open branches sensitivity is
- 0 if contingency is breaking the connectivity
- 0 if contingency is not breaking the connectivity and the branch is in the main component
- NaN if contingency is not breaking the connectivity and the branch is not in the main component

Open branches sensitivity is always 0.

No


annetill:
The issue is more complex indeed, we have to follow the same logic as DC sensitivity analysis.

annetill:
The issue is more complex indeed, we have to follow the same logic as DC sensitivity analysis.

flo-dup:
Why don't we follow the same logic then? In the meantime it's strange to split this setPredefinedResults in parts instead of centralizing it, that was also the goal of this PR

annetill:
> Why don't we follow the same logic then? In the meantime it's strange to split this setPredefinedResults in parts instead of centralizing it, that was also the goal of this PR

I think that the issue is more complex and has to be rethink more globally. As you said, I think that branch in contingency must be treated inside the predefined results. And we have to check what is done compared to DC sensitivity analysis. I have closed this PR but @Hadrien-Godard and me are going to reopen one, don't worry.

PR #514, opened 2022-04-13
Fix outer loops cleanup

geofjamg:

**Other information**:


src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -135,10 +138,14 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
             if (preContingencyComputationOk) {
                 detectViolations(network.getBranches().stream(), network.getBuses().stream(), preContingencyLimitViolations);
 
-                LOGGER.info("Save pre-contingency state");
+                // update voltage and active power targets to saved state
+                networkState.getBusStates().forEach(busState -> {
```
geofjamg, line 142: @annetill this is here that we have to decide what to keep from base case LF

geofjamg:
@annetill need to update AC sensitivity analysis too ?

src/main/java/com/powsybl/openloadflow/network/BusDcState.java
```diff
@@ -14,17 +14,21 @@
...
     public BusDcState(LfBus bus) {
         super(bus);
-        this.loadTargetP = bus.getLoadTargetP();
-        this.generatorsTargetP = bus.getGenerators().stream().collect(Collectors.toMap(LfGenerator::getId, LfGenerator::getTargetP));
-        this.participatingGenerators = bus.getGenerators().stream().collect(Collectors.toMap(LfGenerator::getId, LfGenerator::isParticipating));
-        this.absVariableLoadTargetP = bus.getLoads().getAbsVariableLoadTargetP();
+        updateTargetP();
+    }
+
+    public void updateTargetP() {
```
annetill, line 27: Why calling that method "update" ? It is just a save, no?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -120,6 +120,9 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
 
         // run pre-contingency simulation
         try (AcLoadFlowContext context = new AcLoadFlowContext(network, acParameters)) {
+            // save state for later restoration after base case and after each contingency
+            NetworkState networkState = NetworkState.save(network);
```
annetill, line 124: This is not the right place to save the state. Here, you are loosing the benefits of the first resolution : the slack distribution, the switching of PV buses to PQ buses, the final tap position of transformers and the final section count of shunts. This has to be done after the `run`, for performance mainly.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -135,10 +138,14 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
             if (preContingencyComputationOk) {
                 detectViolations(network.getBranches().stream(), network.getBuses().stream(), preContingencyLimitViolations);
 
-                LOGGER.info("Save pre-contingency state");
+                // update voltage and active power targets to saved state
+                networkState.getBusStates().forEach(busState -> {
+                    busState.updateVoltage();
+                    busState.updateTargetP();
```
annetill, line 144: Yes I get you! Here you re-save specific fields, that is why you call that `update` ! So for me, it is a good design but:
- If some bus where PV and have switch PQ, you have to re-start from the PQ state. All is design for this bus to be able to switch back PV after a contingency. The fields that you need to store are numerous.
- If you distribute the slack on loads keeping power factor constant, you also have to save the `loadTargetQ`.
- After the pre-contingency computation, all the discrete regulation are off. I think that we have to restore them, in order to be able to re-run them after a contingency. It could be optional.

geofjamg:
I tried to:
 - never change an initial control enabled status (so we take the one before running the LF) => I think we had a lot of mistake where we were supposing before starting the LF, all controls are enabled.
 - restore when necessary control status (because of rounding to temporary change like in transformer voltage control for buses)

It seems that I broke a few unit test, 2 possibilities:
 - what I coded is wrong
 - Some unit tests were wrong before

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
```diff
@@ -97,14 +96,16 @@ public OuterLoopStatus check(OuterLoopContext context, Reporter reporter) {
     private OuterLoopStatus firstIteration(OuterLoopContext context) {
 
         List<DiscretePhaseControl> phaseControlsOn = context.getNetwork().getBranches().stream()
-            .map(branch -> branch.getDiscretePhaseControl().filter(dpc -> branch.isPhaseControlled()))
-            .filter(Optional::isPresent)
-            .map(Optional::get)
-            .filter(dpc -> dpc.getMode() == DiscretePhaseControl.Mode.CONTROLLER || dpc.getMode() == DiscretePhaseControl.Mode.LIMITER)
+            .filter(branch -> branch.isPhaseControlled()
+            /*        && branch.isPhaseControlEnabled() FIXME */
```
geofjamg, line 100: @annetill it seems to broke some more unit test if I check branch phase control is on before switching control!

AnkurArohi: What is meant by branch control is on before switching control

annetill: I am checked some tests and it seems that even if you save a property, the properties are null.

annetill:
I have introduced a fix that fix the load flows and the security analysis. Your mechanism seems to improve these two use cases, but it still leads to an issue for AC sensitivity analysis where after the first LF, we need a Jacobian Matrix without any control. But here we still have all.

geofjamg:
> I have introduced a fix that fix the load flows and the security analysis. Your mechanism seems to improve these two use cases, but it still leads to an issue for AC sensitivity analysis where after the first LF, we need a Jacobian Matrix without any control. But here we still have all.   

This is an incorrect fix as it overwrite initial value and situation is exactly the same than without this pr...

geofjamg:
> > I have introduced a fix that fix the load flows and the security analysis. Your mechanism seems to improve these two use cases, but it still leads to an issue for AC sensitivity analysis where after the first LF, we need a Jacobian Matrix without any control. But here we still have all.   
> 
> This is an incorrect fix as it overwrite initial value and situation is exactly the same than without this pr...

It does not explain why i have done in this pr does not work. I still suspect many tests to bé wrong

AnkurArohi:
> > > I have introduced a fix that fix the load flows and the security analysis. Your mechanism seems to improve these two use cases, but it still leads to an issue for AC sensitivity analysis where after the first LF, we need a Jacobian Matrix without any control. But here we still have all.
> > 
> > 
> > This is an incorrect fix as it overwrite initial value and situation is exactly the same than without this pr...
> 
> It does not explain why what i have done in this pr does not work. I still suspect many tests to be wrong

I think @geofjamg  is right. When the initial values are overwritten , the results should be a bit different

geofjamg:
@annetill new try with a different a design and unit test still fails... I almost sure that with this code voltage or phase control is correctly restored, are you sure your unit test is correct?

AnkurArohi:
Normally if you have a jacobian matrix you have all factors you want to know from the ac sensitivity analysis 
I agree with this

AnkurArohi:
@geofjamg we try to have a jacobian matrix without any control or local regulations. 
After that we need to post process the result with respect to regulation calculations.

Sens = dP/dP × local regulation factor

geofjamg:
So, finally (I wish...) I have something that looks good. The design is very different to the one of first commit. Now we never activate transformer voltage and phase, shunt control at network loading but instead we do it before le LF init outer loop initializer and then restoring it to false in cleanup. So it does no bother sensitivity calculation. I think it is a better design because all the enables status of control are juste temporary flags activate or deactivate between the different outer loops. 
@annetill what do you think?

src/main/java/com/powsybl/openloadflow/ac/AbstractTransformerVoltageControlOuterLoop.java
```diff
@@ -12,41 +12,46 @@
...
+            LOGGER.trace("Round voltage ratio of '{}': {} -> {}", controllerBranch.getId(), r1Value, roundedR1Value);
+
+            status = OuterLoopStatus.UNSTABLE;
         }
         return status;
     }
 
     @Override
     public void cleanup(OuterLoopContext context) {
-        for (LfBranch branch : context.getNetwork().getBranches()) {
```
annetill, line 48: Why do you need this clean up? It is already done through the rounded of ratios, no?

src/main/java/com/powsybl/openloadflow/ac/PhaseControlOuterLoop.java
```diff
@@ -165,10 +159,8 @@ boolean isSensitivityCurrentPerA1Positive(LfBranch controllerBranch, DiscretePha
 
     @Override
     public void cleanup(OuterLoopContext context) {
-        for (LfBranch branch : context.getNetwork().getBranches()) {
-            if (branch.isPhaseController()) {
-                branch.setPhaseControlEnabled(true);
-            }
+        for (LfBranch branch : getControllerBranches(context.getNetwork())) {
```
annetill, line 162: Here too: do we really need this clean up?

src/main/java/com/powsybl/openloadflow/ac/ShuntVoltageControlOuterLoop.java
```diff
@@ -27,32 +30,39 @@ public String getType() {
...
         }
         return status;
     }
 
     @Override
     public void cleanup(OuterLoopContext context) {
-        for (LfBus bus : context.getNetwork().getBuses()) {
-            bus.getShuntVoltageControl().ifPresent(b -> b.getControllers().forEach(controllerShunt -> controllerShunt.setVoltageControlEnabled(true)));
-        }
+        getControllerShunts(context.getNetwork()).forEach(controllerShunt -> controllerShunt.setVoltageControlEnabled(false));
```
annetill, line 66: Same question.

annetill (approved):

PR #522, opened 2022-04-22
[Transformer voltage control] Incremental transformer outerloop

annetill:

Not exactly, but the transformer voltage control is a complexe feature. We already have two outerloops, but in some cases, it leads to a divergence or a KLU issue. So, we have decided to try an incremental outerloop.

We add an incremental outerloop. For the moment, at the first resolution, the transformer voltage controls are disabled. The outerloop looks like:
- At each controlled bus, we check if voltage is higher or lower than the target value ;
- If yes, the compute the sensitivity of a rho increase on the controlled bus. Depending on the sign, we move the tap +1 or -1 ;
- We recompute a LF. 

NB:
- Shared controls are not handle ;
- Stopping criteria?

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/PiModel.java
+++ b/src/main/java/com/powsybl/openloadflow/network/PiModel.java
@@ -58,9 +58,7 @@ public interface PiModel {
 
     boolean updateTapPosition(Direction direction);
 
-    boolean increaseTapPosition();
-
-    boolean decreaseTapPosition();
+    boolean updateTapPositionR(Direction direction);
 
     boolean setMinZ(double minZ, boolean dc);
 
--- a/src/main/java/com/powsybl/openloadflow/network/PiModel.java
+++ b/src/main/java/com/powsybl/openloadflow/network/PiModel.java
@@ -76,7 +76,7 @@ public interface PiModel {
 
     boolean updateTapPositionA1(Direction direction);
 
-    Optional<Direction> updateTapPositionR1(double deltaR1, int maxTapIncrement, AllowedDirection allowedDirection);
+    Optional<Direction> updateTapPositionR1(double deltaR1, int maxTapShift, AllowedDirection allowedDirection);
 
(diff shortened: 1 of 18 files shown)


src/main/java/com/powsybl/openloadflow/network/PiModel.java
```diff
@@ -50,12 +50,17 @@ public interface PiModel {
...
     enum Direction {
+        NONE,
         INCREASE,
-        DECREASE
+        DECREASE,
+        INCREASE_THEN_DECREASE,
+        DECREASE_THEN_INCREASE
     }
 
     boolean updateTapPosition(Direction direction);
```
geofjamg, line 60: To rename updateTapPositionA1

src/main/java/com/powsybl/openloadflow/network/PiModel.java
```diff
@@ -50,12 +50,17 @@ public interface PiModel {
     void roundR1ToClosestTap();
 
     enum Direction {
+        NONE,
         INCREASE,
-        DECREASE
+        DECREASE,
+        INCREASE_THEN_DECREASE,
```
geofjamg, line 56: We should not have INCREASE_THEN_DECREASE and DECREASE_THEN_INCREASE in the enum because it is also the one used for phase shifter with meaningless values in this case. Also the memory of past direction should be stored in another way: for instance with a counter a direction change.

src/main/java/com/powsybl/openloadflow/network/PiModelArray.java
```diff
@@ -26,10 +26,13 @@ public class PiModelArray implements PiModel {
 
     private LfBranch branch;
 
+    private Direction r1Variation;
```
geofjamg, line 29: Ideally we should not have anything release to past state inside pi model array but just a counter in the outer loop context.

geofjamg (approved):

Hadrien-Godard (approved):

Issue #511, opened 2022-04-13
AC sensitivity analysis


annetill:
* **Do you want to request a *feature* or report a *bug*?**

Bugs in the mangement of predefined results.

* **What is the current behavior?**

We have a special treatment for factors that contains as function reference a branch that is in contingency. Sensitivity value and function reference are set to zero.  

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

We have to follow the same logic as DC sensitivity analysis.

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #523, opened 2022-04-22
Change AC sensitivities in case of contingency to match DC

Hadrien-Godard:
- Now in case of a connectivity loss, the sensitivity and the reference flow do not automatically value 0 on the disabled branches.contingencyContext
- Now the sensitivity is set to zero if the variable of the factor is in the contingency, when contingency is not breaking connectivity.

#511 

**Other information**:


src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -233,13 +233,13 @@ private void createBranchSensitivityValue(LfSensitivityFactor<DcVariableType, Dc
                     sensiValue += contingencyElement.getAlphaForSensitivityValue() * contingencySensitivity;
```
AnkurArohi, line 233: I do think that sens =0 is confusing here as there could be some cases where contingency has not occurred on one particular line but still there is sens =0 instead of that I would suggest using enum to make it clear that this is the contingency case which has affected this line

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -233,13 +233,13 @@ private void createBranchSensitivityValue(LfSensitivityFactor<DcVariableType, Dc
...
                 }
                 if (contingencyElement.getElement().getId().equals(functionBranchId)) {
-                    // the monitored branch is in contingency, sensitivity and post-contingency flow equals to zero in any case.
-                    flowValue = 0d;
+                    // the monitored branch is in contingency, sensitivity equals 0 and post-contingency flow equals NaN in any case.
+                    flowValue = Double.NaN;
```
AnkurArohi, line 237: Is the post contingency flow NaN 🤔 good 💡,  can we think of something better?
It is clear that after contingency this like is no more active in network
I would actually think of another Enum which would specify this and then even NaN would make sense.
Enum =POST_CONT_OUTOFSERV

annetill (approved):

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
annetill: As the sensitivity analysis through open-loadflow is used a lot by the community of pypowsybl users, `NaN` is much more adapted for them.

Issue #526, opened 2022-05-03
Reactive power outside of reactive limits


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Bug.

* **What is the current behavior?**

It seems that the bus-level reactive power, when distributed on generation units at the end of the computation, can exceed their reactive limits.

See the outcome of a loadflow run, when the bus-level limit is respected, but the limit on GEN-201#2 is exceeded (20 > 16).
```
             bus_id  max_q          q
id                                   
GEN-201    VL-228_0   10.0 -10.000000
GEN-201#0  VL-228_0   10.0 -10.000000
GEN-201#1  VL-228_0   30.0 -20.896113
GEN-201#2  VL-228_0   16.0 -20.896113
```

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Load and run a loadflow with the following file:
[reactive-limit-issue.xiidm.gz](https://github.com/powsybl/powsybl-open-loadflow/files/8610225/reactive-limit-issue.xiidm.gz)

Using pypowsybl, you can follow the following steps :
```python
network = pp.network.load('reactive-limit-issue.xiidm.gz')
pp.loadflow.run_ac(network)
gens = network.get_generators()
print(gens[gens.index.str.contains('GEN-201')][['bus_id', 'max_q', 'q']])
```

* **What is the expected behavior?**

When distributing on generation units, we should take care of keeping Q inside their respective limits.

AnkurArohi:
@sylvlecl Good point
Here the distribution should be a constraint distribution of Q respecting the limits.
It should not be a problem

Idea 💡  we can use 
Import numpy as np
If gen.q in np.arange(-maxq , maxq);
  Accept it 
Else 
  Continue redistribution 
Same for P

sylvlecl:
Actually, we found out there is already such a logic implemented:
https://github.com/powsybl/powsybl-open-loadflow/blob/9215b2c5b4d2e926fed8d278a0d56ab3d5219dcb/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java#L453

But there is a small mistake in the implementation, that @EtienneLt is fixing.

PR #529, opened 2022-05-04
fix dispatchQ method

EtienneLt:

#526

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
@@ -415,22 +415,26 @@ public abstract class AbstractLfBus extends AbstractElement implements LfBus {
         hvdcs.add(Objects.requireNonNull(hvdc));
     }
 
-    private static double dispatchQ(List<LfGenerator> generatorsThatControlVoltage, boolean reactiveLimits, double qToDispatch) {
+    protected static double dispatchQ(List<LfGenerator> generatorsThatControlVoltage, boolean reactiveLimits, double qToDispatch) {
         double residueQ = 0;
-        double calculatedQ = qToDispatch / generatorsThatControlVoltage.size();
+        if (generatorsThatControlVoltage.isEmpty()) {
+            throw new IllegalArgumentException("the generator list to dispatch Q can not be empty");
+        }
+        double qToBeDispatchedByGenerator = qToDispatch / generatorsThatControlVoltage.size();
         Iterator<LfGenerator> itG = generatorsThatControlVoltage.iterator();
         while (itG.hasNext()) {
             LfGenerator generator = itG.next();
-            if (reactiveLimits && calculatedQ < generator.getMinQ()) {
-                generator.setCalculatedQ(generator.getCalculatedQ() + generator.getMinQ());
-                residueQ += calculatedQ - generator.getMinQ();
+            double generatorAlreadyCalculatedQ = generator.getCalculatedQ();
+            if (reactiveLimits && qToBeDispatchedByGenerator + generatorAlreadyCalculatedQ < generator.getMinQ()) {
+                residueQ += qToBeDispatchedByGenerator + generatorAlreadyCalculatedQ - generator.getMinQ();
+                generator.setCalculatedQ(generator.getMinQ());
                 itG.remove();
-            } else if (reactiveLimits && calculatedQ > generator.getMaxQ()) {
-                generator.setCalculatedQ(generator.getCalculatedQ() + generator.getMaxQ());
-                residueQ += calculatedQ - generator.getMaxQ();
+            } else if (reactiveLimits && qToBeDispatchedByGenerator + generatorAlreadyCalculatedQ > generator.getMaxQ()) {
+                residueQ += qToBeDispatchedByGenerator + generatorAlreadyCalculatedQ - generator.getMaxQ();
+                generator.setCalculatedQ(generator.getMaxQ());
                 itG.remove();
             } else {
-                generator.setCalculatedQ(generator.getCalculatedQ() + calculatedQ);
+                generator.setCalculatedQ(generatorAlreadyCalculatedQ + qToBeDispatchedByGenerator);
             }
         }
         return residueQ;
@@ -458,7 +462,7 @@ public abstract class AbstractLfBus extends AbstractElement implements LfBus {
     @Override
     public void updateState(boolean reactiveLimits, boolean writeSlackBus, boolean distributedOnConformLoad, boolean loadPowerFactorConstant) {
         // update generator reactive power
-        updateGeneratorsState(voltageControlEnabled ? q.eval()  * PerUnit.SB + loadTargetQ : generationTargetQ, reactiveLimits);
+        updateGeneratorsState(voltageControlEnabled ? q.eval() * PerUnit.SB + loadTargetQ : generationTargetQ, reactiveLimits);
 
         // update load power
         lfLoads.updateState(getLoadTargetP() - getInitialLoadTargetP(), loadPowerFactorConstant);
--- a/src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowEurostagTutorialExample1Test.java
+++ b/src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowEurostagTutorialExample1Test.java
@@ -409,4 +409,28 @@ class AcLoadFlowEurostagTutorialExample1Test {
         assertActivePowerEquals(Double.NaN, gen.getTerminal());
         assertReactivePowerEquals(Double.NaN, gen.getTerminal());
     }
+
+    @Test
+    void testGeneratorReactiveLimits() {
+        Network network = EurostagTutorialExample1Factory.create();
+        network.getGenerator("GEN").newMinMaxReactiveLimits().setMinQ(0).setMaxQ(120).add();
+        network.getVoltageLevel("VLGEN").newGenerator().setId("GEN1")
+                .setBus("NGEN").setConnectableBus("NGEN")
+                .setMinP(-9999.99D).setMaxP(9999.99D)
+                .setVoltageRegulatorOn(true).setTargetV(24.5D)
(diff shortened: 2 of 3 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -415,19 +415,19 @@ public void addHvdc(LfHvdc hvdc) {
         hvdcs.add(Objects.requireNonNull(hvdc));
     }
 
-    private static double dispatchQ(List<LfGenerator> generatorsThatControlVoltage, boolean reactiveLimits, double qToDispatch) {
+    protected static double dispatchQ(List<LfGenerator> generatorsThatControlVoltage, boolean reactiveLimits, double qToDispatch) {
         double residueQ = 0;
         double calculatedQ = qToDispatch / generatorsThatControlVoltage.size();
```
AnkurArohi, line 420: Here it should be important to exclude generatorsThatControlVoltage.size() == 0

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -415,19 +415,19 @@ public void addHvdc(LfHvdc hvdc) {
...
         Iterator<LfGenerator> itG = generatorsThatControlVoltage.iterator();
         while (itG.hasNext()) {
             LfGenerator generator = itG.next();
-            if (reactiveLimits && calculatedQ < generator.getMinQ()) {
-                generator.setCalculatedQ(generator.getCalculatedQ() + generator.getMinQ());
-                residueQ += calculatedQ - generator.getMinQ();
+            if (reactiveLimits && calculatedQ + generator.getCalculatedQ() < generator.getMinQ()) {
+                residueQ += calculatedQ + generator.getCalculatedQ() - generator.getMinQ();
```
AnkurArohi, line 425: This residue Q must be correctly handled

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
EtienneLt: it is handled by the method calling it : 
while (!generatorsThatControlVoltage.isEmpty() && Math.abs(qToDispatch) > Q_DISPATCH_EPSILON) {
            qToDispatch = dispatchQ(generatorsThatControlVoltage, reactiveLimits, qToDispatch);
        }

EtienneLt: but I can add a check

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
EtienneLt: Yes but I dont know what is wrong ?

AnkurArohi: residueQ must at the end be redistributed so that all the Q is correctly compensated, is that done when yes then all right

EtienneLt: well the thing is it sets the calculatedQ to the max or min bound so the residueQ needs to keep the CalculatedQ that was set on the generator before this. 
I can add a variable with the generator.getCalculatedQ() like that 
```java
if (reactiveLimits && calculatedQ + generator.getCalculatedQ() < generator.getMinQ()) {
                double generatorOldQ = generator.getCalculatedQ();
                generator.setCalculatedQ(generator.getMinQ());
                residueQ += calculatedQ + generatorOldQ - generator.getMinQ();
                
```

AnkurArohi: I think this is problematic
What is the difference between calculatedQ and generator.getCalculatedQ()

AnkurArohi: and even then you need to keep redistributing residueQ till its zero because it has to be conserved Q produced = QConsumed + Q losses

EtienneLt: calculatedQ is equal to dispatchQ/number of generators it is Q we ideally want to set to each generator at each loop.
generator.getCalculatedQ() is Q that was set previously to the generator in a previous loop because we try to dispatch Q with loops : 
```java
while (!generatorsThatControlVoltage.isEmpty() && Math.abs(qToDispatch) > Q_DISPATCH_EPSILON) {
            qToDispatch = dispatchQ(generatorsThatControlVoltage, reactiveLimits, qToDispatch);
}
```

EtienneLt: dispatchedQ is not really a good name it should be something like remainingQToDispatch

AnkurArohi: I would suggest refactoring calculatedQ to tobedispatchedQ_pergen

AnkurArohi: Here we need to include the factor for distributed slack  if any
so that the tobedispatchedQ_pergen would be multiplied by this but not in this PR i guess

AnkurArohi:
Sorry by mistake it was closed

AnkurArohi:
I do not have the option to merge the request but I see all good Thanks.

EtienneLt:
> I do not have the option to merge the request but I see all good Thanks.

EtienneLt:
> I do not have the option to merge the request but I see all good Thanks.

Thank you for the review !

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
EtienneLt: It should not happen so I added a exception

EtienneLt:
> Aha, I still have a point but it is probably seldom this case lets say even after many iterations the tobedispatchedq is not below the tolerance limit what do we do now?

I think all the generators will be at the max or min q bound so the loop stops.
But you're right when this happens there should be at least a log

AnkurArohi:
> >

But then additional to the log the amount of tobedispatchedq should be then dispatched to the slack all of it ,naturally we need another Load Flow to confirm that the network is secured and stable but ideally at the end there should be no tobedispatchedQ left

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowEurostagTutorialExample1Test.java
```diff
@@ -409,4 +409,26 @@ void testWithDisconnectedGenerator() {
         assertActivePowerEquals(Double.NaN, gen.getTerminal());
         assertReactivePowerEquals(Double.NaN, gen.getTerminal());
     }
+
+    @Test
+    void testGeneratorReactiveLimits() {
```
sylvlecl, line 414: Actually, that test also passed without the evolution --> we should try to adapt it so that it really tests the fix.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowEurostagTutorialExample1Test.java
```diff
@@ -409,4 +409,26 @@ void testWithDisconnectedGenerator() {
...
+                .setBus("NGEN").setConnectableBus("NGEN")
+                .setMinP(-9999.99D).setMaxP(9999.99D)
+                .setVoltageRegulatorOn(true).setTargetV(24.5D)
+                .setTargetP(607.0D).setTargetQ(301.0D).add();
+        network.getGenerator("GEN1").newMinMaxReactiveLimits().setMinQ(0).setMaxQ(220).add();
+        LoadFlowParameters parameters = new LoadFlowParameters().setNoGeneratorReactiveLimits(false)
+                .setDistributedSlack(false)
+                .setVoltageInitMode(LoadFlowParameters.VoltageInitMode.DC_VALUES);
+        loadFlowRunner.run(network, parameters);
+        network.getGenerators().forEach(gen -> {
```
sylvlecl, line 427: It would be better to have simple assertions with hard coded value, for example checking that generator GEN1 is at its maximum limit (120) and GEN2 is (or is not I don't know :) ).

sylvlecl (commented):
We should try to improve the "integration test": it passed before the fix

AnkurArohi:
@EtienneLt 
The suggestion for the case of distributed slacks is an intresting one
should i create a new issue or how do we handle this

EtienneLt:
> @EtienneLt The suggestion for the case of distributed slacks is an intresting one should i create a new issue or how do we handle this

It seems that it is already implemented

AnkurArohi:
🏁🏁🏁🏁🏁😉

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -415,22 +415,26 @@ public void addHvdc(LfHvdc hvdc) {
         hvdcs.add(Objects.requireNonNull(hvdc));
     }
 
-    private static double dispatchQ(List<LfGenerator> generatorsThatControlVoltage, boolean reactiveLimits, double qToDispatch) {
+    protected static double dispatchQ(List<LfGenerator> generatorsThatControlVoltage, boolean reactiveLimits, double qToDispatch) {
         double residueQ = 0;
-        double calculatedQ = qToDispatch / generatorsThatControlVoltage.size();
+        if (generatorsThatControlVoltage.isEmpty()) {
```
annetill, line 420: This check is not needed because it is done just before calling this method.

AnkurArohi: This was my suggestion and was to ensure that through debug log User is aeare that there has indeed been no redistribution

AnkurArohi: Plus we should always consider

Separation of concerns based implementation

PR #533, opened 2022-05-12
implements update of specific parameters

EtienneLt:

feature

It implements new method from powsybl core loadflow provider to update specific parameters


src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -333,8 +333,7 @@ public static OpenLoadFlowParameters load(PlatformConfig platformConfig) {
         return parameters;
     }
 
-    public static OpenLoadFlowParameters load(Map<String, String> properties) {
-        OpenLoadFlowParameters parameters = new OpenLoadFlowParameters();
+    public static OpenLoadFlowParameters load(OpenLoadFlowParameters parameters, Map<String, String> properties) {
```
sylvlecl, line 336: I think we can keep the old one and add this one as a "not static" method (then the old one can call the new one)

PR #545, opened 2022-06-05
Assert active power, reactive power and current consistency

geofjamg:

No

Unit test

A unit test has been added to assert p, q and i equations consistency

**Other information**:

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/ac/AcEquationsTest.java
+++ b/src/test/java/com/powsybl/openloadflow/ac/AcEquationsTest.java
@@ -22,6 +22,7 @@ import org.mockito.stubbing.Answer;
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 /**
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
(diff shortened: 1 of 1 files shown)


src/test/java/com/powsybl/openloadflow/ac/AcEquationsTest.java
```diff
@@ -141,6 +142,28 @@ void branchTest() {
...
+        var p1Eq = new ClosedBranchSide1ActiveFlowEquationTerm(branch, bus1, bus2, variableSet, true, true);
+        p1Eq.setStateVector(sv);
+        double p1 = p1Eq.eval();
+        var q1Eq = new ClosedBranchSide1ReactiveFlowEquationTerm(branch, bus1, bus2, variableSet, true, true);
+        q1Eq.setStateVector(sv);
+        double q1 = q1Eq.eval();
+        var i1Eq = new ClosedBranchSide1CurrentMagnitudeEquationTerm(branch, bus1, bus2, variableSet, true, true);
+        i1Eq.setStateVector(sv);
+        double i1 = i1Eq.eval();
+        assertEquals(i1, Math.hypot(p1, q1) / V_1, 10e-14);
```
geofjamg, line 155: @annetill is it normal to not have the sqrt(3) here ?

annetill: After looking at the code, yes it is normal. The sqrt(3) is managed by function PerUnit.ib() for sensitivity and security analysis.

geofjamg: It doesn't mean it is correct. As far as I understand main principle of peruniting is to perunit only 2 quantities (for us it is apparent power and voltage) and all the other quantities are calculated from the 2 chosen ones.
So for me, the values in per unit are not correct even if one in SI are correct.

geofjamg: According https://en.wikipedia.org/wiki/Per-unit_system it seems correct

annetill (approved):

PR #549, opened 2022-06-07
Current limit violations detection in DC security analysis

Hadrien-Godard:

This PR introduces the handling of current limit violations in the DC security analysis.

**Other information**:


src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -91,6 +93,8 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
 
     public static final String TRANSFORMER_VOLTAGE_CONTROL_MODE_NAME = "transformerVoltageControlMode";
 
+    public static final String DC_COS_PHI_NAME = "dcCosPhi";
```
geofjamg, line 96: @annetill are you sure cos(phi) is the right naming? phi is used in OLF as the voltage angle. Would power factor more usual way of naming this?

annetill: Yes right, my mistake, I am going to fix the name.

geofjamg (approved):

PR #572, opened 2022-06-29
[WIP] Support generators connected to same bus but controlling a different bus

geofjamg:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
@@ -488,7 +488,7 @@ public class LfNetwork extends AbstractPropertyBag implements PropertyBag {
                 LfBus bus1 = branch.getBus1();
                 LfBus bus2 = branch.getBus2();
                 // ensure target voltages are consistent
-                if (bus1 != null && bus2 != null) {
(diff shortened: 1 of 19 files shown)


geofjamg:
@annetill @Hadrien-Godard unit test fails with "Expected to have same number of equations (12) and variables (11)". Equation system seems to be inconsistent. Any idea why?

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -488,7 +488,7 @@ private void validateBranches(boolean minImpedance, boolean dc) {
                 LfBus bus1 = branch.getBus1();
                 LfBus bus2 = branch.getBus2();
                 // ensure target voltages are consistent
-                if (bus1 != null && bus2 != null) {
+                if (bus1 != null && bus2 != null && bus1.isVoltageControlled() && bus2.isVoltageControlled()) {
```
geofjamg, line 491: @annetill to handle in a separate PR, but this check is buggy and has to be fixed. Also I am not sure it is needed because we already check this at network loading.

geofjamg: But it does not check it is regulating

PR #573, opened 2022-07-01
Zero impedance flows. Specific solution for open Loadflow

marqueslanauja:
Signed-off-by: José Antonio Marqués <marquesja@aia.es>

Feature

The LoadflowResultsCompletion method is used. This method is only valid for lines.

A general method has been coded for the openLoadflow. This method is able to compute flows for all zero impedance equipment, lines, twoWindingsTransformers, threeWindingsTransformers (including all configurations, one , two or all three legs with zero impedance), danglingLines and switches (the flow of the switches can not be recorded in iidm). 
The method works with both loadflows, Ac and Dc.

In the openLoadflow the algorithm to compute zero impedance flows is easier and more general than in LoadflowResultsCompletion since the LfNetwork model includes the star bus of the threeWindings transformers and the boundary bus of the danglingLines.

At the threeWindingsTransformers, to compute the zero impedance flow we have to compute the flow of the legs with impedance at the star bus side. To do that I modified the `AbstractFictitiousLfBranch` class to allow the evaluation of the flow at both ends. The flow can always be evaluated at the lfNetwork level since the star bus is defined (also valid for danglingLines). 

Since the method works for danglingLines, I have undone a previous fix for calculating the danglingLIne zero impedance flow (`LfDanglingLineBranch` class). Better only a way to compute all the zero impedance flow for all equipment.

**Other information**:


annetill:
> The flow of the lines outside the tree must be zero. As the lines have zero impedance we take the agreement that the flow goes only through the lines in the tree, and there is no flow in the lines outside the tree. In the easiest case, two zero impedance parallel lines one of the lines will carry all the flow and the other zero.

Can you review my small changes?

marqueslanauja:
> > The flow of the lines outside the tree must be zero. As the lines have zero impedance we take the agreement that the flow goes only through the lines in the tree, and there is no flow in the lines outside the tree. In the easiest case, two zero impedance parallel lines one of the lines will carry all the flow and the other zero.
> 
> Can you review my small changes?

I reviewed your commits (Small refactorings and clean) and all changes are ok.

geofjamg (approved):

geofjamg:
@annetill by trying to fix the code duplication issue, I changed a unit test. I think it was a mistake because for dangling lines and leg branches, we didn't store side 2 current (why ????, it seems to me a valid infos).

annetill:
> @annetill by trying to fix the code duplication issue, I changed a unit test. I think it was a mistake because for dangling lines and leg branches, we didn't store side 2 current (why ????, it seems to me a valid infos).

Just because we thought that we don't need it, but it was before we notice that zero impedance lines were not handled.

annetill (approved):

PR #576, opened 2022-07-03
Security analysis functional logs

geofjamg:

No

Feature

There is no functional logs for security analysis

Minimal functionnal logs have been added 

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
@@ -79,7 +79,7 @@ public class LfNetwork extends AbstractPropertyBag implements PropertyBag {
 
     private GraphDecrementalConnectivity<LfBus, LfBranch> connectivity;
 
-    private final Reporter reporter;
+    private Reporter reporter;
 
     public LfNetwork(int numCC, int numSC, SlackBusSelector slackBusSelector,
                      GraphDecrementalConnectivityFactory<LfBus, LfBranch> connectivityFactory, Reporter reporter) {
@@ -105,6 +105,10 @@ public class LfNetwork extends AbstractPropertyBag implements PropertyBag {
 
     public Reporter getReporter() {
         return reporter;
+    }
+
+    public void setReporter(Reporter reporter) {
+        this.reporter = Objects.requireNonNull(reporter);
     }
 
     private void invalidateSlack() {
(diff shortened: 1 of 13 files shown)


src/test/resources/saReport.txt
```diff
@@ -0,0 +1,31 @@
++ Test security analysis report
+  + AC security analysis on network 'sim1'
+    + Network CC0 SC0
+      + Post loading processing
+         Network has 4 buses and 4 branches
+         Network balance: active generation=607.0 MW, active load=600.0 MW, reactive generation=0.0 MVar, reactive load=200.0 MVar
+      + Pre-contingency simulation
+         AC load flow complete with NR status 'CONVERGED'
```
geofjamg, line 8: @annetill @flo-dup here we can see an issue with reporter API, we don't keep sequence of reports and sub reporters

src/test/resources/saReport.txt
```diff
@@ -0,0 +1,31 @@
...
+        + Outer loop Distributed slack on generation
+           Iteration 0: failed to distribute slack bus active power mismatch, -1.4404045651219555 MW remains
+        + Outer loop Reactive limits
```
geofjamg, line 11: Maybe here another possible improvement: should we discard empty child reporters?

geofjamg:
https://github.com/powsybl/powsybl-core/issues/2190
https://github.com/powsybl/powsybl-core/issues/2190
https://github.com/powsybl/powsybl-core/issues/2190

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -107,6 +107,10 @@ public Reporter getReporter() {
         return reporter;
     }
 
+    public void setReporter(Reporter reporter) {
```
flo-dup, line 110: Here you're breaking in a way the way the reporter was at first designed... It sure is handy to put it in LfNetwork but it was first thought to be passed along all the methods and not retained, in the javadoc:

> A reporter is not meant to be shared with other threads nor to be saved as a class parameter

geofjamg: A LF network won't ever be shared between multiple thread as it is used as working data (including writing working data)

flo-dup (approved):

PR #580, opened 2022-07-07
Support of switch contingency in sensitivity analysis and refactoring

annetill:

No.

In case of a switch contingency, and without any `BUS_VOLTAGE` sensitivity function, the sensitivity analysis will be based on bus/breaker view. In the other cases, it remains on the bus view. The contingency propagation is not supported in the PR because it leads to other issues such as shared controls in a zero impedance network: it is not handle because zero impedance branches are managed with a min impedance.

Coding this, I have made a refactoring to unify the treatments between security analysis and sensitivity analysis. Without any switch to open in a security analysis, the computation is based on the bus/view and not longer in the bus/breaker view.

**Other information**:


src/main/java/com/powsybl/openloadflow/sa/AbstractNetworkResult.java
```diff
@@ -54,7 +54,8 @@ protected void addResults(StateMonitor monitor, Consumer<LfBranch> branchConsume
             network.getBuses().stream()
                     .filter(lfBus -> monitor.getVoltageLevelIds().contains(lfBus.getVoltageLevelId()))
                     .filter(lfBus -> !lfBus.isDisabled())
-                    .forEach(lfBus -> busResults.add(lfBus.createBusResult()));
+                    .filter(lfBus -> !lfBus.isFictitious())
```
annetill, line 57: @geofjamg just tell me if we have to avoid fictitious buses here.

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -87,4 +90,10 @@ public void updateState(boolean reactiveLimits, boolean writeSlackBus, boolean d
     public boolean isParticipating() {
         return participating;
     }
+
+    @Override
+    public List<BusResult> createBusResults() {
+        return bus.getVoltageLevel().getBusBreakerView().getBusesFromBusViewBusId(bus.getId())
```
geofjamg, line 96: Does it still work when if bus is a bus/breaker one?

annetill: No :-) I am on it.

geofjamg (approved):
Let's merge it as it is. The issue with missing bus results is not linked to this PR.

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -174,6 +175,14 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
         Objects.requireNonNull(resultWriter);
         Objects.requireNonNull(reporter);
 
+        boolean breakers = !allSwitchesToOpen.isEmpty();
+        if (breakers && hasBusTargetVoltage(factorReader)) {
```
geofjamg, line 179: @annetill Maybe to check the performance impact because we fully read factors 2 times with this. AC sensitivity analysis is quite slow so probably no impact, but to check with a larger number of factor.

geofjamg: Maybe could have been done inside the actual factor read method without doing it 2 times?

geofjamg: I can see that it was already the case to check for transformer voltage control. So new we read all factors and resolve to network 3 times.

annetill: Fixed.

PR #587, opened 2022-07-29
Remove starting generators from voltage control

annetill:

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
```diff
@@ -211,6 +211,11 @@ protected boolean checkVoltageControlConsistency(boolean reactiveLimits, LfNetwo
             report.generatorsDiscardedFromVoltageControlBecauseNotStarted++;
             consistency = false;
         }
+        if (getTargetP() < getMinP() && getMinP() > POWER_EPSILON_SI) {
```
geofjamg, line 214: What is the meaning of "getMinP() > POWER_EPSILON_SI" ?

annetill: It means that this condition means something only for positive Pmin. For negative Pmin, I don’t know what is a starting generator…

geofjamg: yes, but my question wa more about the POWER_EPSILON_SI: "getMinP() > 0"  or even maybe "getMinP() >= 0" would make sense too.

annetill: Indeed I just have copied the lines above in the code, that contains that... the good assertion is maybe Pmin > 0.

geofjamg (approved):

PR #590, opened 2022-08-17
AC Security Analysis with operator strategies

annetill:

No

Feature

We are able to simulate operator strategies, like switch closing and opening, phase shifter tap change, line reconnection.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,108 @@
...
+
+        // update connectivity with disabled branches
+        List<LfBranch> disabledBranches = network.getBranches().stream()
+                .filter(LfElement::isDisabled)
+                .collect(Collectors.toList());
+        GraphConnectivity<LfBus, LfBranch> connectivity = network.getConnectivity();
+        connectivity.startTemporaryChanges();
+        disabledBranches.stream()
+                .filter(b -> b.getBus1() != null && b.getBus2() != null)
+                .filter(b -> !(b instanceof LfSwitch)) // FIXME
```
annetill, line 84: @geofjamg I have to put this ugly and not working filter because I cannot cut twice the same branch... and all the switches to be potentially closed by a remedial action as already been cut after network loading.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -135,13 +189,31 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
             network.setReporter(preContSimReporter);
 
             // run pre-contingency simulation
-            AcLoadFlowResult preContingencyLoadFlowResult = new AcloadFlowEngine(context)
-                    .run();
+            // FIXME
```
annetill, line 192: @geofjamg here is ugly too because I need the equation system and the listener to be created in order to listen to the `setDisabled` I am doing on switches to close.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,114 @@
...
+        if (withConnectivity) {
+            // update connectivity with disabled branches
+            List<LfBranch> disabledBranches = network.getBranches().stream()
+                    .filter(LfElement::isDisabled)
+                    .collect(Collectors.toList());
+            GraphConnectivity<LfBus, LfBranch> connectivity = network.getConnectivity();
+            connectivity.startTemporaryChanges();
+            disabledBranches.stream()
+                    .filter(b -> b.getBus1() != null && b.getBus2() != null)
+                    .filter(b -> !alreadyRemovedEdges.contains(b.getId())) // FIXME
```
annetill, line 89: @geofjamg I have to put this ugly and not working filter because I cannot cut twice the same branch... and all the switches to be potentially closed by a remedial action as already been cut after network loading.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -122,8 +144,40 @@ public static void distributedMismatch(LfNetwork network, double mismatch, LoadF
...
+
+    private static Map<String, List<OperatorStrategy>> indexOperatorStrategiesByContingencyId(List<PropagatedContingency> propagatedContingencies,
+                                                                                              List<OperatorStrategy> operatorStrategies) {
+        Set<String> contingencyIds = propagatedContingencies.stream().map(propagatedContingency -> propagatedContingency.getContingency().getId()).collect(Collectors.toSet());
+        Map<String, List<OperatorStrategy>> operatorStrategiesByContingencyId = new HashMap<>();
+        for (OperatorStrategy operatorStrategy : operatorStrategies) {
+            if (contingencyIds.contains(operatorStrategy.getContingencyId())) {
+                operatorStrategiesByContingencyId.computeIfAbsent(operatorStrategy.getContingencyId(), key -> new ArrayList<>()).add(operatorStrategy);
+            } else {
+                LOGGER.warn("An operator strategy linked to Contingency {} that is not present in the list of Contingencies", operatorStrategy.getContingencyId());
```
geofjamg, line 172: I would throw an exception

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -122,8 +144,40 @@ public static void distributedMismatch(LfNetwork network, double mismatch, LoadF
         }
     }
 
+    public static void getAllSwitchesToOperate(Network network, List<Action> actions, Set<Switch> allSwitchesToClose, Set<Switch> allSwitchesToOpen) {
+        actions.stream().filter(action -> action.getType().equals(SwitchAction.NAME))
+                .map(action -> ((SwitchAction) action).getSwitchId())
+                .forEach(id -> {
+                    Switch sw = network.getSwitch(id);
+                    if (sw.isOpen()) {
+                        allSwitchesToClose.add(sw);
```
geofjamg, line 153: How do we known that the switch will be closed if not looking at action state?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -98,18 +113,25 @@ SecurityAnalysisReport runSync(String workingVariantId, SecurityAnalysisParamete
...
+        String variantId = network.getVariantManager().getWorkingVariantId();
         network.getVariantManager().cloneVariant(network.getVariantManager().getWorkingVariantId(), tmpVariantId);
+        network.getVariantManager().setWorkingVariant(tmpVariantId);
         try {
             network.getSwitchStream().filter(sw -> sw.getVoltageLevel().getTopologyKind() == TopologyKind.NODE_BREAKER)
                     .forEach(sw -> sw.setRetained(false));
             allSwitchesToOpen.forEach(sw -> sw.setRetained(true));
+            allSwitchesToClose.forEach(sw -> {
+                sw.setRetained(true);
+                sw.setOpen(false); // in order to be present in the network.
```
geofjamg, line 129: This is the ugly part of this PR. Not easy to do otherwise.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -135,13 +189,25 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
             network.setReporter(preContSimReporter);
 
             // run pre-contingency simulation
-            AcLoadFlowResult preContingencyLoadFlowResult = new AcloadFlowEngine(context)
-                    .run();
+            AcloadFlowEngine engine = new AcloadFlowEngine(context);
+            engine.getContext().getEquationSystem();
+            if (!allSwitchesToClose.isEmpty()) {
+                var connectivity = network.getConnectivity();
+                allSwitchesToClose.stream().map(Identifiable::getId).forEach(id -> {
```
geofjamg, line 196: I'am going to think about a way to avoid this hack (probably no way...)

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -173,6 +239,22 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
...
+                                    if (operatorStrategiesForThisContingency.size() == 1) {
+                                        Optional<OperatorStrategyResult> optionalOperatorStrategyResult = runActionSimulation(network, context, propagatedContingency.getContingency(),
+                                                operatorStrategiesForThisContingency.get(0), preContingencyLimitViolationManager,
+                                                securityAnalysisParameters.getIncreasedViolationsParameters(), postContingencyResult.getLimitViolationsResult(), lfActionById,
+                                                preContingencyNetworkResult, createResultExtension, allSwitchesToClose.stream().map(Switch::getId).collect(Collectors.toList()));
+                                        if (optionalOperatorStrategyResult.isPresent()) {
+                                            operatorStrategyResults.add(optionalOperatorStrategyResult.get());
+                                        }
+                                    } else {
+                                        LOGGER.warn("A contingency has several operator strategies: not supported yet");
```
geofjamg, line 254: Do you plan to implement it in this PR?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -230,4 +312,96 @@ private PostContingencyResult runPostContingencySimulation(LfNetwork network, Ac
...
+                    .map(id -> {
+                        LfAction lfAction = lfActionById.get(id);
+                        if (lfAction == null) {
+                            throw new PowsyblException("Action '" + id + "' of operator strategy '" + operatorStrategy.getId() + "' not found");
+                        }
+                        return lfAction;
+                    })
+                    .collect(Collectors.toList());
+            // FIXME to remove when connectivity will be fixed
+            operatorStrategyLfActions.stream().limit((long) operatorStrategyLfActions.size() - 1).forEach(LfAction::apply);
```
geofjamg, line 338: As far as I understand this won't be necessary anymore when addEdge and removeEdge will be callable multiple tile with same branch without any exception?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -230,4 +312,96 @@ private PostContingencyResult runPostContingencySimulation(LfNetwork network, Ac
...
+            Stopwatch stopwatch = Stopwatch.createStarted();
+
+            // restart LF on post contingency and post actions equation system
+            context.getParameters().setVoltageInitializer(new PreviousValueVoltageInitializer());
+            AcLoadFlowResult postActionsLoadFlowResult = new AcloadFlowEngine(context)
+                    .run();
+
+            boolean postActionsComputationOk = postActionsLoadFlowResult.getNewtonRaphsonStatus() == NewtonRaphsonStatus.CONVERGED;
+            // FIXME: to be checked.
+            var postActionsViolationManager = new LimitViolationManager(preContingencyLimitViolationManager, violationsParameters);
```
geofjamg, line 350: So here reference is still the base case, for violation increase criteria?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -230,4 +312,96 @@ private PostContingencyResult runPostContingencySimulation(LfNetwork network, Ac
...
+            var postActionsNetworkResult = new PostContingencyNetworkResult(network, monitorIndex, createResultExtension, preContingencyNetworkResult, contingency);
```
geofjamg, line 351: I don't think this is correct to use PostContingencyNetworkResult, flow transfer will still be calculated for contingency. What do we really want here?

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,114 @@
...
+        apply(false, Collections.emptyList());
+    }
+
+    public void apply(boolean withConnectivity, List<String> alreadyRemovedEdges) {
+        if (disabledBranch != null) {
+            disabledBranch.setDisabled(true);
+        }
+        if (enabledBranch != null) {
+            enabledBranch.setDisabled(false);
+            enabledBranch.getBus1().setDisabled(false);
```
geofjamg, line 80: Do we need to enable buses ? In which case?

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private final String id;
+
+    private LfBranch disabledBranch; // switch to open
+
+    private LfBranch enabledBranch; // switch to close
+
+    private LfNetwork network;
+
+    public LfAction(Action action, LfNetwork network) {
```
geofjamg, line 37: So the plan here is to support a list of action and that LfAction will be the sum/result of PowSyBl Core actions for the current strategy?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
annetill: It is not so simple: it is also for performance issues. As we can have a list of actions, the connectivity is computed after applying the last one in order to find propagated effects. The FIXME is for 'allSwitchesToCloseIds' argument of the apply, that will be removed once remove/add of edge can be done several times.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
annetill: Right, thanks!

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
annetill: :-) thanks! Note that there is for me a bug in the current code because we create a variant but we don't use it.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
annetill: No, because it requires a save/restore of the network state just after contingency.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
annetill: @sylvlecl and @EtienneLt : it was my question... I don't know :-) Any opinion ?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
geofjamg: Connectivity is lazily computed, so we cannot have perf issue

src/main/java/com/powsybl/openloadflow/network/LfAction.java
annetill: Yes we need it (see the unit tests) and indeed in it really not enough. For example, when we reconnect a switch that will reconnect a small component, the buses and branches of this component but be enabled. This is not handle here. @flo-dup  how to do this with your connectivity?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
annetill: @sylvlecl and @EtienneLt : good question !

src/main/java/com/powsybl/openloadflow/network/LfAction.java
annetill: No: we keep a list of LfAction for an operator strategy. I am not able to do better...

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
annetill: Internal processes compare post-action state to pre-contingency state to detect worsened violations.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
annetill: Flow transfer will be at NaN after actions.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
geofjamg: When all design issues will be solved it should be easy to support

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,103 @@
...
+        connectivity.startTemporaryChanges();
+
+        for (LfAction action : actions) {
+            action.apply(connectivity);
+        }
+
+        Set<LfBus> buses = connectivity.getSmallComponents().stream().flatMap(Set::stream).collect(Collectors.toSet());
+        for (LfBus bus : buses) {
+            bus.setDisabled(true);
+            bus.getBranches().forEach(branch -> branch.setDisabled(true));
```
geofjamg, line 81: I don't think this is correct to disable all branch neighbors. The right thing to do is to disabled branches part of the small components.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
geofjamg: So yes, this is not the right thing to do. We should be able to get from connectivity, the list of buses and branches that have been reconnected to main connected component. Here by just enabling side buses, we only enable a small part of potential sub network reconnected.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,103 @@
...
+        Objects.requireNonNull(network);
+
+        var connectivity = network.getConnectivity();
+        connectivity.startTemporaryChanges();
+
+        for (LfAction action : actions) {
+            action.apply(connectivity);
+        }
+
+        Set<LfBus> buses = connectivity.getSmallComponents().stream().flatMap(Set::stream).collect(Collectors.toSet());
```
geofjamg, line 78: A better approach would be to get from temporary changes, the list of buses and branches that have been disconnected. Here using small components we also get already disconnected buses (on base case, or by the contingency)

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,103 @@
...
+        for (LfBus bus : buses) {
+            bus.setDisabled(true);
+            bus.getBranches().forEach(branch -> branch.setDisabled(true));
+        }
+        connectivity.undoTemporaryChanges();
+
+        network.getBuses().forEach(bus -> LOGGER.info("LfBus {} is disabled: {}", bus.getId(), bus.isDisabled()));
```
geofjamg, line 85: just for debug. to refactor to only print the list of buses and branches enables and disabled during this temporary connectivity change.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -135,13 +191,25 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
             network.setReporter(preContSimReporter);
 
             // run pre-contingency simulation
-            AcLoadFlowResult preContingencyLoadFlowResult = new AcloadFlowEngine(context)
-                    .run();
+            AcloadFlowEngine engine = new AcloadFlowEngine(context);
+            engine.getContext().getEquationSystem();
```
geofjamg, line 195: I missed this in my first review :-)
This is a workaround, and it should be fixed
As far as I understand, you do this to force the equation system creation and its connection to network though listener. 
But the right fix is to be able is equation system creation to better handle initially disabled buses and branches. (in actual code we probably suppose all are initially enabled)  so that we can create equation system from any network with some already disable elements.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
geofjamg: And why is it a problem? Is there something different to do that what we are doing for contingencies?

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -1697,4 +1736,157 @@ void reportTest() throws IOException {
 
         assertEquals(refLogExport, logExport);
     }
+
+    @Test
+    void testSecurityAnalysisWithOperatorStrategy() {
+        MatrixFactory matrixFactory = new DenseMatrixFactory();
+        GraphConnectivityFactory<LfBus, LfBranch> connectivityFactory = new NaiveGraphConnectivityFactory<>(LfBus::getNum);
+        securityAnalysisProvider = new OpenSecurityAnalysisProvider(matrixFactory, connectivityFactory);
```
geofjamg, line 1744: to fix by reusing the runSecurityAnalysis with a different connectivity factory

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/network/LfAction.java
geofjamg: Now fixed thanks to #593

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
geofjamg: Now, refactored to avoid this

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
geofjamg: Ok make sense

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
annetill: I don't understand your comment. If I use another factory, it is not working...

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public static void apply(List<LfAction> actions, LfNetwork network, LfContingency contingency) {
+        Objects.requireNonNull(actions);
+        Objects.requireNonNull(network);
+
+        GraphConnectivity<LfBus, LfBranch> connectivity = network.getConnectivity();
+        connectivity.setMainComponentVertex(network.getSlackBus());
+        connectivity.startTemporaryChanges();
+        contingency.getDisabledBranches().stream()
+                .filter(Objects::nonNull) // could be in another component
```
geofjamg, line 74: How could is be null?

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,116 @@
...
+                .filter(b -> b.getBus1() != null && b.getBus2() != null)
```
geofjamg, line 75: Already done in PropagatedContingency?

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,116 @@
...
+                .forEach(connectivity::removeEdge);
+        Set<LfBus> postContingencyRemovedBuses = connectivity.getVerticesRemovedFromMainComponent();
+        postContingencyRemovedBuses.stream().forEach(bus -> bus.setDisabled(false)); // undo.
```
geofjamg, line 78: Why do you need to undo? Why not just getting directly contingency + action state?

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -123,13 +170,25 @@ private SecurityAnalysisResult runSimulations(LfNetwork network, List<Propagated
...
+            if (!allSwitchesToClose.isEmpty()) {
+                var connectivity = network.getConnectivity();
+                allSwitchesToClose.stream().map(Identifiable::getId).forEach(id -> {
+                    LfBranch branch = network.getBranchById(id);
+                    branch.setDisabled(true);
```
geofjamg, line 179: @annetill we only re-enable switches to close, but we should do it to for sub network that could be move out of main cc by this branch disabling?

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    private static void updateConnectivity(List<LfAction> actions, LfNetwork network, LfContingency contingency) {
+        GraphConnectivity<LfBus, LfBranch> connectivity = network.getConnectivity();
+        connectivity.setMainComponentVertex(network.getSlackBus());
+
+        // re-update connectivity according to post contingency state (revert after LfContingency apply)
+        connectivity.startTemporaryChanges();
+        contingency.getDisabledBranches().forEach(connectivity::removeEdge);
```
annetill, line 131: Do we support a bus contingency (bus/bar section or bus from a bus/breaker topology)? If one day yes, we also have to remove the disabled buses.

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -546,6 +546,7 @@ public GraphConnectivity<LfBus, LfBranch> getConnectivity() {
             getBranches().stream()
                     .filter(b -> b.getBus1() != null && b.getBus2() != null)
                     .forEach(b -> connectivity.addEdge(b.getBus1(), b.getBus2(), b));
+            connectivity.setMainComponentVertex(getSlackBus());
```
annetill, line 549: @flo-dup can you please just take a look at this change and tell us if it is okay for you. Thanks!

src/main/java/com/powsybl/openloadflow/network/LfAction.java
geofjamg: yes, we should add it now to not forget it later

PR #592, opened 2022-08-23
Aggredated loads properties

geofjamg:

No

Feature

Only aggregated load are available in LF model. 

For some use cases (short circuit calculation) we need to access all individual load and associate specific properties to each of them.
To not waste memory for classical use cases (where no detail is needed), lfloads are created lazily.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfLoadImpl.java
```diff
@@ -0,0 +1,40 @@
...
+        this.load = Objects.requireNonNull(load);
+    }
+
+    @Override
+    public String getId() {
+        return load.getId();
+    }
+
+    @Override
+    public double getTargetP() {
```
geofjamg, line 32: @annetill Maybe we should name it getInitialTargetP because it is not updated when aggregated load is updated?

src/main/java/com/powsybl/openloadflow/network/impl/LfAggregatedLoadsImpl.java
```diff
@@ -53,11 +58,12 @@ public void setAbsVariableLoadTargetP(double absVariableLoadTargetP) {
     }
 
     private void init() {
-        if (isInitialized) {
+        if (initialized) {
             return;
         }
 
         participationFactors = new double[loads.size()];
+        absVariableLoadTargetP = 0;
```
annetill, line 66: We also have `private double absVariableLoadTargetP = 0;`, so we must clean?

annetill (approved):

PR #627, opened 2022-10-17
Fix line open one side in security and sensitivity analyses

annetill:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
@@ -290,7 +290,7 @@ public class PropagatedContingency {
         // add to contingency description buses and branches that won't be part of the main connected
         // component in post contingency state
         Set<LfBus> buses = connectivity.getVerticesRemovedFromMainComponent();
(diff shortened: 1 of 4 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -292,6 +292,11 @@ public Optional<LfContingency> toLfContingency(LfNetwork network) {
         Set<LfBus> buses = connectivity.getVerticesRemovedFromMainComponent();
         Set<LfBranch> branches = connectivity.getEdgesRemovedFromMainComponent();
 
+        // we should manage branches open at one side.
+        for (LfBus bus : buses) {
+            bus.getBranches().stream().filter(b -> !b.isConnectedAtBothSides()).forEach(b -> branches.add(b));
```
geofjamg, line 297: Maybe we should copy branches because it could be an internal data structure of connectivity?

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
geofjamg: Fixed

PR #633, opened 2022-10-20
Operator strategy simulation fixes

geofjamg:

No

Bug fix

2 bugs:
 - if network element of action are not found in the LfNetwork, an exception in raised even if element is not just in same connected component.
 - a performance bug, state restoration is always done even if no action simulation has been done.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -181,10 +202,4 @@ public void apply() {
             branch.getPiModel().setTapPosition(newTapPosition);
         }
     }
-
-    private static void checkBranch(LfBranch branch, String branchId) {
```
annetill, line 185: So, in that case, no log to the user? And ok no exception.

annetill (approved):

src/main/java/com/powsybl/openloadflow/network/LfAction.java
geofjamg: I removed the check at lfnetwork level but it would be better to keep it on iidm network

annetill: There is nothing at iidm network I think.

geofjamg: Yes i will add it in this pr

geofjamg: I added to check of actions at IIDM network level and at LfNetwork level we can suppose that if not found it is in another component. what are we supposed to do in that case? Maybe we new status added in next core release, we can have the status "no impact" for this operator strategy result?

Also for IIDM check, it is relevant to throw an exception if an action does not exist in the network ? this is what we are doing for contingencies but maybe we should only warm the user that this contingency or action is invalid and discarded?

geofjamg:
@annetill it is now ready for review

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -174,23 +174,46 @@ private static void findAllSwitchesToOperate(Network network, List<Action> actio
...
+                        throw new PowsyblException("An action '" + action1.getId() + "' already exist");
+                    }
+                ));
+    }
+
     private static Map<String, List<OperatorStrategy>> indexOperatorStrategiesByContingencyId(List<PropagatedContingency> propagatedContingencies,
-                                                                                              List<OperatorStrategy> operatorStrategies) {
+                                                                                              List<OperatorStrategy> operatorStrategies,
+                                                                                              Map<String, Action> actionsById,
+                                                                                              Set<Action> neededActions) {
```
annetill, line 198: What I understand from previous talk with @sylvlecl and you is that the actions order matters. That is why I used a list instead of a set. With your modifications, are we sure that order is conserved?

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -131,20 +174,46 @@ private static void findAllSwitchesToOperate(Network network, List<Action> actio
...
         Set<String> contingencyIds = propagatedContingencies.stream().map(propagatedContingency -> propagatedContingency.getContingency().getId()).collect(Collectors.toSet());
         Map<String, List<OperatorStrategy>> operatorStrategiesByContingencyId = new HashMap<>();
         for (OperatorStrategy operatorStrategy : operatorStrategies) {
             if (contingencyIds.contains(operatorStrategy.getContingencyId())) {
-                operatorStrategiesByContingencyId.computeIfAbsent(operatorStrategy.getContingencyId(), key -> new ArrayList<>()).add(operatorStrategy);
+                // check actions IDs exists
+                for (String actionId : operatorStrategy.getActionIds()) {
```
geofjamg, line 204: @annetill this is here that the order is defined and it is a List

PR #635, opened 2022-10-22
Fast restart from previous LF

geofjamg:

No

Feature

When running multiple times a LF, we need to recreate each time all internal data structures (LF network, matrices, etc).

When parameter is activated, all internal data structures are kept into a cache, updated as best as we can and re-use for next runs.

**Other information**:


geofjamg:
@annetill switch removed

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowWithCachingTest.java
```diff
@@ -0,0 +1,141 @@
...
+        assertActivePowerEquals(64.706, l14.getTerminal1());
+        assertActivePowerEquals(129.412, l24.getTerminal1());
+        assertActivePowerEquals(-58.824, l34.getTerminal1());
+
+        l2.setP0(75);
+        l4.setP0(125);
+        result = loadFlowRunner.run(network, parameters);
+        assertEquals(LoadFlowResult.ComponentResult.Status.CONVERGED, result.getComponentResults().get(0).getStatus());
+        assertEquals(1, result.getComponentResults().get(0).getIterationCount());
+        assertActivePowerEquals(75.0, l2.getTerminal()); // FIXME, should be 88.235
```
geofjamg, line 91: @annetill this is because LfAggregatedLoadsImpl is not updated (absVariableLoadTargetP for load balance) and this is why I tried to fix in closed PR https://github.com/powsybl/powsybl-open-loadflow/pull/640

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,367 @@
...
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public enum NetworkCache {
+    INSTANCE;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NetworkCache.class);
+
+    public static class Entry extends DefaultNetworkListener {
```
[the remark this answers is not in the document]
flo-dup: ok: `entry.reset()` is explaining this

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,367 @@
...
+
+        private final WeakReference<Network> networkRef;
```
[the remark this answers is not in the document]
flo-dup: I got confused! if network is not referenced anywhere anymore there's no need to keep any reference to it.

src/main/java/com/powsybl/openloadflow/AcLoadFlowFromCache.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private List<AcLoadFlowContext> initContexts(NetworkCache.Entry entry) {
+        List<AcLoadFlowContext> contexts;
+        try (LfNetworkList lfNetworkList = Networks.load(network, acParameters.getNetworkParameters(), Collections.emptySet(), Collections.emptySet(), reporter)) {
+            contexts = lfNetworkList.getList()
+                    .stream()
+                    .map(n -> new AcLoadFlowContext(n, acParameters))
+                    .collect(Collectors.toList());
+            entry.setContexts(contexts);
+            entry.setVariantCleaner(lfNetworkList.release());
```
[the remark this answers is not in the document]
flo-dup: not so clear to use `release()` as it's not the `lfNetworkList` which is released, maybe `removeVariantCleaner()` or `detachNetwork()`?

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowWithCachingTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+        // check that cache entry could be correctly garbage collected
+        network = null;
+        int retry = 0;
+        do {
+            System.gc();
+            retry++;
+            TimeUnit.MILLISECONDS.sleep(100);
+        } while (NetworkCache.INSTANCE.getEntryCount() > 0 && retry < 10);
+        assertEquals(0, NetworkCache.INSTANCE.getEntryCount());
```
flo-dup, line 150:
```suggestion
        for (int i = 0; i < 10; i++) {
            loadFlowRunner.run(FictitiousSwitchFactory.create(), parameters);
        }
        assertEquals(10, NetworkCache.INSTANCE.getEntryCount());
        System.gc();
        assertTrue(NetworkCache.INSTANCE.getEntryCount() < 10);
```
I think we could change the logic by rising up the cache with several times the "same" network and look if at least one network has been removed. Note that I put the `System.gc()` outside the loop and not inside only to underline the logic (same holds for first assertion).

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,367 @@
...
+
+                return entry;
+            } else {
+                // restart from previous state
+                if (entry.getContexts() != null) {
+                    LOGGER.info("Network cache reused for network '{}'", network.getId());
+
+                    for (AcLoadFlowContext context : entry.getContexts()) {
+                        AcLoadFlowResult result = context.getResult();
+                        if (result != null && result.getNewtonRaphsonStatus() == NewtonRaphsonStatus.CONVERGED) {
```
flo-dup, line 341: Launching 3 times the load flow on a network:
- 1st time converging
- 2nd time: reusing previous values as converged before, but that time the NR diverges (as one load was changed too much for instance), `MAX_ITERATION_REACHED`
- 3rd time: we shouldn't use previous incoherent values, should we? but we are as the voltage initializer remains the same

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,367 @@
...
+
+        private static Bus getBus(Injection<?> injection, AcLoadFlowContext context) {
+            return context.getParameters().getNetworkParameters().isBreakers()
+                    ? injection.getTerminal().getBusBreakerView().getBus()
+                    : injection.getTerminal().getBusView().getBus();
+        }
+
+        private boolean onLoadUpdate(Load load, String attribute, Object oldValue, Object newValue) {
+            boolean found = false;
+            for (AcLoadFlowContext context : contexts) {
```
annetill, line 109: Finally, I think that for a first version the load support is not needed. It complexes the feature and it is not a real need for the IA use cases I know.

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,367 @@
...
+
+        private boolean onGeneratorUpdate(Generator generator, String attribute, Object oldValue, Object newValue) {
+            boolean found = false;
+            for (AcLoadFlowContext context : contexts) {
+                Bus bus = getBus(generator, context);
+                if (bus != null) {
+                    LfNetwork lfNetwork = context.getNetwork();
+                    LfBus lfBus = lfNetwork.getBusById(bus.getId());
+                    if (lfBus != null) {
+                        if (attribute.equals("targetV")) {
```
annetill, line 151: Just a remark here: in case of a shared control, the change of only one targetV will impact the voltage control targetV, even if we have other generators controlling that point with a different targetV.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowWithCachingTest.java
geofjamg: feature discarded from PR

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,367 @@
...
+                entry.close();
+                it.remove();
+                LOGGER.info("Dead network removed from cache ({} remains)", entries.size());
+            }
+        }
+    }
+
+    private boolean equals(LoadFlowParameters parameters1, LoadFlowParameters parameters2) {
+        try {
+            return objectMapper.writeValueAsString(parameters1)
```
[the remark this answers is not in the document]
geofjamg: I implemented parameters comparison.
And also parameters cloning as we need to store a copy of input parameters to later compare as same instance of parameters could be pass to multiple runs.

src/main/java/com/powsybl/openloadflow/AcLoadFlowFromCache.java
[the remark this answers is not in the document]
geofjamg: Waiting for a better name, we keep release for now.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowWithCachingTest.java
geofjamg: Ok done, but also disabled to test to avoid possible failures.

src/main/java/com/powsybl/openloadflow/NetworkCache.java
geofjamg: Ok, discarded from the PR

src/main/java/com/powsybl/openloadflow/NetworkCache.java
geofjamg: yes right

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,367 @@
...
+
+            if (entry == null) {
+                entry = new Entry(new WeakReference<>(network), parameters);
+                entries.add(entry);
+                network.addListener(entry);
+
+                LOGGER.info("Network cache created for network '{}'", network.getId());
+
+                return entry;
+            } else {
```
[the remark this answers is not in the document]
geofjamg: As discussed, I replaced the lock with a synchronized list.

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    private final List<Entry> entries = new ArrayList<>();
+
+    private final Lock lock = new ReentrantLock();
+
+    private final ObjectMapper objectMapper = new ObjectMapper()
+            .registerModule(new LoadFlowParametersJsonModule());
+
+    private void evictDeadEntries() {
```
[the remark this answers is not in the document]
geofjamg: It could be a good idea, but as discussed let's start simple and see later if we need to more aggressive cache cleanup.

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -0,0 +1,303 @@
...
+        public void close() {
+            reset();
+            if (variantCleaner != null) {
+                variantCleaner.clean();
+                variantCleaner = null;
+            }
+        }
+    }
+
+    private final List<Entry> entries = Collections.synchronizedList(new ArrayList<>());
```
flo-dup, line 220: Working with a synchronized list allowed you to remove the lock for unitary operations but as mentioned in the javadoc you still need to synchronize the list manually when iterating over it.

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/NetworkCache.java
geofjamg: CopyOnWriteArrayList should be the right one

flo-dup: Yes, you can do this with `CopyOnWriteArrayList` to avoid that, but you're not allowed to remove an entry with the iterator then in `evictDeadEntries()`, the unit test `testCacheEviction` fails. You need to do it with `entries.remove(entry)` then.

flo-dup: or `entries.removeIf(Predicate<Entry> p)`?

PR #636, opened 2022-10-23
SVC with stand by automaton 

annetill:

**Other information**:


geofjamg:
The only thing that should remain to do is to move standby automaton to its own outer loop. Looking at the code the standby automaton code seems to be totally disjoint from classical reactive limit management

src/main/java/com/powsybl/openloadflow/DefaultOuterLoopConfig.java
```diff
@@ -32,6 +32,7 @@ public List<OuterLoop> configure(LoadFlowParameters parameters, OpenLoadFlowPara
         if (!parameters.isNoGeneratorReactiveLimits()) {
             outerLoops.add(new ReactiveLimitsOuterLoop());
         }
+        outerLoops.add(new MonitoringVoltageOuterLoop());
```
geofjamg, line 35: Should be conditioned by reactiveLimits params too ?

src/main/java/com/powsybl/openloadflow/ac/MonitoringVoltageOuterLoop.java
```diff
@@ -0,0 +1,158 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * @author Anne Tilloy <anne.tilloy at rte-france.com>
+ */
+public class MonitoringVoltageOuterLoop implements OuterLoop {
```
geofjamg, line 28: StandbyAutomatonOuterLoop?

src/main/java/com/powsybl/openloadflow/ac/MonitoringVoltageOuterLoop.java
```diff
@@ -0,0 +1,158 @@
...
+        private PqToPvBus(LfBus controllerBus, VoltageLimitDirection voltageLimitDirection) {
+            this.controllerBus = controllerBus;
+            this.voltageLimitDirection = voltageLimitDirection;
+        }
+    }
+
+    @Override
+    public void initialize(OuterLoopContext context) {
+        for (LfBus bus : context.getNetwork().getBuses()) {
+            bus.setVoltageControlSwitchOffCount(0);
```
geofjamg, line 57: We don't need switch off counter for SVC?

src/main/java/com/powsybl/openloadflow/ac/MonitoringVoltageOuterLoop.java
```diff
@@ -0,0 +1,158 @@
...
+                    if (pqToPvBus.voltageLimitDirection == VoltageLimitDirection.MAX) {
+                        LOGGER.trace("Switch bus '{}' PQ -> PV with high targetV={}", controllerBus.getId(), newTargetV);
+                    } else {
+                        LOGGER.trace("Switch bus '{}' PQ -> PV with low targetV={}", controllerBus.getId(), newTargetV);
+                    }
+                }
+            }
+        }
+
+        Reports.reportPqToPvBuses(reporter, pqPvSwitchCount, 0);
```
geofjamg, line 85: We should have a dedicated log, this is not like a generator PQ PV bus

src/main/java/com/powsybl/openloadflow/ac/MonitoringVoltageOuterLoop.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public OuterLoopStatus check(OuterLoopContext context, Reporter reporter) {
+        OuterLoopStatus status = OuterLoopStatus.STABLE;
+
+        List<PqToPvBus> pqToPvBuses = new ArrayList<>();
+        for (LfBus bus : context.getNetwork().getBuses()) {
+            if (bus.hasVoltageControllerCapability() && !bus.isDisabled()) {
+                if (!bus.hasGeneratorsWithSlope()) {
```
geofjamg, line 143: Why do we need to check this? We can never have both a generator controlling reactive power and a SVC with a standby automaton?

src/main/java/com/powsybl/openloadflow/DefaultOuterLoopConfig.java
annetill: I think there is no link with reactive limits as it is just if we have to switch once from a PQ bus to a PV bus. You can do that without any reactive limits. But after switching, the generator can be out of reactive limits...

src/main/java/com/powsybl/openloadflow/ac/MonitoringVoltageOuterLoop.java
annetill: For me, monitoring voltage is more explicit.

src/main/java/com/powsybl/openloadflow/ac/MonitoringVoltageOuterLoop.java
annetill: Yes, you are right because it can become a classical PV -> PQ -> PV bus after.

annetill: Indeed it is done by the reactive limits outerloop so I can delete this.

src/main/java/com/powsybl/openloadflow/DefaultOuterLoopConfig.java
geofjamg: so we should add another parameters to be able to disable this outerloop

src/main/java/com/powsybl/openloadflow/ac/MonitoringVoltageOuterLoop.java
geofjamg: Voltage monitoring is very general

geofjamg:
@annetill should be discard SVC with automaton in reactive outerloop. Imagine a SVC turns to voltage control because reaching a min or max, then it could be processed by normal outerloop like any other generator and switched PQ again?

annetill:
> @annetill should be discard SVC with automaton in reactive outerloop. Imagine a SVC turns to voltage control because reaching a min or max, then it could be processed by normal outerloop like any other generator and switched PQ again?

I is what we want. After becoming a PV bus, it should behave as classical PV bus!

geofjamg:
> > @annetill should be discard SVC with automaton in reactive outerloop. Imagine a SVC turns to voltage control because reaching a min or max, then it could be processed by normal outerloop like any other generator and switched PQ again?
> 
> I is what we want. After becoming a PV bus, it should behave as classical PV bus!

But not going back PQ

annetill:
> > > @annetill should be discard SVC with automaton in reactive outerloop. Imagine a SVC turns to voltage control because reaching a min or max, then it could be processed by normal outerloop like any other generator and switched PQ again?
> > 
> > 
> > I is what we want. After becoming a PV bus, it should behave as classical PV bus!
> 
> But not going back PQ

Yes, it could become PQ again.

geofjamg:
> > > > @annetill should be discard SVC with automaton in reactive outerloop. Imagine a SVC turns to voltage control because reaching a min or max, then it could be processed by normal outerloop like any other generator and switched PQ again?
> > > 
> > > 
> > > I is what we want. After becoming a PV bus, it should behave as classical PV bus!
> > 
> > 
> > But not going back PQ
> 
> Yes, it could become PQ again.

this not like a classical generator where PQ means, reactive limit has been reached.

geofjamg (approved):

PR #637, opened 2022-10-24
Fix network variant removal issue

geofjamg:

No

Bug fix

When we create a LF network from an IIDM network in bus/breaker topo, we create it from a temporary variant. The variant is removed just after and created LF network still wraps IIDM objects from removed variants (especially calculated bus which are no more part of the IIDM network). The only reason for we didn't seen any bug is that we don't update the IIDM network at the end. 

IIDM temporary variant is removed at the end of the calculation when we don't need it anymore.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
@@ -1,0 +1,22 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.openloadflow.network.impl;
+
+import com.powsybl.openloadflow.network.LfNetwork;
+
+import java.util.List;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public interface LfNetworkList extends AutoCloseable {
+
+    List<LfNetwork> getList();
+
+    @Override
+    void close();
+}
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
@@ -6,17 +6,64 @@
  */
 package com.powsybl.openloadflow.network.impl;
 
+import com.powsybl.iidm.network.Network;
 import com.powsybl.openloadflow.network.LfNetwork;
 
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 
 /**
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-public interface LfNetworkList extends AutoCloseable {
+public class LfNetworkList implements AutoCloseable {
 
-    List<LfNetwork> getList();
+    public static class VariantCleaner {
+
+        private final Network network;
+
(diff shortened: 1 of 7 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
```diff
@@ -0,0 +1,22 @@
...
+import com.powsybl.openloadflow.network.LfNetwork;
+
+import java.util.List;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public interface LfNetworkList extends AutoCloseable {
+
+    List<LfNetwork> getList();
```
flo-dup, line 18: we're not so much interested in the list itself, but in the largest element. Shouldn't we add a `Optional<LfNetwork> getLargestNetwork()`? If not we should at least add in the javadoc that the list should be sorted.

src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
```diff
@@ -106,36 +106,74 @@ public static List<LfNetwork> load(Network network, LfNetworkParameters paramete
...
+            return list;
+        }
+
+        @Override
+        public void close() {
+            // nothing to clean
+        }
+    }
+
+    static class LfNetworkListVariantCleanupImpl implements LfNetworkList {
```
flo-dup, line 132: Wouldn't it be clearer/simpler to use if we had only one class? with `tmpVariantId == null` if we didn't clone the working variant, to expose clearly why there's nothing to clean.

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
geofjamg: Ok, added

src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
geofjamg: Variant clean requires 3 attributes (network, old and new tmp variant), it would not be clean to only rely on one of them to decide if we must clean or not. I agree that it would be better to only have one LfNetworkList so I tried another design with just LfNetworkList and an optional embedded object VariantCleaner that is responsible for variant cleaning.

src/main/java/com/powsybl/openloadflow/sa/AcSecurityAnalysis.java
```diff
@@ -91,28 +92,29 @@ SecurityAnalysisReport runSync(String workingVariantId, SecurityAnalysisParamete
...
-            if (largestNetwork.isValid()) {
-                Map<String, LfAction> lfActionsById = createLfActions(largestNetwork, actions);
-                Map<String, List<OperatorStrategy>> operatorStrategiesByContingencyId = indexOperatorStrategiesByContingencyId(propagatedContingencies, operatorStrategies);
-                result = runSimulations(largestNetwork, propagatedContingencies, acParameters, securityAnalysisParameters, operatorStrategiesByContingencyId, lfActionsById, allSwitchesToClose);
-            } else {
+        try (LfNetworkList lfNetworks = Networks.load(network, acParameters.getNetworkParameters(), allSwitchesToOpen, allSwitchesToClose, saReporter)) {
+
+            // run simulation on largest network
+            SecurityAnalysisResult result;
+            LfNetwork largestNetwork = lfNetworks.getLargest().orElse(null);
```
flo-dup, line 99: It makes me sad you didn't use filter and map on the optional!
```java
SecurityAnalysisResult result = lfNetworks.getLargest().filter(LfNetwork::isValid).map(...).orElse(createNoResult());
```

flo-dup (approved):

PR #640, opened 2022-10-26
[WIP] Fix calculated p load in security analysis

geofjamg:

No

Bug fix

No fully sure of the diagnostic: but I think that in a security analysis with a contingency load, distribution p factor for load update are not recalculated taking into account the missing load.

**Other information**:


src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -1178,9 +1178,9 @@ void testSaWithLoadContingency() {
 
         // post-contingency tests
         PostContingencyResult l4ContingencyResult = getPostContingencyResult(result, "l4");
```
annetill, line 1180: For me, there is no bug in main branch. On this case, I have run a power flow disconnecting l4 and the flows on branches are the actual ones and not what you have found. 
When we loose a load on a bus, we change 2 things :
- The bus `loadTargetP` of course ;
- We also change the variable part of loads connected to the bus, removing the part that has been lost. It will change the participating factor through method `getParticipationFactor`.
Why do you think there is a mistake?

PR #641, opened 2022-11-04
Fix non impedant phase shifter with active power flow control activated

geofjamg:

No

Bug fix

If a phase shifter is non impedant or has a low impedance and that regulation is activated, we get an exception

**Other information**:


geofjamg:
@annetill @Hadrien-Godard what is the best way to fix this? I would say non or low impedance phase shifter should not be modelled as perfect non impedant branches as for this kind of branch we cannot calculate the flow during calculation (just as a post processing) and so on we cannot target a specific flow or detect a current violation. So for theses specific pĥase shifters, we should just use a min impedance even if mode is REPLACE_BY_ZERO_IMPEDANCE_LINE.

Hadrien-Godard:
> @annetill @Hadrien-Godard what is the best way to fix this? I would say non or low impedance phase shifter should not be modelled as perfect non impedant branches as for this kind of branch we cannot calculate the flow during calculation (just as a post processing) and so on we cannot target a specific flow or detect a current violation. So for theses specific pĥase shifters, we should just use a min impedance even if mode is REPLACE_BY_ZERO_IMPEDANCE_LINE.

I am ok with proposed solution.
If the non impedant part is limited to only one branch, it may be possible to keep the perfect non impedant branch, as the DUMMY_P variable introduced for the branch represents the active flow on the branch and can be used into the power control.

geofjamg:
Ok for me, but I cannot approve because I am the initial creator of the PR...

annetill (approved):

PR #650, opened 2022-11-14
Test Metrix for violations in Current, Voltage, Active and Apparent Power

caioluke:
ADD: Test Metrix for violations in Current, Voltage, Active and Apparent Power.

Signed-off-by: cluke <caio.luke@artelys.com>

**Other information**:


annetill (changes requested):
After just a quick look, I think that you have to create a dedicated factory in order to ass the limits you are using in your tests. In the same network, you can add current, apparent power and active power limits without any conflict. You have to tried to find coherent values and it is better not to change network characteristics (or to explain why you do that). After that, even in tests, we must avoid code duplication. I am sure you can find a way!

geofjamg:
What is the status of this PR?

annetill:
> What is the status of this PR?

As I said in my comment, this should be done in a dedicated factory in order to be reusable. And the good practice is to find relevant values without any characterictics change.

annetill (approved):

PR #651, opened 2022-11-14
[WIP] Low impedance threshold into OLF parameters

caioluke:
Move `LOW_IMPEDANCE_THRESHOLD`, currently hardcoded, into OLF parameters.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/VoltageMagnitudeInitializer.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/VoltageMagnitudeInitializer.java
@@ -11,6 +11,7 @@ import com.powsybl.commons.PowsyblException;
 import com.powsybl.math.matrix.MatrixFactory;
 import com.powsybl.openloadflow.equations.*;
 import com.powsybl.openloadflow.network.*;
+import com.powsybl.openloadflow.OpenLoadFlowParameters;
 import com.powsybl.openloadflow.network.util.VoltageInitializer;
 import gnu.trove.list.array.TDoubleArrayList;
 import org.slf4j.Logger;
@@ -108,7 +109,7 @@ public class VoltageMagnitudeInitializer implements VoltageInitializer {
                 double r = 0;
                 for (LfBranch neighborBranch : neighborBranches) {
                     PiModel piModel = neighborBranch.getPiModel();
-                    double x = Math.max(Math.abs(piModel.getX()), LfBranch.LOW_IMPEDANCE_THRESHOLD); // to void issue with negative reactances
+                    double x = Math.max(Math.abs(piModel.getX()), OpenLoadFlowParameters.LOW_IMPEDANCE_THRESHOLD); // to void issue with negative reactances
                     b += 1 / x;
                     r += neighborBranch.getBus1() == bus ? 1 / piModel.getR1() : piModel.getR1();
                 }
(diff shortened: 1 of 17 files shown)


caioluke:
I think in some occurencies where I put `OpenLoadFlowParameters.LOW_IMPEDANCE_THRESHOLD_DEFAULT_VALUE` should be something like `OpenLoadFlowParameters.getLowImpedanceThreshold()`.
If that is the case how to implement it properly? (since the line above doesn't do the job)

geofjamg:
> I think in some occurencies where I put `OpenLoadFlowParameters.LOW_IMPEDANCE_THRESHOLD_DEFAULT_VALUE` should be something like `OpenLoadFlowParameters.getLowImpedanceThreshold()`.
> If that is the case how to implement it properly? (since the line above doesn't do the job)

Yes, right you should pass the configured value (which could be different to default one) though the full method call stack.
I would recommend you to store it into `LfNetworkParameter` (see `OpenLoadFlowParameters.getNetworkParameters`) and then pass it to all needed methods.

src/main/java/com/powsybl/openloadflow/ac/VoltageMagnitudeInitializer.java
```diff
@@ -98,6 +98,7 @@ public InitVmBusEquationTerm(LfBus bus, VariableSet<InitVmVariableType> variable
             variables = new ArrayList<>(neighbors.size());
             der = new TDoubleArrayList(neighbors.size());
             double bs = 0; // neighbor branches susceptance sum
+            double lowImpedanceThreshold = new LfNetworkParameters().getLowImpedanceThreshold();
```
annetill, line 101: That is not correct: you have to carry the `LfNetworkParameters`  object until here.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -245,10 +245,11 @@ public double computeApparentPower2() {
 
     @Override
     public boolean isZeroImpedanceBranch(boolean dc) {
+        double lowImpedanceThreshold = new LfNetworkParameters().getLowImpedanceThreshold();
```
annetill, line 248: Same remark: you cannot create new LfNetworkParameters, you have to carry them here.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -288,8 +289,9 @@ public boolean isConnectedAtBothSides() {
 
     @Override
     public void setMinZ(boolean dc) {
-        if (piModel.setMinZ(LOW_IMPEDANCE_THRESHOLD, dc)) {
-            LOGGER.trace("Branch {} has a low impedance, set to min {}", getId(), LOW_IMPEDANCE_THRESHOLD);
+        double lowImpedanceThreshold = new LfNetworkParameters().getLowImpedanceThreshold();
```
annetill, line 292: Same remark.

src/test/java/com/powsybl/openloadflow/network/impl/LfSwitchTest.java
```diff
@@ -68,7 +67,7 @@ void getterTest() {
 
     @Test
     void setterTest() {
-        lfSwitch.getPiModel().setX(LfBranch.LOW_IMPEDANCE_THRESHOLD); //FIXME
+        lfSwitch.getPiModel().setX(OpenLoadFlowParameters.load().getLowImpedanceThreshold());
```
annetill, line 70: Can you add a test where you change the default value with `OpenLoadFlowParameters`? It will show you if all is working or not.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
annetill: Indeed you only have to carry the threshold, but the good one.

PR #660, opened 2022-11-28
[WIP] Support of LoadAction in AC security analysis

annetill:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfAction.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfAction.java
@@ -6,11 +6,13 @@
  */
 package com.powsybl.openloadflow.network;
 
+import com.powsybl.iidm.network.Bus;
+import com.powsybl.iidm.network.Load;
+import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.network.Terminal;
 import com.powsybl.openloadflow.graph.GraphConnectivity;
-import com.powsybl.security.action.Action;
-import com.powsybl.security.action.LineConnectionAction;
-import com.powsybl.security.action.PhaseTapChangerTapPositionAction;
-import com.powsybl.security.action.SwitchAction;
+import com.powsybl.openloadflow.util.PerUnit;
+import com.powsybl.security.action.*;
 
 import java.util.*;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -54,58 +58,90 @@ public boolean isRelative() {
...
         }
     }
 
-    private static Optional<LfAction> create(PhaseTapChangerTapPositionAction action, LfNetwork network) {
-        LfBranch branch = network.getBranchById(action.getTransformerId()); // only two windings transformer for the moment.
+    private static Optional<LfAction> create(LoadAction action, LfNetwork lfNetwork, Network network, boolean breakers) {
+        Load load = network.getLoad(action.getLoadId());
+        Terminal terminal = load.getTerminal();
+        Bus bus = breakers ? terminal.getBusBreakerView().getBus() : terminal.getBusView().getBus();
+        if (bus != null) {
```
annetill, line 96: For me: I think I have to check that the load is not in contingency. Add a unit test!

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -387,13 +384,13 @@ void testHighVoltageLimitViolations() {
...
         Network network = createNodeBreakerNetwork();
         network.getLine("L1").newActivePowerLimits1()
-               .setPermanentLimit(1.0)
-               .beginTemporaryLimit()
-               .setName("60")
-               .setAcceptableDuration(60)
-               .setValue(1.2)
-               .endTemporaryLimit()
-               .add();
+                .setPermanentLimit(1.0)
```
annetill, line 387: @geofjamg I have an issue with indentation... and I don't know how to revert this. I can dot it by hand if no other solution...

src/main/java/com/powsybl/openloadflow/network/LfAction.java
Hadrien-Godard: Maybe test it in the apply method?

Hadrien-Godard (approved):

src/main/java/com/powsybl/openloadflow/network/impl/LfAggregatedLoadsImpl.java
```diff
@@ -112,6 +114,26 @@ public double getLoadTargetQ(double diffLoadTargetP) {
...
+
+    @Override
+    public Map<String, Boolean> getLoadsStatus() {
+        return loadsStatus;
+    }
+
+    @Override
+    public void setLoadsStatus(Map<String, Boolean> loadsStatus) {
+        this.loadsStatus = loadsStatus;
+    }
```
jeandemanged, line 135: my understanding is that these get/set methods are used only for the purpose of saving and restoring a BusDcState. In order to avoid the map to be manipulated outside this class in the future, maybe we should:
- for the getter: return an immutable copy of the map
- for the setter: make a copy of the input map

?

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
```diff
@@ -38,12 +39,14 @@ public class LfContingency {
 
     private final Map<LfBus, PowerShift> busesLoadShift;
 
+    private final Set<String> originalPowerShiftIds;
```
geofjamg, line 42: At first reading it was really not clear to me that this is load IDs.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -54,58 +58,93 @@ public boolean isRelative() {
...
+            LfBus lfBus = lfNetwork.getBusById(bus.getId());
+            if (!lfBus.getAggregatedLoads().isDisabled(action.getLoadId())) {
```
geofjamg, line 98: Are you sure we have to check this here?

src/main/java/com/powsybl/openloadflow/network/BusDcState.java
```diff
@@ -18,13 +18,15 @@ public class BusDcState extends ElementState<LfBus> {
     private final Map<String, Double> generatorsTargetP;
     private final Map<String, Boolean> participatingGenerators;
     private final double absVariableLoadTargetP;
+    private final Map<String, Boolean> loadsStatus;
```
geofjamg, line 21: This is load disabling status? Why not loadsDisable ?

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -54,58 +58,93 @@ public boolean isRelative() {
...
                 throw new UnsupportedOperationException("Unsupported action type: " + action.getType());
         }
     }
 
-    private static Optional<LfAction> create(PhaseTapChangerTapPositionAction action, LfNetwork network) {
-        LfBranch branch = network.getBranchById(action.getTransformerId()); // only two windings transformer for the moment.
+    private static Optional<LfAction> create(LoadAction action, LfNetwork lfNetwork, Network network, boolean breakers) {
+        Load load = network.getLoad(action.getLoadId());
+        Terminal terminal = load.getTerminal();
+        Bus bus = breakers ? terminal.getBusBreakerView().getBus() : terminal.getBusView().getBus();
```
geofjamg, line 95: There is a utility method for that in Networks

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -54,58 +58,93 @@ public boolean isRelative() {
 
     private final TapPositionChange tapPositionChange;
 
-    private LfAction(String id, LfBranch disabledBranch, LfBranch enabledBranch, TapPositionChange tapPositionChange) {
+    private final Map<Pair<LfBus, String>, Pair<Double, PowerShift>> busesLoadShift; // key: bus and loadId.
```
geofjamg, line 61: That would probably worth a dedicated objet, this are too many Pair which are difficult to read. You don't really need to map as you never search into it.

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/network/LfAction.java
annetill: Yes, I just get what you mean. Yes, just test that in the apply method.

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
annetill: It is not only load ids but also the other equipments converted as power shift as LCC converter stations.

geofjamg (approved):

PR #662, opened 2022-11-29
Support of operator strategies in DC security analysis

obrix:

Support of operator strategies in the DC security analysis

Operator strategies are ignored in DC mode

Operator strategies are correctly handled

**Other information**:


src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -139,7 +137,7 @@ private static Pair<Boolean, double[]> run(LfNetwork network, DcLoadFlowParamete
             distributeSlack(remainingBuses, parameters.getBalanceType());
         }
 
-        var targetVectorArray = targetVector.getArray();
+        var targetVectorArray = targetVector.getArray().clone();
```
geofjamg, line 140: @annetill this fix is the good one for me, because indeed we need to work on a copy of target vector, because first it could be hacked with disabled element and also because it is updated by the LU decomposition result and so on if we want to keep targets for later use we need to copy.

geofjamg: maybe just add an explanation

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -62,16 +62,8 @@ private static double getActivePowerMismatch(Collection<LfBus> buses) {
 
     @Override
     public DcLoadFlowResult run() {
-        EquationSystem<DcVariableType, DcEquationType> equationSystem = context.getEquationSystem();
-
-        boolean succeeded = false;
-        try (JacobianMatrix<DcVariableType, DcEquationType> j = context.getJacobianMatrix()) {
```
annetill, line 68: @geofjamg and @obrix I have been obliged to delete the exception catching, but I am still not sure that it is the good fix. What do you think?

src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystem.java
```diff
@@ -143,14 +143,18 @@ private static void createBranches(LfNetwork network, EquationSystem<DcVariableT
...
-                                                                        double lowImpedanceThreshold) {
+                                                                        double lowImpedanceThreshold, boolean withListener) {
         EquationSystem<DcVariableType, DcEquationType> equationSystem = new EquationSystem<>(creationParameters.isIndexTerms());
 
         createBuses(network, equationSystem);
         createBranches(network, equationSystem, creationParameters, lowImpedanceThreshold);
 
         EquationSystemPostProcessor.findAll().forEach(pp -> pp.onCreate(equationSystem));
 
+        if (withListener) {
```
annetill, line 154: The listener is optional and cannot be used easily in DC sensitivity analysis.

src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystemCreationParameters.java
```diff
@@ -39,6 +39,10 @@ public boolean isIndexTerms() {
         return indexTerms;
     }
 
+    public void setIndexTerms(boolean indexTerms) {
```
annetill, line 42: @geofjamg I am not sure I have to change this parameter this way. Maybe in create dc parameters ?

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -193,13 +193,18 @@ public void updateConnectivity(GraphConnectivity<LfBus, LfBranch> connectivity)
...
 
-    public void apply() {
+    public void apply(LfNetwork network, boolean dc) {
         if (tapPositionChange != null) {
             LfBranch branch = tapPositionChange.getBranch();
             int tapPosition = branch.getPiModel().getTapPosition();
             int value = tapPositionChange.getValue();
             int newTapPosition = tapPositionChange.isRelative() ? tapPosition + value : value;
             branch.getPiModel().setTapPosition(newTapPosition);
+            if (dc) { // FIXME
```
annetill, line 203: @geofjamg I think that it is not the good fix, maybe it will be better in the setTapPosition but I prefer to ask for our opinion before. We need to update the target vector because right hand side has changed.

src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystemUpdater.java
```diff
@@ -33,19 +33,7 @@ private void updateElementEquations(LfElement element, boolean enable) {
...
-                }
-            }
-
-            // update all equation terms related to the element
-            for (var equationTerm : equationSystem.getEquationTerms(element.getType(), element.getNum())) {
-                if (equationTerm.isActive() != enable) {
-                    equationTerm.setActive(enable);
-                }
-            }
+            nonBranchEquationsUpdate(element, enable);
```
annetill, line 36: The first part is dedicated to updating zero impedance branch equations. The second part is for the rest, included impedant branch equations.

obrix: Yes I figured the name would be problematic, I was interested in factorization for ci, do you have a suggestion ?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -241,4 +243,57 @@ protected static void findAllSwitchesToOperate(Network network, List<Action> act
...
     }
+
+    protected Optional<OperatorStrategyResult> runActionSimulation(LfNetwork network, LoadFlowContext context, OperatorStrategy operatorStrategy,
+                                                                 LimitViolationManager preContingencyLimitViolationManager,
+                                                                 SecurityAnalysisParameters.IncreasedViolationsParameters violationsParameters,
+                                                                 LimitViolationsResult postContingencyLimitViolations, Map<String, LfAction> lfActionById,
+                                                                 boolean createResultExtension, LfContingency contingency, boolean dc) {
+        OperatorStrategyResult operatorStrategyResult = null;
+
+        if (checkCondition(operatorStrategy, postContingencyLimitViolations)) {
```
annetill, line 254: The condition is checked twice for DC security analysis, no?

annetill: For sure you cannot apply the contingency if not relevant and restore the network. It is a lost of performance and it increases the probability of errors. The mutualisation cannot be done despide that points.

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
geofjamg: This is ok

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
geofjamg: Done

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
geofjamg: I refactor the code to only check violation one time in DC SA and keep as much code as possible in common. It should be ok now.

src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystemCreationParameters.java
geofjamg: Yes, moved to createDcParameters

src/main/java/com/powsybl/openloadflow/network/LfAction.java
geofjamg: Indeed right place is in setTapPosition.
Done.

annetill (approved):

geofjamg (approved):

PR #663, opened 2022-11-29
Fix DcEquationSystem with disabled non impedant branch

annetill:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystem.java
+++ b/src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystem.java
@@ -51,19 +51,24 @@ public final class DcEquationSystem {
             equationSystem.createEquation(branch.getNum(), DcEquationType.ZERO_PHI)
                     .addTerm(equationSystem.getVariable(bus1.getNum(), DcVariableType.BUS_PHI).createTerm())
                     .addTerm(equationSystem.getVariable(bus2.getNum(), DcVariableType.BUS_PHI).<DcEquationType>createTerm()
-                                         .minus());
+                                         .minus()).setActive(!branch.isDisabled());
 
             // add a dummy active power variable to both sides of the non impedant branch and with an opposite sign
             // to ensure we have the same number of equation and variables
+            var dummyP = equationSystem.getVariable(branch.getNum(), DcVariableType.DUMMY_P);
             equationSystem.getEquation(bus1.getNum(), DcEquationType.BUS_TARGET_P)
                     .orElseThrow()
-                    .addTerm(equationSystem.getVariable(branch.getNum(), DcVariableType.DUMMY_P)
-                            .createTerm());
+                    .addTerm(dummyP.createTerm());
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystem.java
```diff
@@ -51,19 +51,24 @@ public static void createNonImpedantBranch(EquationSystem<DcVariableType, DcEqua
             equationSystem.createEquation(branch.getNum(), DcEquationType.ZERO_PHI)
                     .addTerm(equationSystem.getVariable(bus1.getNum(), DcVariableType.BUS_PHI).createTerm())
                     .addTerm(equationSystem.getVariable(bus2.getNum(), DcVariableType.BUS_PHI).<DcEquationType>createTerm()
-                                         .minus());
+                                         .minus()).setActive(!branch.isDisabled());
```
annetill, line 54: I really don't understand why but this first ``setActive`` is not needed in AcEquationSystem. So I don't know if this fix is the good one.

geofjamg: EquationSystem.createEquation has 2 overloads:
 - initial one just take an element num
 - newest one take directly the LfElement and so on is able to internally disable equation according to LfElement status

DcEquationSystem had unfortunately not been migrated to new one

src/test/java/com/powsybl/openloadflow/ac/NonImpedantBranchDisablingTest.java
```diff
@@ -78,4 +87,41 @@ void testOpenBranch() {
...
+                                                            true,
+                                                            false);
+        try (LfNetworkList lfNetworks = Networks.load(network, acLoadFlowParameters.getNetworkParameters(), Collections.emptySet(), Set.of(c1), Reporter.NO_OP)) {
+            LfNetwork largestNetwork = lfNetworks.getLargest().orElseThrow();
+            largestNetwork.getBranchById("C1").setDisabled(true);
+            try (AcLoadFlowContext context = new AcLoadFlowContext(largestNetwork, acLoadFlowParameters)) {
+                new AcloadFlowEngine(context).run();
+            }
+            // should be the same as with previous LF
+            assertEquals(401.740, largestNetwork.getBranchById("L1").getP1().eval() * PerUnit.SB, LoadFlowAssert.DELTA_POWER);
```
geofjamg, line 122: @annetill why we don't get exactly the value we had with the initial LF: 401.757 vs 401.740 ?

annetill: Maybe because in the first case, we have a system resolution based on the bus/view and in the second case, we are based on the bus/breaker view.

geofjamg: Yes right, this is not exactly the same system, so make sense to not have numerically the same solution

geofjamg (approved):

PR #666, opened 2022-12-07
Zero impedance branch management refactoring

annetill:

No.

The status of zero impedance branch is defined once, when the LfNetwork is created, in a AC context or in a DC context. So:
- if during the simulation the tap position changes, it does not change this status. Which is not so bad indeed because we don't change the branch equation from an impedant branch equation to a zero impedance one or inversely.
- In case of an initialisation relying on DC loadflow, the zero impedance status is defined in an AC context.    

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfBranch.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfBranch.java
@@ -139,13 +139,11 @@ public interface LfBranch extends LfElement {
 
     double computeApparentPower2();
 
-    boolean isZeroImpedanceBranch();
+    boolean isZeroImpedance();
 
     void setSpanningTreeEdge(boolean spanningTreeEdge);
 
     boolean isSpanningTreeEdge();
-
-    boolean isZeroImpedanceBranchWithEquation();
 
     Evaluable getA1();
 
--- a/src/main/java/com/powsybl/openloadflow/network/LfBranch.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfBranch.java
@@ -139,11 +139,13 @@ public interface LfBranch extends LfElement {
 
     double computeApparentPower2();
 
-    boolean isZeroImpedanceBranch(boolean dc, double lowImpedanceThreshold);
+    boolean isZeroImpedanceBranch();
 
     void setSpanningTreeEdge(boolean spanningTreeEdge);
 
     boolean isSpanningTreeEdge();
+
+    boolean isZeroImpedanceBranchWithEquation();
 
     Evaluable getA1();
 
--- a/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
@@ -49,12 +49,19 @@ public abstract class AbstractLfBranch extends AbstractElement implements LfBran
 
     private ReactivePowerControl reactivePowerControl;
 
-    protected AbstractLfBranch(LfNetwork network, LfBus bus1, LfBus bus2, PiModel piModel) {
+    protected boolean zeroImpedanceBranch;
+
+    protected AbstractLfBranch(LfNetwork network, LfBus bus1, LfBus bus2, PiModel piModel, boolean dc, double lowImpedanceThreshold) {
         super(network);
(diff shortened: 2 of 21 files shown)


src/main/java/com/powsybl/openloadflow/network/LfBranch.java
```diff
@@ -139,12 +139,14 @@ default List<LfLimit> getLimits2(LimitType type) {
...
     double computeApparentPower2();
 
-    boolean isZeroImpedanceBranch(boolean dc, double lowImpedanceThreshold);
+    boolean isZeroImpedanceBranch();
 
     void setSpanningTreeEdge(boolean spanningTreeEdge);
 
     boolean isSpanningTreeEdge();
 
+    boolean isZeroImpedanceBranchWithEquation();
```
geofjamg, line 148: I would remove this method and keep composite check isZeroImpedanceBranch() && isSpanningTreeEdge()

src/main/java/com/powsybl/openloadflow/network/LfBranch.java
```diff
@@ -139,12 +139,14 @@ default List<LfLimit> getLimits2(LimitType type) {
 
     double computeApparentPower2();
 
-    boolean isZeroImpedanceBranch(boolean dc, double lowImpedanceThreshold);
+    boolean isZeroImpedanceBranch();
```
geofjamg, line 142: Just isZeroImpedance() ?

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -49,12 +49,19 @@ public abstract class AbstractLfBranch extends AbstractElement implements LfBran
...
-    protected AbstractLfBranch(LfNetwork network, LfBus bus1, LfBus bus2, PiModel piModel) {
+    protected boolean zeroImpedanceBranch;
+
+    protected AbstractLfBranch(LfNetwork network, LfBus bus1, LfBus bus2, PiModel piModel, boolean dc, double lowImpedanceThreshold) {
         super(network);
         this.bus1 = bus1;
         this.bus2 = bus2;
         this.piModel = Objects.requireNonNull(piModel);
         this.piModel.setBranch(this);
+        if (dc) {
```
geofjamg, line 60: Let's keep a static function for that called to init this constructor?

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -244,12 +251,8 @@ public double computeApparentPower2() {
     }
 
     @Override
-    public boolean isZeroImpedanceBranch(boolean dc, double lowImpedanceThreshold) {
```
geofjamg, line 247: Keep it but static, called in the constructor

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -262,6 +265,11 @@ public boolean isSpanningTreeEdge() {
         return this.spanningTreeEdge;
     }
 
+    @Override
+    public boolean isZeroImpedanceBranchWithEquation() {
```
geofjamg, line 269: There is no really added value for this method

geofjamg (changes requested):
A few minor change to do, but this is a very good idea and it simplifies a lot the code.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -291,5 +299,6 @@ public void setMinZ(boolean dc, double lowImpedanceThreshold) {
         if (piModel.setMinZ(lowImpedanceThreshold, dc)) {
             LOGGER.trace("Branch {} has a low impedance, set to min {}", getId(), lowImpedanceThreshold);
         }
+        this.zeroImpedanceBranch = false;
```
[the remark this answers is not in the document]
annetill: I think I need it, because this is done after the branch has been conctructed, so we have to override the boolean that defines if it is a zeroi impedance branch or not.

geofjamg (approved):

PR #680, opened 2022-12-22
[WIP] Better nominal voltage per uniting

geofjamg:

No

Feature

We recently had some real cases where some branches has a slightly different nominal voltage at both side. In such a case OLF add a voltage ratio to the branch to "fix" the per-uniting. the problem is that if those lines have a very small impedance, the resulting case become "ill conditioned" and a large amount of reactive power flows in that branch causing divergence of the NR. 

To avoid this phenomena, we reduce the number of nominal voltages that will be used to per unit. All nominal voltages that are closed enough (10%) are "merged" resulting to a smaller number of nominal voltage and no more small ratio added to fix the per uniting. The real cases now easily converge.

**Other information**:


geofjamg:
Still to do:
 - merge threshold configuration in OLF specific parameters
 - create `NominalVoltageMapping` instance outside of `LfNetworkLoader`(to avoid multiple create during a single calculation)

geofjamg:
> Still to do:
> 
> * merge threshold configuration in OLF specific parameters
> * create `NominalVoltageMapping` instance outside of `LfNetworkLoader`(to avoid multiple create during a single calculation)

Done

src/main/java/com/powsybl/openloadflow/network/NominalVoltageMapping.java
```diff
@@ -0,0 +1,98 @@
...
+    private static Map<Double, MutableInt> createHistogram(Network network) {
+        Map<Double, MutableInt> nominalVoltageHistogram = new TreeMap<>();
+        for (var vl : network.getVoltageLevels()) {
+            nominalVoltageHistogram.computeIfAbsent(vl.getNominalV(), k -> new MutableInt())
+                    .increment();
+        }
+        return nominalVoltageHistogram;
+    }
+
+    public static NominalVoltageMapping create(Network network, double resolution) {
```
sylvlecl, line 63: The only thing I wonder is if we could need different "strategies" to decide on the mapping:
for example should we take into account connectivity, so that a nominal voltage in Portugal does not affect the per-unit in Germany.
In that case we could make it an interface with other impls in the future. But this can be done when it will really be needed, in any case.

sylvlecl (commented):
No formal approval because tests are still to be implemented, but the feature looks good to me :)

src/main/java/com/powsybl/openloadflow/network/NominalVoltageMapping.java
geofjamg: I tried to do that first, but the problem I had is that I need to map new nominal voltages on bus ids (we cannot associate it to voltage level id) and for branches connected to only one side, if side 2 is open I cannot get nominal voltage to per unit  the branch. Any idea if there is a simple way to overcome this?

sylvlecl: Indeed I didn't think of those "details", it gets tricky ... even more if the topology changes during computation, so it does not seem such a good idea after all !

geofjamg:
> No formal approval because tests are still to be implemented, but the feature looks good to me :)

Unit test added

src/main/java/com/powsybl/openloadflow/network/NominalVoltageMapping.java
geofjamg: I refactored to have a generic  interface and a simple implementation so that if one day we want to have a more complex implementations, we can dot it without changing to much code

annetill (changes requested):
I have performed some tests on real networks without any issue. But I think we have an issue with nominal voltage of star bus that is still rated U0.

geofjamg:
> I have performed some tests on real networks without any issue. But I think we have an issue with nominal voltage of star bus that is still rated U0.

I'm not sure we have to map this nominal voltage as it is not a real one (one coming from network) but just one set by convention equals to rated voltage of the middle point.

geofjamg:
In DC mode, nominal is not only for per unit but is also the bus voltage used to calculate flows. So I think this PR break DC mode. We have to think how to deal with this.

geofjamg:
@annetill I suspect we have an issue with an AC LF and a DC init because in that case we could aggregate nominal voltages and result of won't be correct?

geofjamg:
Was it really a good idea?

geofjamg:
The issue this PR try to solve could also be solved by a voltage init mode to VOLTAGE_MAGNITUDE. I will try to detect and automatically switch to this mode if not selected in another PR.

PR #683, opened 2022-12-28
DcValueVoltageInitializer fails in presence of resistive only branches

jeandemanged:
Signed-off-by: Damien Jeandemange <damien.jeandemange@artelys.com>

No

Bug fix

Exception thrown by DcValueVoltageInitializer if the newtwork contains a resistive only branch.

TBD

No

**Other information**:

```
java.lang.IllegalArgumentException: Branch 'l12' has reactance equal to zero

	at com.powsybl.openloadflow.dc.equations.AbstractClosedBranchDcFlowEquationTerm.<init>(AbstractClosedBranchDcFlowEquationTerm.java:42)
	at com.powsybl.openloadflow.dc.equations.ClosedBranchSide1DcFlowEquationTerm.<init>(ClosedBranchSide1DcFlowEquationTerm.java:24)
	at com.powsybl.openloadflow.dc.equations.ClosedBranchSide1DcFlowEquationTerm.create(ClosedBranchSide1DcFlowEquationTerm.java:33)
	at com.powsybl.openloadflow.dc.equations.DcEquationSystem.createImpedantBranch(DcEquationSystem.java:75)
	at com.powsybl.openloadflow.dc.equations.DcEquationSystem.createBranches(DcEquationSystem.java:116)
	at com.powsybl.openloadflow.dc.equations.DcEquationSystem.create(DcEquationSystem.java:126)
	at com.powsybl.openloadflow.dc.DcLoadFlowContext.getEquationSystem(DcLoadFlowContext.java:32)
	at com.powsybl.openloadflow.dc.DcLoadFlowEngine.run(DcLoadFlowEngine.java:65)
	at com.powsybl.openloadflow.dc.DcValueVoltageInitializer.prepare(DcValueVoltageInitializer.java:58)
	at com.powsybl.openloadflow.ac.DcValueVoltageInitializerTest.testFourBusNetworkResistiveOnlyBranch(DcValueVoltageInitializerTest.java:92)
``

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/ac/DcValueVoltageInitializerTest.java
+++ b/src/test/java/com/powsybl/openloadflow/ac/DcValueVoltageInitializerTest.java
@@ -1,0 +1,99 @@
+/**
+ * Copyright (c) 2022, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.openloadflow.ac;
+
+import com.powsybl.iidm.network.Network;
+import com.powsybl.loadflow.LoadFlowParameters;
+import com.powsybl.math.matrix.DenseMatrixFactory;
+import com.powsybl.math.matrix.MatrixFactory;
+import com.powsybl.openloadflow.dc.DcValueVoltageInitializer;
+import com.powsybl.openloadflow.network.*;
+import com.powsybl.openloadflow.network.impl.LfNetworkLoaderImpl;
+import com.powsybl.openloadflow.network.util.VoltageInitializer;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+/**
+ * @author Damien Jeandemange <damien.jeandemange at artelys.com>
+ */
+class DcValueVoltageInitializerTest {
+
+    private Network network;
+    private SlackBusSelector slackBusSelector;
+    private MatrixFactory matrixFactory;
+    private LfNetworkParameters lfNetworkParameters;
+
+    private static void assertBusVoltage(LfNetwork network, VoltageInitializer initializer, String busId, double angleRef) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by geofjamg** [comment]
In fact, in addition to the regression from #666, there is also several issues that were already before:
 - spanning tree which is based in zero impedance criteria was only store for current dc mode, it should be available for both dc and ac mode in case of init DC + LF AC in order to have a consistent spanning tree with zero impedance booleans.
 - non impedance branches graph and spanning tree should be cached as created several times (especially at the end for flow calculation on zero impedance branches)

So I tried in this PR to store all zero impedance infos for both AC and DC and as much as possible set everything in a lazy way to only calculate what is needed.

File: src/main/java/com/powsybl/openloadflow/network/LfNetwork.java [hunk 1]
```diff
@@ -447,10 +465,10 @@ public void reportBalance(Reporter reporter) {
             this, activeGeneration, activeLoad, reactiveGeneration, reactiveLoad);
     }
 
-    public void fix(boolean minImpedance, boolean dc, double lowImpedanceThreshold) {
+    public void fix(boolean minImpedance, double lowImpedanceThreshold) {
         if (minImpedance) {
             for (LfBranch branch : branches) {
-                branch.setMinZ(dc, lowImpedanceThreshold);
```
geofjamg, line 453: I think we should not do this here at a post loading fix. z cut could be done in a cleaner way at branch creation.

geofjamg:
Instead of the boolean dc which is not very readable we could have an enum like ?

enum EquationMode {
    AC,
    DC_APPROX
}

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -75,6 +77,14 @@ public class LfNetwork extends AbstractPropertyBag implements PropertyBag {
 
     private GraphConnectivity<LfBus, LfBranch> connectivity;
 
+    private Graph<LfBus, LfBranch> dcZeroImpedanceSubGraph;
```
geofjamg, line 80: Maybe zeroImpedanceSubGraph and  zeroImpedanceSpanningTree could be grouped in another object.

annetill: Yes, it could be a good idea.

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
annetill: It has been designed like that because min z what dependent of the AC or DC context. Now, as we store both, we can do that at branch creation. Do you want me to do this?

annetill (approved):

PR #693, opened 2023-01-05
Fix incremental transformer voltage control loop

Hadrien-Godard:

**Other information**:


src/main/java/com/powsybl/openloadflow/network/PiModelArray.java
```diff
@@ -200,7 +200,7 @@ private Range<Integer> getAllowedPositionRange(AllowedDirection allowedDirection
             case INCREASE:
                 return Range.between(tapPosition - lowTapPosition, models.size());
             case DECREASE:
-                return Range.between(0, tapPosition - lowTapPosition);
+                return Range.between(0, tapPosition - lowTapPosition + 1);
```
annetill, line 203: I agree with this +1 as the maximum of the range is not included in our code. It could be nice to add a unit test that shows the bug.

src/main/java/com/powsybl/openloadflow/network/PiModelArray.java
```diff
@@ -218,9 +218,10 @@ public Optional<Direction> updateTapPositionR1(double deltaR1, int maxTapShift,
         int oldTapPosition = tapPosition;
         // find tap position with the closest r1 value without exceeding the maximum of taps to switch.
         double smallestDistance = Math.abs(deltaR1);
-        for (int p = positionRange.getMinimum();
-             p < positionRange.getMaximum() && Math.abs(lowTapPosition + p - oldTapPosition) <= maxTapShift;
-             p++) {
+        for (int p = positionRange.getMinimum(); p < positionRange.getMaximum(); p++) {
```
annetill, line 221: I am not sure to get the difference here. Can you explain me with words?

Hadrien-Godard: I think that in the first case, if (Math.abs(lowTapPosition + p - oldTapPosition) <= maxTapShift) is false, then the loop is stopped. In the second case, the loop goes to next iterate.

In case p goes from 0 to (oldTapPosition - lowTapPosition) and lowTapPosition and oldTapPosition are distant by more than maxTapShift, then the loop starts at p=0 and directly stops (in the first case) because the condition (Math.abs(lowTapPosition + p - oldTapPosition) <= maxTapShift) is false.

annetill: Yes, agreed.

src/main/java/com/powsybl/openloadflow/network/PiModelArray.java
annetill: @geofjamg I am not able to add a unit test that shows that this is a bug. Can you just confirm us that this +1 is needed. The PR will be okay for me after your confirmation.

annetill (approved):

geofjamg (approved):

annetill (approved):

PR #699, opened 2023-01-10
Newton raphson new stopping criteria

lejeana:

feature: add a new Newton-Raphson convergence criteria for OPLF.

Only one Newton-Raphson convergence criteria that is hardcoded using mismatch vector norm

A new Newton-Raphson convergence criteria that different value based on the unit for Voltage, Active Power, Reactive Power.


src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -57,6 +55,15 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
      */
     public static final double SLACK_BUS_P_MAX_MISMATCH_DEFAULT_VALUE = 1.0;
 
+    // FIXME What is the value?
+    public static final double MAX_ACTIVE_POWER_MISMATCH = 1.0;
```
obrix, line 59: Can you name it with _DEFAULT_VALUE like the other parameters ?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -715,9 +728,28 @@ public static AcLoadFlowParameters createAcParameters(Network network, LoadFlowP
...
+    private static NewtonRaphsonStoppingCriteria generateNewtonRaphsonStoppingCriteria(OpenLoadFlowParameters parametersExt,
+                                                                                       NewtonRaphsonStoppingCriteriaType typeCriteria) {
+        switch (typeCriteria) {
+            case DEFAULT_CRITERIA:
+                return new DefaultNewtonRaphsonStoppingCriteria(parametersExt.getNewtonRaphsonConvEpsPerEq());
+            case CUSTOM_CRITERIA:
+                return new CustomNewtonRaphsonStoppingCriteria(parametersExt.getNewtonRaphsonConvEpsPerEq(),
+                        parametersExt.maxActivePowerMismatch, parametersExt.maxReactivePowerMismatch, parametersExt.maxVoltageMismatch);
+            default:
+                return null;
```
obrix, line 740: Throw an exception here rather than returning null. You can look at other switches to see what kind of exception you can throw in similar cases.

src/main/java/com/powsybl/openloadflow/ac/nr/CustomNewtonRaphsonStoppingCriteria.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+package com.powsybl.openloadflow.ac.nr;
+
+import com.powsybl.openloadflow.ac.equations.AcEquationType;
+import com.powsybl.openloadflow.ac.equations.AcVariableType;
+import com.powsybl.openloadflow.equations.EquationSystem;
+import com.powsybl.openloadflow.equations.Vectors;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
```
obrix, line 15: Author to fix.

src/main/java/com/powsybl/openloadflow/ac/nr/CustomNewtonRaphsonStoppingCriteria.java
```diff
@@ -0,0 +1,85 @@
+/**
+ * Copyright (c) 2019, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
```
obrix, line 6: The copyright header should have the licence identifier "SPDX-License-Identifier: MPL-2.0" as the last line.

src/main/java/com/powsybl/openloadflow/ac/nr/CustomNewtonRaphsonStoppingCriteria.java
```diff
@@ -0,0 +1,85 @@
...
+                    break;
+                case BUS_TARGET_V:
+                case BUS_TARGET_V_WITH_SLOPE:
+                case ZERO_V:
+                    if (fx[idx] >= maxVoltageMismatch) {
+                        return false;
+                    }
+                    break;
+                default:
+                    //FIXME how to deal default case?
```
obrix, line 74: You can also throw an exception for unsupported type here.

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphsonStoppingCriteriaType.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2019, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.openloadflow.ac.nr;
+
+public enum NewtonRaphsonStoppingCriteriaType {
+    DEFAULT_CRITERIA,
+    CUSTOM_CRITERIA
```
obrix, line 11: We have to find a proper name for those new stopping criteria, maybe @Hadrien-Godard has an idea ?

src/main/java/com/powsybl/openloadflow/ac/nr/CustomNewtonRaphsonStoppingCriteria.java
```diff
@@ -0,0 +1,85 @@
...
+public class CustomNewtonRaphsonStoppingCriteria implements NewtonRaphsonStoppingCriteria {
+
+    /**
+     * Convergence epsilon per equation: 10^-4 in p.u => 10^-2 in Kv, Mw or MVar
+     */
+    //FIXME what to do with this value?
+    public static final double DEFAULT_CONV_EPS_PER_EQ = Math.pow(10, -4);
+
+    //FIXME what to do with this value?
+    private final double convEpsPerEq;
```
obrix, line 26: You can probably drop it.

obrix (changes requested):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -173,6 +180,12 @@ public enum LowImpedanceBranchMode {
 
     private double plausibleActivePowerLimit = LfNetworkParameters.PLAUSIBLE_ACTIVE_POWER_LIMIT_DEFAULT_VALUE;
 
+    private double maxActivePowerMismatch = MAX_ACTIVE_POWER_MISMATCH;
```
obrix, line 183: NewtonRaphsonStoppingCriteriaType is also a new parameter here, for those 4 new parameters you have to create getters / setters like for all the other parameters to integrate them properly into OpenLoadFlowParameters.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
Hadrien-Godard, line 59: 10^-4 seems ok here.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -57,6 +55,15 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
      */
     public static final double SLACK_BUS_P_MAX_MISMATCH_DEFAULT_VALUE = 1.0;
 
+    // FIXME What is the value?
+    public static final double MAX_ACTIVE_POWER_MISMATCH = 1.0;
+
+    // FIXME What is the value?
+    public static final double MAX_REACTIVE_POWER_MISMATCH = 1.0;
```
Hadrien-Godard, line 62: 10^-2 seems ok here. Please add in a comment that the value is in Mvar

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -57,6 +55,15 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
...
+
+    // FIXME What is the value?
+    public static final double MAX_VOLTAGE_MISMATCH = 1.0;
```
Hadrien-Godard, line 65: 10^-2 seems ok here. Please add in a comment that the value is in kV here.

src/main/java/com/powsybl/openloadflow/ac/nr/CustomNewtonRaphsonStoppingCriteria.java
Hadrien-Godard: Yes you can drop it.

src/main/java/com/powsybl/openloadflow/ac/nr/CustomNewtonRaphsonStoppingCriteria.java
Hadrien-Godard: No, we have to deal with other cases. Therefore, there should be a parameter for other equation types (one for angle, one for unitless quantities, etc.). If we do not want to expose these parameters to the user then they should be hardcoded as it was done before for convEpsPerEq. You can take 10^-5 as value for the moment I guess.

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphsonStoppingCriteriaType.java
Hadrien-Godard: DEFAULT_CRITERIA could be named UNIFORM_CRITERIA (and should be let as default type).
CUSTOM_CRITERIA could be named PER_EQUATION_TYPE_CRITERIA.

src/main/java/com/powsybl/openloadflow/ac/nr/CustomNewtonRaphsonStoppingCriteria.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private boolean computeStop(double[] fx, EquationSystem<AcVariableType, AcEquationType> equationSystem) {
+        for (var eq : equationSystem.getIndex().getSortedEquationsToSolve()) {
+            var type = eq.getType();
+            var idx = eq.getColumn();
+            switch (type) {
+                case BRANCH_TARGET_P:
+                case BUS_TARGET_P:
+                case DUMMY_TARGET_P:
+                    if (fx[idx] >= maxActivePowerMismatch) {
```
Hadrien-Godard, line 54: For each test, you need to compute abs(fx[idx]) since fx contains the difference and not the absolute difference (thus can be negative).
Moreover, there is another difficult part here as the equations have been normalized in per unit, you need to recover the mismatch in physical units to compare to the criteria.
For power quantities this is easy because each equation is normalized by PerUnit.SB, see the PerUnit class.
However, for voltage quantities, they have been normalized according to "Nominal Voltage" at the corresponding bus. So you need to get the original bus whose equation is referring to.

AnkurArohi:
I do not understand what exactly is here the criterion

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -57,6 +55,15 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
      */
     public static final double SLACK_BUS_P_MAX_MISMATCH_DEFAULT_VALUE = 1.0;
 
+    /** Default value of the maximum active power mismatch in MW **/
+    public static final double MAX_ACTIVE_POWER_MISMATCH_DEFAULT_VALUE = Math.pow(10, -2);
+
+    /** Default value of the maximum reactive power mismatch in Mvar **/
+    public static final double MAX_REACTIVE_POWER_MISMATCH_DEFAULT_VALUE = Math.pow(10, -2);
```
AnkurArohi, line 62: I find the default value a bit large as compared to other softwares , I was thinking something like 10-5 would make more sense

Hadrien-Godard:
> I do not understand what exactly is here the criterion

Hi Ankur, in current main branch, the criterion is to have a norm 2 difference between target vector and equations evaluation under a threshold (10^-4) by default.
Here the criterion is to check, equation by equation, that norm 1 difference between the target and the equation evaluation is under a threshold. The threshold is different for each equation type based on the physical unit involved.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
Hadrien-Godard: Just for intel, here it is 10^-2 MVar for each equation (basically the balancing at each PQ node). So for instance 10^-5 means a precision of 100 Var at each node.

src/main/java/com/powsybl/openloadflow/ac/nr/PerEquationTypeStoppingCriteria.java
```diff
@@ -0,0 +1,113 @@
...
+                case BUS_TARGET_V:
+                case BUS_TARGET_V_WITH_SLOPE:
+                case ZERO_V:
+                    if (Math.abs(fx[idx]) >= maxVoltageMismatch) {
+                    // FIXME, condition must be : Math.abs(fx[idx]) * PerUnit.zb(idx_nominal_voltage) >= maxVoltageMismatch
+                    //  We need the value NominalVoltage associated to the equation
+                    //  either get the value or the associated BUS to fetch the information
+                        return false;
+                    }
+                    break;
```
jeandemanged, line 79: I might disagree with the FIXME and leave maxVoltageMismatch in p.u.

Also Entsoe uses this parameter in pu, see [here](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf) line 5567

src/main/java/com/powsybl/openloadflow/ac/nr/CustomNewtonRaphsonStoppingCriteria.java
jeandemanged: According to me, for volatge we want the threshold in pu

src/main/java/com/powsybl/openloadflow/ac/nr/PerEquationTypeStoppingCriteria.java
Hadrien-Godard: Ok then. You can leave voltage mismatch in pu and put default threshold to 10^-4 as in this document.

src/main/java/com/powsybl/openloadflow/ac/nr/PerEquationTypeStoppingCriteria.java
```diff
@@ -0,0 +1,113 @@
...
+import com.powsybl.openloadflow.equations.Vectors;
+import com.powsybl.openloadflow.util.PerUnit;
+
+/**
+ * @author Alexandre Le Jean <alexandre.le-jean at artelys.com>
+ */
+public class PerEquationTypeStoppingCriteria implements NewtonRaphsonStoppingCriteria {
+
+    // FIXME Any idea for the name?
+    private final double defaultValueAlpha = Math.pow(10, -5);
```
Hadrien-Godard, line 22: maxDefaultAlphaAngleMismatch

src/main/java/com/powsybl/openloadflow/ac/nr/PerEquationTypeStoppingCriteria.java
```diff
@@ -0,0 +1,113 @@
...
+
+    // FIXME Any idea for the name?
+    private final double defaultValueRho = Math.pow(10, -5);
```
Hadrien-Godard, line 25: maxDefaultRatioMismatch

src/main/java/com/powsybl/openloadflow/ac/nr/PerEquationTypeStoppingCriteria.java
```diff
@@ -0,0 +1,113 @@
...
+
+    // FIXME Any idea for the name?
+    private final double defaultValueB = Math.pow(10, -5);
```
Hadrien-Godard, line 28: maxDefaultSusceptanceMismatch.

Maybe Math.pow(10, -5) is too big. I do not well know typical values for B susceptance of shunts. Maybe Math.pow(10, -7) is more secure.

src/main/java/com/powsybl/openloadflow/ac/nr/PerEquationTypeStoppingCriteria.java
```diff
@@ -0,0 +1,113 @@
...
+
+    // FIXME Any idea for the name?
+    private final double defaultValuePhi = Math.pow(10, -5);
```
Hadrien-Godard, line 31: You can merge this one with defaultValueAlpha since both work on angle values.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -167,7 +196,14 @@ private static <E extends Enum<E>> List<Object> getEnumPossibleValues(Class<E> e
         new Parameter(MAX_SLACK_BUS_COUNT_NAME, ParameterType.INTEGER, "Maximum slack buses count", LfNetworkParameters.DEFAULT_MAX_SLACK_BUS_COUNT),
         new Parameter(DEBUG_DIR_PARAM_NAME, ParameterType.STRING, "Directory to dump debug files", LfNetworkParameters.DEBUG_DIR_DEFAULT_VALUE, Collections.emptyList(), ParameterScope.TECHNICAL),
         new Parameter(INCREMENTAL_TRANSFORMER_VOLTAGE_CONTROL_OUTER_LOOP_MAX_TAP_SHIFT_PARAM_NAME, ParameterType.INTEGER, "Incremental transformer voltage control maximum tap shift per outer loop", IncrementalTransformerVoltageControlOuterLoop.DEFAULT_MAX_TAP_SHIFT),
-        new Parameter(SECONDARY_VOLTAGE_CONTROL_PARAM_NAME, ParameterType.BOOLEAN, "Secondary voltage control simulation", LfNetworkParameters.SECONDARY_VOLTAGE_CONTROL_DEFAULT_VALUE)
+        new Parameter(SECONDARY_VOLTAGE_CONTROL_PARAM_NAME, ParameterType.BOOLEAN, "Secondary voltage control simulation", LfNetworkParameters.SECONDARY_VOLTAGE_CONTROL_DEFAULT_VALUE),
+        new Parameter(NEWTONRAPHSON_STOPPING_CRITERIA_TYPE_PARAM_NAME, ParameterType.STRING, "Newton raphson stopping criteria type", NEWTONRAPHSON_STOPPING_CRITERIA_TYPE_DEFAULT_VALUE.name()),
```
geofjamg, line 200: Possible values to add like for other enums.

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -377,6 +427,69 @@ public OpenLoadFlowParameters setNewtonRaphsonConvEpsPerEq(double newtonRaphsonC
...
+        this.newtonRaphsonStoppingCriteriaType = newtonRaphsonStoppingCriteriaType;
+        return this;
+    }
+
+    public double getMaxActivePowerMismatch() {
+        return maxActivePowerMismatch;
+    }
+
+    public OpenLoadFlowParameters setMaxActivePowerMismatch(double maxActivePowerMismatch) {
+        this.maxActivePowerMismatch = maxActivePowerMismatch;
```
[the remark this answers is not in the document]
lejeana: I agree that we must have positive values, but should it be strictly positive?

src/main/java/com/powsybl/openloadflow/ac/nr/LineSearchStateVectorScaling.java
```diff
@@ -78,7 +78,7 @@ public NewtonRaphsonStoppingCriteria.TestResult applyAfter(StateVector stateVect
                 equationVector.minus(targetVector);
 
                 // and recompute new norm
-                currentTestResult = stoppingCriteria.test(equationVector.getArray());
+                currentTestResult = stoppingCriteria.test(equationVector.getArray(), null);
```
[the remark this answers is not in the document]
lejeana: How should I fix this? Create a function test() with 1 argument that calls test() in its current version with second argument as "null"?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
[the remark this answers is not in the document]
geofjamg: I would say yes

src/main/java/com/powsybl/openloadflow/ac/nr/LineSearchStateVectorScaling.java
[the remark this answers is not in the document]
geofjamg: I change a little bit `StateVectorScaling`interface to be able to pass equation system

Hadrien-Godard (approved):

geofjamg (approved):

PR #701, opened 2023-01-11
Populate ConnectivityResult in SA post contingency results

gverger:

Connectivity Results have been added to the API of Post Contingency Results for Security Analysis (issue: https://github.com/powsybl/powsybl-core/issues/2183, PR: https://github.com/powsybl/powsybl-core/pull/2413)
Here we want to populate these results.

Currently the connectivity results are not computed and a default value is returned.

In the PostContingencyResults, we will now have the number of synchronous components created when applying a contingency, and the load and generation lost.

This PR is not meant to deal with connected components since there is more work to be done with HVDC lines.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/LfContingency.java
```diff
@@ -75,6 +92,10 @@ public int getIndex() {
         return index;
     }
 
+    public int getNbSynchronousComponents() {
```
annetill, line 95: I think that here you have to store directly the number of components created by the contingency and add the -1 in that class.

pom.xml
```diff
@@ -47,7 +47,6 @@
         <jafama.version>2.3.2</jafama.version>
         <junit-jupiter.version>5.8.2</junit-jupiter.version>
         <slf4jtoys.version>1.6.3</slf4jtoys.version>
-
```
annetill, line 50: Can you leave the space please?

src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystemUpdater.java
```diff
@@ -45,6 +45,7 @@ public void onDisableChange(LfElement element, boolean disabled) {
                 break;
             case BRANCH:
             case HVDC:
+            case SHUNT_COMPENSATOR:
```
annetill, line 48: @obrix can you remove this fix? We have to open a dedicated PR that shows the bug. Can you open it in parallel and I am going to do a use case. Thanks !

obrix: Okay, I just opened a separate pull request.

src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
```diff
@@ -231,8 +245,8 @@ private List<OperatorStrategyResult> createOperatorStrategyResults(DcSecurityAna
         try (LfNetworkList lfNetworks = Networks.load(network, dcParameters.getNetworkParameters(), allSwitchesToOpen, allSwitchesToClose, Reporter.NO_OP)) {
             return lfNetworks.getLargest().filter(LfNetwork::isValid)
                     .map(largestNetwork -> runActionSimulations(context, largestNetwork, dcParameters, propagatedContingencies,
-                            operatorStrategies, actionsById, neededActions))
-                    .orElse(Collections.emptyList());
+                                operatorStrategies, actionsById, neededActions)
+                    ).orElse(Collections.emptyList());
```
annetill, line 249: @annetill fix this

src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
```diff
@@ -294,9 +307,39 @@ private List<OperatorStrategyResult> runActionSimulations(DcSecurityAnalysisCont
...
+                                lfContingency.apply(context.getParameters().getLoadFlowParameters().getBalanceType());
+                                // we build the connectivity result linked to this contingency by opportunity.
+                                ConnectivityResult connectivityResult = new ConnectivityResult(lfContingency.getCreatedSynchronousComponentsCount(), 0,
+                                        lfContingency.getDisconnectedLoadActivePower() * PerUnit.SB,
+                                        lfContingency.getDisconnectedGenerationActivePower() * PerUnit.SB,
+                                        lfContingency.getDisconnectedElementIds());
+                                networkState.restore();
+                                return connectivityResult;
+                            })
+                            .orElse(new ConnectivityResult(0, 0, 0, 0, Collections.emptySet()))
```
geofjamg, line 338: We have a API design issue, maybe we should have done this ConnectivityResult optional.

annetill: If you have a `PostContingencyResult`, you have a `ConnectivityResult`, no?

geofjamg: Bu if you don't have any loss of connectivity for you contingency, you should not provider this object?

annetill: No... right...

geofjamg (approved):

annetill (approved):

PR #704, opened 2023-01-16
Remedial action: partial support of GeneratorAction

jlbouchot:
Signed-off-by: Jean-Luc Bouchot <jlbouchot@gmail.com>

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -30,11 +30,17 @@ public final class LfGeneratorImpl extends AbstractLfGenerator {
...
+
+    private LfNetworkLoadingReport report;
+
     private LfGeneratorImpl(Generator generator, LfNetwork network, LfNetworkParameters parameters, LfNetworkLoadingReport report) {
         super(network, generator.getTargetP());
         this.generatorRef = new Ref<>(generator);
         participating = true;
         droop = DEFAULT_DROOP;
+        this.parameters = parameters;
+        this.report = report;
```
jlbouchot, line 43: These parameters are used, for now, to call the setVoltageControl method for the setTargetV update of the GeneratorAction. Are they better ways to do so? Should we re-think the way these things are passed?

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -133,4 +139,20 @@ public void updateState() {
                 .setP(-targetP)
                 .setQ(Double.isNaN(calculatedQ) ? -generator.getTargetQ() : -calculatedQ);
     }
+
+    @Override
+    public void setRemoteTargetQ(double targetQ) {
+        setReactivePowerControl(getGenerator().getRegulatingTerminal(), targetQ);
```
jlbouchot, line 145: I am unsure this is what is meant with the targetQ option passed to the GeneratorAction from powsybl-core. Can someone confirm this or give a pointer to what is meant?

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -46,6 +44,56 @@ public boolean isRelative() {
         }
     }
 
+    public static final class GeneratorUpdates {
+
+        // Reference to the generator
+        private final LfGenerator generator;
+
+        private final Optional<Boolean> relativeActivePower;
```
annetill, line 52: I am not sure it is the best way to do that:
- You can have a new targetP ; I prefer that we store here only the new targetP directly if the "relative" treatment can be done at creation. It should be in per unit (see load action). 
- You can have a new targetQ ; It should be in per unit.
- You can have a new targetV; it should be in per unit. You cannot change the target V of the generator. You have to get the voltage control and change it... but we have an issue ! Because in case of shared control, be need an action that says that we want to change the targetV of all generators of the control... maybe just an exception as a first approach.
- The regulation is more more complex... if the voltage control already exists, you just have to activate the controller. But if not, you have to create the voltage control at network creation and desactivate the controller (that will be activated later...). Maybe leave it to me if you don't need it.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -68,6 +70,50 @@ private LoadShift(LfBus bus, String loadId, double p0, PowerShift powerShift) {
         }
     }
 
+    public static final class GeneratorUpdates {
```
annetill, line 73: @Hadrien-Godard and me have exchanged about this feature. We think that it is more complex that we though. So we agree on a first light PR with only changes in `targetP`, relatively and absolutely. So in the `GeneratorUpdate` (I prefer in singular) it is better to have just what you need. Just throw in the `create` function to say that it is not yet supported.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -2417,4 +2420,114 @@ void testConnectivityResultOnSplitThreeCC() {
         assertSame(PostContingencyComputationStatus.CONVERGED, postContingencyResult.getStatus());
         assertEquals(2, postContingencyResult.getConnectivityResult().getCreatedSynchronousComponentCount());
     }
+
+    @Test
+    void testSecurityAnalysisWithGeneratorActionRelativeTargetV() {
```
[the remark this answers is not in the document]
jlbouchot: Absolutely right, sorry about that!

jlbouchot:
Note that some work has already been carried over and kept in another branch for whenever we are ready to extend the GeneratorAction to more actions. See https://github.com/powsybl/powsybl-open-loadflow/tree/feature/GeneratorUpdateExtended

Hadrien-Godard (approved):

geofjamg (approved):

PR #718, opened 2023-02-01
Fix zero impedance subgraph

geofjamg:

**Other information**:


geofjamg:
@annetill most of the code is here. There is still some failing tests but help would be welcome to finish the work.
 - Some unit test are not relevant anymore and need to be updated as they have been designed with the check done at loading time.
 - I suspect some tests to be wrong.
 - There is an issue with the shunt + gen controlling same bus (not same var and eq), so there is a bug somewhere.
 - we miss a lot of new unit test to test various spanning tree configurations.
 - ...

src/test/java/com/powsybl/openloadflow/ac/NonImpedantBranchWithBreakerIssueTest.java
```diff
@@ -50,8 +50,8 @@ void busBreakerAndNonImpedantBranchIssue() {
             assertEquals(400, bus.getV(), 0);
             assertEquals(0, bus.getAngle(), 0);
         }
-        assertEquals(-100, network.getGenerator("G1").getTerminal().getQ(), 0);
```
annetill, line 53: @geofjamg here you have a relevant test that must be unchanged by your PR. We have at loading two voltage controls, both local. At equation system creation, we have to "merge" this voltage controls in order to create a remote voltage control with a distrib_q equation. If we don't do that, as you see, just one generator takes all the reactive power and reaches its limits too fast and the other one stay always PV. It is the issue on metrix test cases for instance.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -429,10 +429,12 @@ void sharedVoltageControlT2wtWithZeroImpedanceLinesTest() {
 
         LoadFlowResult result = loadFlowRunner.run(network, parameters);
         assertTrue(result.isOk());
-        assertVoltageEquals(134.279, bus2);
-        assertVoltageEquals(35.73, t2wt.getTerminal2().getBusView().getBus());
-        assertEquals(2, t2wt.getRatioTapChanger().getTapPosition());
-        assertEquals(2, t2wt.getRatioTapChanger().getTapPosition());
+        // where is the zero impedance line?
```
annetill, line 432: Both transformers are controlling their secondary side. Both secondary sides are linked through 2 zero impedance lines.

src/test/java/com/powsybl/openloadflow/NonImpedantBranchTest.java
```diff
@@ -274,10 +282,10 @@ void nonImpedentNetworkWithTwoPVBusesTest() {
         assertActivePowerEquals(-2, l12.getTerminal2());
         assertActivePowerEquals(-2, l23.getTerminal1());
         assertActivePowerEquals(2, l23.getTerminal2());
-        assertReactivePowerEquals(1, l12.getTerminal1());
-        assertReactivePowerEquals(-1, l12.getTerminal2());
-        assertReactivePowerEquals(-1, l23.getTerminal1());
-        assertReactivePowerEquals(1, l23.getTerminal2());
+        assertReactivePowerEquals(2, l12.getTerminal1());
```
annetill, line 285: For me, it must be unchanged.

geofjamg:
@annetill we also need to update incremental outer loops. In order to not update targets 2 times pour a single controlled buses in case of merged voltage control we have to:
 - only process MAIN voltage controls
 - choose the right target v using the same also as classical control
 - get merged controllers and not only local controllers

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
```diff
@@ -0,0 +1,215 @@
...
+                    VoltageControl<?> vc = voltageControls.get(i);
+                    vc.mergeStatus = VoltageControl.MergeStatus.DEPENDENT;
+                    linkVoltageControls(mainVc, vc);
+                }
+            }
+        }
+    }
+
+    public void tryToSplit() {
+        List<LfBranch> disabledBranches = graph.edgeSet().stream()
```
annetill, line 140: @geofjamg and @flo-dup here we remove edges linked to disabled branches. I am not sure to understand where we do the reverse action and if it is needed ?

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
```diff
@@ -0,0 +1,215 @@
...
+
+                for (LfNetworkListener listener : network.getListeners()) {
+                    listener.onZeroImpedanceNetworkSplit(this, splitZns, dc);
+                }
+            } else {
+                boolean atLeastOneOfDisablingBranchIsPartOfSpanningTree = disabledBranches.stream()
+                        .anyMatch(branch -> branch.isSpanningTreeEdge(dc));
+                if (atLeastOneOfDisablingBranchIsPartOfSpanningTree) {
+                    // just update the spanning
+                    updateSpanningTree();
```
flo-dup, line 176: Here you should also remove the disabledBranches from the spanning tree. They've been removed from the graph and therefore are not updated in `updateSpanningTree()`.
```suggestion
                    updateSpanningTree();
                    for (LfBranch branch : disabledBranches) {
                        branch.setSpanningTreeEdge(dc, false);
                    }
```

flo-dup (changes requested):

annetill (changes requested):

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    public static void addBranchAndMerge(LfZeroImpedanceNetwork zn1, LfZeroImpedanceNetwork zn2, LfBranch enabledBranch) {
+        Objects.requireNonNull(zn1);
+        Objects.requireNonNull(zn2);
+        Objects.requireNonNull(enabledBranch);
+        LfNetwork network = zn1.getNetwork();
+        boolean dc = zn1.isDc();
+        Set<LfZeroImpedanceNetwork> zeroImpedanceNetworks = network.getZeroImpedanceNetworks(dc);
+        Graph<LfBus, LfBranch> mergedGraph = new Pseudograph<>(LfBranch.class);
```
flo-dup, line 181: This is a new `Graph` and not a `AsSubgraph` anymore, so we're mixing subgraphs of `createZeroImpedanceSubGraph` and graphs. But I don't see any problems there as we don't play with the parent graph, we don't even retain it explicitly.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -285,4 +295,42 @@ private static boolean isZeroImpedanceBranch(PiModel piModel, boolean dc, double
...
+                        zn1.removeBranchAndTryToSplit(this);
+                    } else {
+                        throw new IllegalStateException("Should not happen");
+                    }
+                } else {
+                    if (zn1 != zn2) {
+                        // zero impedance network merge
+                        LfZeroImpedanceNetwork.addBranchAndMerge(zn1, zn2, this);
+                    } else {
+                        // we need to add the branch again to zero impedance graph and update spanning tree as
```
flo-dup, line 316: the branch needs to be added, sure, but the previous spanning tree is still **a** spanning tree. It might not be a minimal spanning tree, but is that a problem? If the minimal part of the spanning tree is essential, maybe we should emphasize it in the namings.

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
```diff
@@ -0,0 +1,212 @@
...
+            mergedZn.updateVoltageControlMergeStatus();
+        }
+
+        for (LfNetworkListener listener : network.getListeners()) {
+            listener.onZeroImpedanceNetworkMerge(zn1, zn2, mergedZn, dc);
+        }
+    }
+
+    public void addBranch(LfBranch branch) {
+        graph.addEdge(branch.getBus1(), branch.getBus2(), branch);
```
flo-dup, line 201: You cannot add an edge to a `AsSubgraph` if it is not in the parent graph, or you'll get an `IllegalArgumentException`.

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -76,6 +77,7 @@ private void createBusesEquations(EquationSystem<AcVariableType, AcEquationType>
     private void createGeneratorVoltageControlEquations(LfBus bus,
                                                         EquationSystem<AcVariableType, AcEquationType> equationSystem) {
         bus.getGeneratorVoltageControl()
+                .filter(voltageControl -> voltageControl.getMergeStatus() == VoltageControl.MergeStatus.MAIN)
                 .ifPresent(voltageControl -> {
                     if (bus.isGeneratorVoltageControlled()) {
                         if (voltageControl.isLocalControl()) {
```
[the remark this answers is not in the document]
geofjamg: Renamed to createGeneratorLocalVoltageControlEquation

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -281,88 +325,111 @@ private List<EquationTerm<AcVariableType, AcEquationType>> createReactiveTerms(L
...
+        checkNotDependentVoltageControl(voltageControl);
         LfBus controlledBus = voltageControl.getControlledBus();
         if (voltageControl.isLocalControl()) {
-            equationSystem.getEquation(controlledBus.getNum(), AcEquationType.BUS_TARGET_V)
-                    .orElseThrow()
-                    .setActive(!controlledBus.isDisabled() && controlledBus.isGeneratorVoltageControlEnabled());
-            equationSystem.getEquation(controlledBus.getNum(), AcEquationType.BUS_TARGET_Q)
-                    .orElseThrow()
-                    .setActive(!controlledBus.isDisabled() && !controlledBus.isGeneratorVoltageControlEnabled());
+            if (voltageControl.isHidden()) {
```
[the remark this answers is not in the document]
geofjamg: Right, fixed

src/main/java/com/powsybl/openloadflow/network/LfBus.java
```diff
@@ -28,6 +28,12 @@ public interface LfBus extends LfElement {
 
     void setReference(boolean reference);
 
+    List<VoltageControl<?>> getVoltageControls();
```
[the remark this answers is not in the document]
geofjamg: I added Javadoc to explain the purpose of each of this 3 methods. Is there something more I can do ? Rename with a more explicit name?

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -177,8 +136,8 @@ public void updateSlackBuses() {
     }
 
     private void invalidateZeroImpedanceNetworks() {
```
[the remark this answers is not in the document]
geofjamg: Yes I see than this badly designed but a better solution is not clear to me. To handle in a next work.

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
geofjamg: I replaced by a subgraph + copy

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
geofjamg: Fixed by subgraph + copy

src/main/java/com/powsybl/openloadflow/network/ShuntVoltageControl.java
```diff
@@ -12,8 +12,10 @@
  */
 public class ShuntVoltageControl extends DiscreteVoltageControl<LfShunt> {
 
+    private static final int PRIORITY = 2;
```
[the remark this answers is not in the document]
geofjamg: Yes it should better be configurable

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -285,4 +295,42 @@ private static boolean isZeroImpedanceBranch(PiModel piModel, boolean dc, double
             return piModel.getZ() < lowImpedanceThreshold;
         }
     }
+
+    private void updateZeroImpedanceNetworks(boolean disabled, boolean dc) {
+        if (isZeroImpedance(dc)) {
+            LfZeroImpedanceNetwork zn1 = bus1.getZeroImpedanceNetwork(dc);
+            LfZeroImpedanceNetwork zn2 = bus2.getZeroImpedanceNetwork(dc);
+            if (zn1 != null && zn2 != null) {
```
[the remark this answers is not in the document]
geofjamg: I remember now which I added this check. It is to avoid updating dc part in case of ac calculation and the invert. as LfBus.getZeroImpedanceNetwork returns null if no zero impedance infos (spanning tree) as been queried in the model.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
geofjamg: Not clear anymore for me what was the original issue we tried to solve with this. This line of code is not covered by a unit test but we absolutely need to do it. @annetill help would be welcome to try to create a use case from the original real data to cover this case.

annetill: Are we talking about the Sunday's fix?

geofjamg: Yes

geofjamg: @flo-dup looking at new TU added @annetill I think this case covers an addition to the zero impedance graph (so a new bus that has just joined another zero impedanc graph). So it was not anymore a spanning tree of the graph as a vertex was missing.

geofjamg: @flo-dup I think you are right, the unit test added by @annetill still pass by just adding the edge but not recomputing the spanning tree

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
```diff
@@ -0,0 +1,217 @@
...
```
geofjamg, line 207: @flo-dup so here we don't need to update the spanning tree anymore. Are we agree?

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
```diff
@@ -0,0 +1,218 @@
...
+
+    public void updateVoltageControlMergeStatus() {
+        Map<VoltageControl.Type, List<VoltageControl<?>>> voltageControlsByType = new EnumMap<>(VoltageControl.Type.class);
+        for (LfBus zb : graph.vertexSet()) { // all enabled by design
+            for (VoltageControl<?> vc : zb.getVoltageControls()) {
+                voltageControlsByType.computeIfAbsent(vc.getType(), k -> new ArrayList<>())
+                        .add(vc);
+                vc.getMergedDependentVoltageControls().clear();
+                vc.mainMergedVoltageControl = null;
+                vc.mergeStatus = VoltageControl.MergeStatus.MAIN;
```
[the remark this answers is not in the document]
geofjamg: Done

src/main/java/com/powsybl/openloadflow/network/LfZeroImpedanceNetwork.java
```diff
@@ -0,0 +1,218 @@
...
+    @SuppressWarnings("unchecked")
+    private static void linkVoltageControls(VoltageControl<?> mainVc, VoltageControl<?> vc) {
+        mainVc.mergedDependentVoltageControls.add((VoltageControl) vc);
+        vc.mainMergedVoltageControl = (VoltageControl) mainVc;
+    }
+
+    public void updateVoltageControlMergeStatus() {
+        Map<VoltageControl.Type, List<VoltageControl<?>>> voltageControlsByType = new EnumMap<>(VoltageControl.Type.class);
+        for (LfBus zb : graph.vertexSet()) { // all enabled by design
+            for (VoltageControl<?> vc : zb.getVoltageControls()) {
```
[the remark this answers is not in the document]
flo-dup: Created issue #765

src/main/java/com/powsybl/openloadflow/ac/AcTargetVector.java
```diff
@@ -37,19 +34,10 @@ private static double getBusTargetV(LfBus bus) {
         return targetV;
     }
 
-    private static Optional<Double> getVoltageControlledTargetValue(LfBus bus) {
-        return bus.getGeneratorVoltageControl().filter(vc -> bus.isGeneratorVoltageControlled()).map(vc -> {
-            if (vc.getControllerElements().stream().noneMatch(LfBus::isGeneratorVoltageControlEnabled)) {
```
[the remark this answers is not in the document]
flo-dup: Created issue #765

flo-dup (approved):
Nice job! :tada:

PR #719, opened 2023-02-03
DcSecurityAnalysis : Use reader/writer API instead of the result object based one

obrix:

**Other information**:


src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
```diff
@@ -136,87 +137,90 @@ SecurityAnalysisReport runSync(String workingVariantId, SecurityAnalysisParamete
...
+
+                //Work based on the fact all sensitivity values from pre contingency will be processed before post contingency
+                if (contingencyIndex == -1) {
+                    String branchId = factor.getFunctionId();
+                    Branch<?> branch = network.getBranch(branchId);
+                    double i1 = currentActivePower(Math.abs(functionReference), branch.getTerminal1().getVoltageLevel().getNominalV(), context.getDcPowerFactor());
+                    double i2 = currentActivePower(Math.abs(functionReference), branch.getTerminal2().getVoltageLevel().getNominalV(), context.getDcPowerFactor());
+                    BranchResult branchResult = new BranchResult(branchId, functionReference, Double.NaN, i1, -functionReference, Double.NaN, i2, Double.NaN);
+                    context.getPreContingencyAllBranchResults().put(branchId, branchResult);
+                    context.getDetector().checkActivePower(branch, Branch.Side.ONE, Math.abs(functionReference), violation -> context.getPreContingencyLimitViolationsMap().put(Pair.of(violation.getSubjectId(), violation.getSide()), violation));
```
geofjamg, line 163: checkActivePower and checkCurrent seems to be quite slow.

src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
```diff
@@ -136,87 +137,90 @@ SecurityAnalysisReport runSync(String workingVariantId, SecurityAnalysisParamete
...
-            context.getFactors().add(new SensitivityFactor(SensitivityFunctionType.BRANCH_ACTIVE_POWER_1, b.getId(), SensitivityVariableType.INJECTION_ACTIVE_POWER,
-                    variableId, false, contingencyContext));
-        }
-        SensitivityAnalysisResult result = new SensitivityAnalysis.Runner(sensitivityAnalysisProvider)
-                .run(network, workingVariantId, context.getFactors(), context.getContingencies(), variableSets, sensitivityAnalysisParameters, computationManager, Reporter.NO_OP);
+        SensitivityFactorReader factorReader = handler -> {
+            for (Branch<?> b : network.getBranches()) {
+                SensitivityFactor f = new SensitivityFactor(SensitivityFunctionType.BRANCH_ACTIVE_POWER_1, b.getId(), SensitivityVariableType.INJECTION_ACTIVE_POWER,
+                        variableId, false, contingencyContext);
+                context.getFactors().add(f);
```
geofjamg, line 144: We don't really need to save factors because we just get the branch id when writing values and we could directly using the factorIndex go into the branches array, as there is as many factors than branches.

geofjamg: Anyway it won't change anything to performance

geofjamg (approved):

PR #722, opened 2023-02-06
Incremental phase shifter outer loop

geofjamg:

No

Feature

We have a phase shifter outer loop based on continuous solving (modeled inside Jacobian) + rounding. 

We have another phase shifter outer loop based on sensi calculation.

**Other information**:


Hadrien-Godard (approved):

src/main/java/com/powsybl/openloadflow/DefaultOuterLoopConfig.java
```diff
@@ -43,6 +43,17 @@ private static OuterLoop createShuntVoltageControlOuterLoop(OpenLoadFlowParamete
         }
     }
 
+    private static OuterLoop createPhaseShifterControlOuterLoop(OpenLoadFlowParameters parametersExt) {
+        switch (parametersExt.getPhaseShifterControlMode()) {
+            case CONTINUOUS_WITH_DISCRETISATION:
```
annetill, line 48: `CONTINUOUS_WITH_DISCRETISATION` is a good parameter name, but it is not really coherent with the choices we have made for generator and shunt voltage control. For both we have the incremental mode but also a `WITH_GENERATOR` mode. So what is your proposal to make everything more coherent?

PR #731, opened 2023-02-13
Fix sensitivity calculation

geofjamg:

No

Bug fix

Even if there is no error in sensi calculation, calculation is not properly done. 

Sensi calculation is better implemented.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/ClosedBranchSide1ReactiveFlowEquationTerm.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/ClosedBranchSide1ReactiveFlowEquationTerm.java
@@ -40,7 +40,7 @@ public class ClosedBranchSide1ReactiveFlowEquationTerm extends AbstractClosedBra
                 + dq1dv1(y, cosKsi, b1, v1, r1, v2, cosTheta) * dv1
                 + dq1dv2(y, v1, r1, cosTheta) * dv2
                 + dq1da1(y, v1, r1, v2, sinTheta) * da1
-                + dq1dr1(y, cosKsi, b1, v1, r1, v2, cosTheta);
+                + dq1dr1(y, cosKsi, b1, v1, r1, v2, cosTheta) * dr1;
     }
 
     public static double q1(double y, double cosKsi, double b1, double v1, double r1, double v2, double cosTheta) {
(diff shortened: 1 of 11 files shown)


Hadrien-Godard (approved):

src/main/java/com/powsybl/openloadflow/ac/equations/ClosedBranchSide1ReactiveFlowEquationTerm.java
```diff
@@ -34,10 +34,13 @@ protected double calculateSensi(double dph1, double dph2, double dv1, double dv2
         double theta = theta1(ksi, ph1(), a1(), ph2());
         double cosTheta = FastMath.cos(theta);
         double sinTheta = FastMath.sin(theta);
+        double cosKsi = FastMath.cos(ksi);
         return dq1dph1(y, v1, r1, v2, sinTheta) * dph1
                 + dq1dph2(y, v1, r1, v2, sinTheta) * dph2
-                + dq1dv1(y, FastMath.cos(ksi), b1, v1, r1, v2, cosTheta) * dv1
-                + dq1dv2(y, v1, r1, cosTheta) * dv2;
+                + dq1dv1(y, cosKsi, b1, v1, r1, v2, cosTheta) * dv1
```
AnkurArohi, line 40: @geofjamg 
Is this not reactive power sensitivity calculation, based on the nodal voltage variations, at this particular node.
One would just couple it up with the admittance matrix to get all nodes based reactive power sensitivity,  your opinion?

geofjamg: This is part of reactive power sensitivity calculation. 
This is what is done bit it is managed at the `Equation` level.

AnkurArohi: This is exactly what would be intresting,just the management at equation level should be coupled with the admittance matrix in order to be able to determine the reactive power sensitivity at he NVKP level.
This is similar to GLSK factor for the reactive power.

AnkurArohi: Can this info be ported out?

PR #740, opened 2023-02-24
Add Newton Raphson report

caioluke:

It adds a report for the Newton Raphson's iterations


src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -195,4 +196,74 @@ public static Reporter createPostContingencySimulation(Reporter reporter, String
         return reporter.createSubReporter("postContingencySimulation", "Post-contingency simulation '${contingencyId}'",
                 "contingencyId", contingencyId);
     }
+
+    public static Reporter createNewtonRaphsonReporter(Reporter reporter, int networkNumCc, int networkNumSc) {
+        return reporter.createSubReporter("newtonRaphson", "Network CC${networkNumCc} SC${networkNumSc}",
```
obrix, line 201: Maybe we can put something more explicit here Like "Newton Raphson on Network CC#X SC#X".

geofjamg (changes requested):
Be careful of not having to many reports. What you did is ok for a simple LF but probably not acceptable when used in a security analysis. We could end up with 1000 or 10000 reports and associated performance issue (for instance in gridsuite we store them in a DB). Also main idea of reports (aka functional logs) are to be human readable so it should always be sized to be readable analysable by a human with complex search engine.

obrix:
@geofjamg The issue with the size of logs for security analysis has been discussed on our side, but the only way I see to filter some of it is to filter it from the reporter after the run. Is this really an issue in any situation where we don't use a default serializer ?

geofjamg:
> @geofjamg The issue with the size of logs for security analysis has been discussed on our side, but the only way I see to filter some of it is to filter it from the reporter after the run. Is this really an issue in any situation where we don't use a default serializer ?

Instead of having just a `Reporter` traversing all OLF code we could have something like:
```java
class ReportContext {
    Reporter reporter;
    boolean detailed;
}
``` 

And depending of the calling context of the `NewtonRaphson` (SA or LF) set a different value for `detailled`

Also we could let the user configure an OLF specific parameter to force having detailled report everywhere for debugging purpose.

Also @annetill @flo-dup maybe we should do better and use the Report severity instead of the boolean to better filter what will be added to the reporter. That would mean that all OLF logs should have a severity. So in that case we would have:
```java
class ReportContext {
    Reporter reporter;
    TypedValue severityThreshold;
}
```

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -64,7 +66,36 @@ public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findL
...
+                .filter(e -> Math.abs(e.getValue()) > Math.pow(10, -7))
+                .filter(e -> e.getKey().getType() == equationType)
+                .sorted(Comparator.comparingDouble((Map.Entry<Equation<AcVariableType, AcEquationType>, Double> e) -> Math.abs(e.getValue())).reversed())
+                .limit(1)
+                .collect(Collectors.toList());
+    }
+
+    public static void reportLargestMismatch(Reporter reporter, EquationSystem<AcVariableType, AcEquationType> equationSystem, double[] mismatch, LfNetwork network, int iteration) {
+        for (AcEquationType acEquationType : List.of(AcEquationType.BUS_TARGET_P, AcEquationType.BUS_TARGET_Q, AcEquationType.BUS_TARGET_V)) {
+            findLargestMismatch(equationSystem, mismatch, acEquationType)
```
jeandemanged, line 81: here calling 3 times `findLargestMismatch`, hence hitting all equations 3 times, probably a performance hit.

also separating the 3 equation types in 3 reports is not very convenient. put them all in single report.

jeandemanged:
Use case for iteration logs is mostly for LF. I do not see any practical use case for SA.

ReportContext is a nice idea. Just wondering if it should hold just global thresholds, of reporting options per functionality:

class ReportContext {
    Reporter reporter;
    TypedValue nrIterationsSeverityThreshold; // severity threshold for newton raphson iterations
    TypedValue someOtherFunctionalitySeverityThreshold;
}

+ we need options in OpenLoadFlow parameters to setup these thresholds

flo-dup:
After discussing with @obrix I think that it would be nice to add the threshold within the reporter, like it is done in slf4j. This would lighten the code, and we could still add a `if (reporter.isEnabled(severity))` for the reports which might need some extra computation time, or which occur within a loop.

Note that, as @geofjamg said, the latter reports (within a loop) are not recommended, you'd rather put a summary at the end of the loop.

About thresholds, there's no ordering in severities so far. If we want to speak about thresholds, we should add ordering, or otherwise speak about a list of enabled severities.

obrix:
@geofjamg @jeandemanged  @flo-dup To sum up a bit, @caioluke has opened a pull request on core side to begin the addition of severity filtering from within the reporter. But I think we can probably continue the work on openload flow side without these changes for the moment if we are not sure about how we want to change the reporter API.
If we go with parameters for NR severity threshold we can do the tests for severity on olf side before calling Reporter.report.
In this case the severity attribute of the report itself is unused though because we will just check the severity treshold in various place of the code to do or not do the Reporter.report.

jeandemanged:
@caioluke @obrix 

after discussion at TSC:
- no modification will be made to reporter api in powsybl-core
- add parameters in OLFParameters to choose whether NR iteration logs should be reported true/false
  - for LoadFlow
  - for SecurityAnalysis (only for diverging contingencies)

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -64,7 +66,36 @@ public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findL
                 .collect(Collectors.toList());
     }
 
-    private NewtonRaphsonStatus runIteration(StateVectorScaling svScaling) {
+    public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findLargestMismatch(EquationSystem<AcVariableType, AcEquationType> equationSystem, double[] mismatch, AcEquationType equationType) {
```
flo-dup, line 69: No need for a list, you'd better return an `Optional<Pair<Equation<AcVariableType, AcEquationType>, Double>>`

geofjamg:
I propose a more generic and extensible way to implement this feature. Instead of having to add a new parameter each time we want add log from a new functional module, why not having more generic parameters "addFeaturesToReports" which would be a STRING_LIST.
addFeaturesToReports: nrLf, nrSaDivergingContingencies,... 

And still with a ContextReport, but with a Set<String> addFeaturesToReports inside.

@obrix @caioluke @jeandemanged @flo-dup ?

obrix:
@geofjamg Okay for a list as parameters for reporting. Maybe we can use an enum for the possible values : 

public enum ReportedFeatures{
    NEWTON_RAPHSON_LOAD_FLOW,
    NEWTON_RAPHSON_SA_DIVERGING_CONTINGENCIES,
}

geofjamg:
> @geofjamg Okay for a list as parameters for reporting. Maybe we can use an enum for the possible values :
> 
> public enum ReportedFeatures{ NEWTON_RAPHSON_LOAD_FLOW, NEWTON_RAPHSON_SA_DIVERGING_CONTINGENCIES, }

Yes good idea

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -322,6 +331,10 @@ public enum ReactiveRangeCheckMode {
 
     private PhaseShifterControlMode phaseShifterControlMode = PHASE_SHIFTER_CONTROL_MODE_DEFAULT_VALUE;
 
+    private Set<ReportedFeatures> reportedFeatures = REPORTED_FEATURES_DEFAULT_VALUE;
+
+    private boolean detailedNrReport = false;
```
geofjamg, line 336: This one is not needed anymore? A empty reportedFeatures means no additional reports?

caioluke: *A empty reportedFeatures means no additional reports?* Yes, it is verified via `.contains(key)`

*This one is not needed anymore?* In the way I coded it is still neded. The problem I faced is that inside the NR it doesn't know if it's running a LF or a SA. So I created this variable and I set it to

`parametersExt.getReportedFeatures().contains(OpenLoadFlowParameters.ReportedFeatures.NEWTON_RAPHSON_LOAD_FLOW)`

inside the creation of OLF parameters, and to `lfParametersExt.getReportedFeatures().contains(OpenLoadFlowParameters.ReportedFeatures.NEWTON_RAPHSON_SA))` 

inside the creation of SA parameters.

flo-dup: It's really confusing to do this, you're hijacking the parameters class by doing this. No, you'd better find another way to do this, like passing the corresponding boolean in `OpenLoadFlowParameters::createAcParameters`, similarly to `forceA1Var` variable.

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -63,7 +65,36 @@ public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findL
                 .collect(Collectors.toList());
     }
 
-    private NewtonRaphsonStatus runIteration(StateVectorScaling svScaling, MutableInt iterations) {
+    public static void reportLargestMismatch(Reporter reporter, EquationSystem<AcVariableType, AcEquationType> equationSystem, double[] mismatch, LfNetwork network, int iteration) {
+        List<Pair<Equation<AcVariableType, AcEquationType>, Double>> mismatchEquations = equationSystem.getIndex().getSortedEquationsToSolve().stream()
```
flo-dup, line 69: You want to find the largest mismatch by equation type, so you want to group the mismatches by equation type and retain only the max by mismatch. There's `Collectors::groupingBy` which does exactly this kind of operation. This will simplify things, as you don't need to create a full list which will be later filtered by equation type.
```java
Map<AcEquationType, Optional<Pair<Equation<AcVariableType, AcEquationType>, Double>>> mismatchEquations = equationSystem.getIndex().getSortedEquationsToSolve().stream()
        .map(equation -> Pair.of(equation, mismatch[equation.getColumn()]))
        .collect(Collectors.groupingBy(e -> e.getKey().getType(), Collectors.maxBy(Comparator.comparingDouble(e -> Math.abs(e.getValue())))));
```
I don't think it's still useful to compare with `Math.pow(10, -7)` in that case.

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -63,7 +65,36 @@ public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findL
...
+                .sorted(Comparator.comparingDouble((Map.Entry<Equation<AcVariableType, AcEquationType>, Double> e) -> Math.abs(e.getValue())).reversed())
+                .collect(Collectors.toList());
+
+        List<AcEquationType> acEquationTypes = List.of(AcEquationType.BUS_TARGET_P, AcEquationType.BUS_TARGET_Q, AcEquationType.BUS_TARGET_V);
+
+        for (AcEquationType acEquationType : acEquationTypes) {
+            mismatchEquations.stream()
+                    .filter(e -> e.getKey().getType() == acEquationType)
+                    .limit(1)
+                    .forEach(e -> {
```
flo-dup, line 81: If you use the map you won't need this anymore, but don't do `limit(1).forEach()`, it would be clearer to use `findFirst().ifPresent()` instead.

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
caioluke: Thank you @flo-dup  for the advice!

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
geofjamg: What could be the purpose of having something like Map<X, Optional<Y>> ?
If optional is empty, there is no reason to have not in the map, right?
So why not just Map<X, Y> ?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
geofjamg: @caioluke I pushed a design fix proposition. Are you ok with it?

geofjamg:
> > @geofjamg Okay for a list as parameters for reporting. Maybe we can use an enum for the possible values :
> > public enum ReportedFeatures{ NEWTON_RAPHSON_LOAD_FLOW, NEWTON_RAPHSON_SA_DIVERGING_CONTINGENCIES, }
> 
> Yes good idea

So we don't do it ? (not a problem for me)

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -238,6 +243,11 @@ public enum PhaseShifterControlMode {
         INCREMENTAL
     }
 
+    public enum ReportedFeatures {
+        NEWTON_RAPHSON_LOAD_FLOW,
+        NEWTON_RAPHSON_SA,
```
geofjamg, line 248: SA might be confusing here as traditionally it means "Security Analysis" but looking at your implementation it is used for both security analysis and sensitivity analysis (only in sensitivity analysis in fact before my push) . Is it on purpose?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -718,6 +730,15 @@ public OpenLoadFlowParameters setAlwaysUpdateNetwork(boolean alwaysUpdateNetwork
         return this;
     }
 
+    public Set<ReportedFeatures> getReportedFeatures() {
+        return reportedFeatures;
+    }
+
+    public OpenLoadFlowParameters setReportedFeatures(Set<ReportedFeatures> reportedFeatures) {
+        this.reportedFeatures = reportedFeatures;
```
geofjamg, line 738: Objects.requireNotNull(...)

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -1195,7 +1223,8 @@ public static boolean equals(LoadFlowParameters parameters1, LoadFlowParameters
                 extension1.isSecondaryVoltageControl() == extension2.isSecondaryVoltageControl() &&
                 extension1.getReactiveLimitsMaxPqPvSwitch() == extension2.getReactiveLimitsMaxPqPvSwitch() &&
                 extension1.getPhaseShifterControlMode() == extension2.getPhaseShifterControlMode() &&
-                extension1.isAlwaysUpdateNetwork() == extension2.isAlwaysUpdateNetwork();
+                extension1.isAlwaysUpdateNetwork() == extension2.isAlwaysUpdateNetwork() &&
+                extension1.getReportedFeatures() == extension2.getReportedFeatures();
```
geofjamg, line 1227: This is not the way to compare 2 sets!

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -78,7 +78,7 @@ private void runOuterLoop(OuterLoop outerLoop, OuterLoopContextImpl outerLoopCon
                 LOGGER.debug("Start outer loop '{}' iteration {}", outerLoop.getType(), runningContext.outerLoopTotalIterations);
 
                 // if not yet stable, restart Newton-Raphson
-                runningContext.lastNrResult = newtonRaphson.run(new PreviousValueVoltageInitializer());
+                runningContext.lastNrResult = newtonRaphson.run(new PreviousValueVoltageInitializer(), context.getNetwork().getReporter(), outerLoopIteration.toInteger() + 1, outerLoop.getType());
```
geofjamg, line 81: We cannot do that! Separation of concern is a key of the design. NR does not known anything about outer loops.

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -63,7 +65,33 @@ public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findL
                 .collect(Collectors.toList());
     }
 
-    private NewtonRaphsonStatus runIteration(StateVectorScaling svScaling, MutableInt iterations) {
+    public static void reportLargestMismatch(Reporter reporter, EquationSystem<AcVariableType, AcEquationType> equationSystem, double[] mismatch, LfNetwork network, int iteration) {
+        Map<AcEquationType, Optional<Pair<Equation<AcVariableType, AcEquationType>, Double>>> mismatchEquations = equationSystem.getIndex().getSortedEquationsToSolve().stream()
```
geofjamg, line 69: Optional here looks not good.

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -63,7 +65,33 @@ public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findL
...
+                    .ifPresent(equationPair -> {
+                        Equation<AcVariableType, AcEquationType> equation = equationPair.getKey();
+                        double equationMismatch = equationPair.getValue();
+                        int elementNum = equation.getElementNum();
+                        String elementId = equation.getElement(network).map(LfElement::getId).orElse("?");
+                        int busVRow = equationSystem.getVariable(elementNum, AcVariableType.BUS_V).getRow();
+                        int busPhiRow = equationSystem.getVariable(elementNum, AcVariableType.BUS_PHI).getRow();
+                        double busV = equationSystem.getStateVector().get(busVRow);
+                        double busPhi = equationSystem.getStateVector().get(busPhiRow);
+                        if (LOGGER.isTraceEnabled()) {
```
geofjamg, line 86: So it does not only report? It also log using SLF4J. Is it intended to replace the previous one (X largest mismatch) or be in addition to?

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -63,7 +65,33 @@ public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findL
                 .collect(Collectors.toList());
     }
 
-    private NewtonRaphsonStatus runIteration(StateVectorScaling svScaling, MutableInt iterations) {
+    public static void reportLargestMismatch(Reporter reporter, EquationSystem<AcVariableType, AcEquationType> equationSystem, double[] mismatch, LfNetwork network, int iteration) {
+        Map<AcEquationType, Optional<Pair<Equation<AcVariableType, AcEquationType>, Double>>> mismatchEquations = equationSystem.getIndex().getSortedEquationsToSolve().stream()
+                .map(equation -> Pair.of(equation, mismatch[equation.getColumn()]))
+                .collect(Collectors.groupingBy(e -> e.getKey().getType(), Collectors.maxBy(Comparator.comparingDouble(e -> Math.abs(e.getValue())))));
+
+        List<AcEquationType> acEquationTypes = List.of(AcEquationType.BUS_TARGET_P, AcEquationType.BUS_TARGET_Q, AcEquationType.BUS_TARGET_V);
```
geofjamg, line 73: It should be a constant.

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphsonParameters.java
```diff
@@ -83,6 +85,15 @@ public NewtonRaphsonParameters setStateVectorScalingMode(StateVectorScalingMode
         return this;
     }
 
+    public boolean getDetailedNrReport() {
```
geofjamg, line 88: isDetailedNrReport

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -195,4 +196,86 @@ public static Reporter createPostContingencySimulation(Reporter reporter, String
         return reporter.createSubReporter("postContingencySimulation", "Post-contingency simulation '${contingencyId}'",
                 "contingencyId", contingencyId);
     }
+
+    public static Reporter createNewtonRaphsonReporter(Reporter reporter, boolean detailedNrLogs, int networkNumCc, int networkNumSc, int outerLoopIteration, String outerLoopType) {
+        if (detailedNrLogs) {
+            if (outerLoopIteration == 0) {
```
geofjamg, line 202: As previously explained, there is something badly designed using this info here.

src/test/java/com/powsybl/openloadflow/sensi/ac/AcSensitivityAnalysisReportTest.java
```diff
@@ -59,4 +61,78 @@ void testEsgTuto() throws IOException {
...
         assertEquals(refLogExport, logExport);
     }
+
+    @Test
+    void testEsgTutoDetailedNrLogsLf() throws IOException {
+        Network network = EurostagFactory.fix(EurostagTutorialExample1Factory.create());
+        ReporterModel reporter = new ReporterModel("testEsgTutoReport", "Test ESG tutorial report");
+        var lfParameters = new LoadFlowParameters();
+        OpenLoadFlowParameters.create(lfParameters);
+        OpenLoadFlowParameters.create(lfParameters).setReportedFeatures(Set.of(OpenLoadFlowParameters.ReportedFeatures.NEWTON_RAPHSON_LOAD_FLOW));
```
geofjamg, line 71: You create OLF specific parameters 2 times

src/test/java/com/powsybl/openloadflow/sensi/ac/AcSensitivityAnalysisReportTest.java
```diff
@@ -59,4 +61,78 @@ void testEsgTuto() throws IOException {
         String logExport = normalizeLineSeparator(sw.toString());
         assertEquals(refLogExport, logExport);
     }
+
+    @Test
+    void testEsgTutoDetailedNrLogsLf() throws IOException {
```
geofjamg, line 66: Is it the right place (class) for a LF test?

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
caioluke: The main purpose was for it to be in addition to the existing one. But now that you mentioned it, maybe it could replace the previous one as it logs a more digestible information...

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -195,4 +196,82 @@ public static Reporter createPostContingencySimulation(Reporter reporter, String
...
+                break;
+            case BUS_TARGET_V:
+                prefixAcEquationType = "TargetV";
+                break;
+            default:
+                // not implemented for other ac equation types
+                return;
+        }
+
+        Reporter subReporter = reporter.createSubReporter(String.format("%s%s", prefixAcEquationType, suffixIteration), String.format("%s on %s", messageIteration, prefixAcEquationType));
```
flo-dup, line 240: Same remark here and in all the reports below

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphson.java
```diff
@@ -63,7 +69,42 @@ public static List<Pair<Equation<AcVariableType, AcEquationType>, Double>> findL
                 .collect(Collectors.toList());
     }
 
-    private NewtonRaphsonStatus runIteration(StateVectorScaling svScaling, MutableInt iterations) {
+    public static Map<AcEquationType, Pair<Equation<AcVariableType, AcEquationType>, Double>> getLargestMismatchByAcEquationType(EquationSystem<AcVariableType, AcEquationType> equationSystem, double[] mismatch) {
+        return equationSystem.getIndex().getSortedEquationsToSolve().stream()
+                .map(equation -> Pair.of(equation, mismatch[equation.getColumn()]))
+                .collect(Collectors.toMap(e -> e.getKey().getType(), Function.identity(), BinaryOperator.maxBy(Comparator.comparingDouble(e -> Math.abs(e.getValue())))));
```
flo-dup, line 75: easier that way indeed!

src/main/java/com/powsybl/openloadflow/ac/nr/NewtonRaphsonParameters.java
```diff
@@ -83,6 +85,15 @@ public NewtonRaphsonParameters setStateVectorScalingMode(StateVectorScalingMode
         return this;
     }
 
+    public boolean isDetailedNrReport() {
```
flo-dup, line 88: We're in `NewtonRaphsonParameters`, so `isDetailedReport` should be enough?

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -195,4 +196,82 @@ public static Reporter createPostContingencySimulation(Reporter reporter, String
...
+        return reporter.createSubReporter("newtonRaphson", "Newton Raphson on Network CC${newtonRaphsonNetworkNumCc} SC${newtonRaphsonNetworkNumSc} || Outer loop iteration ${newtonRaphsonOuterLoopIteration} and type `${newtonRaphsonOuterLoopType}`",
+                Map.of("newtonRaphsonNetworkNumCc", new TypedValue(networkNumCc, TypedValue.UNTYPED),
+                        "newtonRaphsonNetworkNumSc", new TypedValue(networkNumSc, TypedValue.UNTYPED),
+                        "newtonRaphsonOuterLoopIteration", new TypedValue(outerLoopIteration, TypedValue.UNTYPED),
+                        "newtonRaphsonOuterLoopType", new TypedValue(outerLoopType, TypedValue.UNTYPED)));
+    }
+
+    public static Reporter createNewtonRaphsonMismatchReporter(Reporter reporter, int iteration) {
+        String var1 = iteration == -1 ? "mismatchInitial" : String.format("mismatchIteration%s", iteration);
+        String var2 = iteration == -1 ? "Initial mismatch" : String.format("Iteration %s mismatch", iteration);
```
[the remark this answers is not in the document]
obrix: The issue is not clear to me here. I think it is an issue with the reporterModelSerializer from core and the way it uses keys ? What is the alternative you sugges ?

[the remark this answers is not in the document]
flo-dup: In fact, one of the goals of the reporter is to have all values displayed stored as values, to be able to interpret them. It allows also to serialize the reports with only the values and the keys, not the default messages in English. And then let the end software replace the keys with the string in the wanted language. I don't know if that's any clearer?

So the correct way is to add the value as a parameter of the sub-reporter:
```suggestion
        if (iteration == -1) {
            return reporter.createSubReporter("mismatchInitial", "Initial mismatch");
        } else {
            return reporter.createSubReporter("mismatchIteration", "Iteration ${iteration} mismatch", "iteration", iteration);
        }
```

PR #749, opened 2023-03-07
Log all parameters formatted as a table

geofjamg:

No

Feature

Most of the parameters are logs but not all. They are all logged as an invidual line.

A generic processing of parameter is done to automatically log all of them. 
They are formatted as a table for better readability.

**Other information**:


geofjamg:
> The table is better, but with what you have done, the parameter description only relies on the parameter java naming. It is less explicit than what we have in main branch.

But this is the same as the one your have to put in config.yml or json

annetill (approved):

PR #751, opened 2023-03-11
Refactor voltage control equations

geofjamg:

No

Refactor/Simplify

 - BUS_TARGET_V_WITH_SLOPE is removed and replaced by BUS_TARGET_V to be consistent with other voltage control.
 - In AC equation system, BUS_TARGET_V is first and always created so that we don't need to create it in each voltage control and create missing at the end.
 - In AC equation system, code is shared for the 3 voltage control update (code was duplicated and also buggy for transformer).

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
@@ -49,6 +49,14 @@ public class AcEquationSystemCreator {
             p.setActive(false);
         }
 
+        // maybe to fix later, but there is so part of OLF (like sensitivity) that needs a voltage target equation
+        // deactivated
+        EquationTerm<AcVariableType, AcEquationType> vTerm = equationSystem.getVariable(bus.getNum(), AcVariableType.BUS_V).createTerm();
+        equationSystem.createEquation(bus, AcEquationType.BUS_TARGET_V)
+                .addTerm(vTerm)
+                .setActive(false);
+        bus.setCalculatedV(vTerm);
+
         createGeneratorControlEquations(bus, equationSystem);
 
         createShuntEquations(bus, equationSystem);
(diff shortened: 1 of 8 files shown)


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -49,23 +49,22 @@ private void createBusEquation(LfBus bus,
             p.setActive(false);
         }
 
-        createGeneratorControlEquations(bus, equationSystem);
+        // maybe to fix later, but there is so part of OLF (like sensitivity) that needs a voltage target equation
```
annetill, line 52: This is not equivalent as what we do on main branch. I understand that here you create an inactive equation for all buses. You prefer like this? No performance issue?

annetill: In main we add an if condition on `(!equationSystem.hasEquation(bus.getNum(), AcEquationType.BUS_TARGET_V))`.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -91,25 +90,18 @@ private void createGeneratorControlEquations(LfBus bus,
...
             // equation is: V + slope * qSVC = targetV
             // which is modeled here with: V + slope * (sum_branch qBranch) = TargetV - slope * qLoads + slope * qGenerators
-            equationSystem.createEquation(bus, AcEquationType.BUS_TARGET_V_WITH_SLOPE)
-                    .addTerm(vTerm)
+            equationSystem.getEquation(bus.getNum(), AcEquationType.BUS_TARGET_V).orElseThrow()
                     .addTerms(createReactiveTerms(bus, equationSystem.getVariableSet())
                             .stream()
                             .map(term -> term.multiply(slope))
                             .collect(Collectors.toList()));
-        } else {
```
annetill, line 110: Where are these lines now? You only create the slope equation now.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
geofjamg: Look at the code, this is already what is done. Creating all V equation at beginning or creating necessary equations and adding missing one at is end is exactly the same.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
geofjamg: Already created

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
annetill: yes, your are right! All bus_v equations are activated at the end.

annetill (approved):

PR #755, opened 2023-03-16
New parameter to always update iIDM network even in case of non-convergence

p-arvy:

No

The parameter alwaysUpdateNetwork (in class OpenLoadFlowParameters) allows to update the state of the network even if the final status of Newton-Raphson calculation is not "CONVERGED".

After Newton-Raphson calculation, the state of the network is updated only if the status of the calculation is "CONVERGED".

If parameter alwaysUpdateNetwork equals true, the network state is updated after Newton-Raphson calculation, whatever the result.

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
@@ -1146,7 +1146,7 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
             return false;
         }
 
-        return  extension1.getSlackBusSelectionMode() == extension2.getSlackBusSelectionMode() &&
+        return extension1.getSlackBusSelectionMode() == extension2.getSlackBusSelectionMode() &&
                 extension1.getSlackBusesIds().equals(extension2.getSlackBusesIds()) &&
                 extension1.isThrowsExceptionInCaseOfSlackDistributionFailure() == extension2.isThrowsExceptionInCaseOfSlackDistributionFailure() &&
                 extension1.hasVoltageRemoteControl() == extension2.hasVoltageRemoteControl() &&


Hadrien-Godard (approved):

jeandemanged:
When the NR converges to an unrealistic state, it may happen to end up with negative voltages.

iIDM refuses to set negative voltage, with an exception. We might want to truncate these voltage values to zero ?

Hadrien-Godard:
> When the NR converges to an unrealistic state, it may happen to end up with negative voltages.
> 
> iIDM refuses to set negative voltage, with an exception. We might want to truncate these voltage values to zero ?

Yes indeed. Maybe in another PR?

jeandemanged:
> > When the NR converges to an unrealistic state, it may happen to end up with negative voltages.
> > iIDM refuses to set negative voltage, with an exception. We might want to truncate these voltage values to zero ?
> 
> Yes indeed. Maybe in another PR?

i would do it in com.powsybl.openloadflow.network.impl.LfBusImpl#updateState, just with

`bus.setV(Math.max(v, 0.0)).setAngle(angle);`

@geofjamg @annetill your opinion ?

geofjamg:
> > > When the NR converges to an unrealistic state, it may happen to end up with negative voltages.
> > > iIDM refuses to set negative voltage, with an exception. We might want to truncate these voltage values to zero ?
> > 
> > 
> > Yes indeed. Maybe in another PR?
> 
> i would do it in com.powsybl.openloadflow.network.impl.LfBusImpl#updateState, just with
> 
> `bus.setV(Math.max(v, 0.0)).setAngle(angle);`
> 
> @geofjamg @annetill your opinion ?

yes, let's do it in this PR.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -207,7 +211,8 @@ private static <E extends Enum<E>> List<Object> getEnumPossibleValues(Class<E> e
         new Parameter(MAX_ANGLE_MISMATCH_PARAM_NAME, ParameterType.DOUBLE, "Maximum angle for per equation stopping criteria", MAX_ANGLE_MISMATCH_DEFAULT_VALUE),
         new Parameter(MAX_RATIO_MISMATCH_PARAM_NAME, ParameterType.DOUBLE, "Maximum ratio for per equation stopping criteria", MAX_RATIO_MISMATCH_DEFAULT_VALUE),
         new Parameter(MAX_SUSCEPTANCE_MISMATCH_PARAM_NAME, ParameterType.DOUBLE, "Maximum susceptance for per equation stopping criteria", MAX_SUSCEPTANCE_MISMATCH_DEFAULT_VALUE),
-        new Parameter(PHASE_SHIFTER_CONTROL_MODE_PARAM_NAME, ParameterType.STRING, "Phase shifter control mode", PHASE_SHIFTER_CONTROL_MODE_DEFAULT_VALUE.name(), getEnumPossibleValues(PhaseShifterControlMode.class))
+        new Parameter(PHASE_SHIFTER_CONTROL_MODE_PARAM_NAME, ParameterType.STRING, "Phase shifter control mode", PHASE_SHIFTER_CONTROL_MODE_DEFAULT_VALUE.name(), getEnumPossibleValues(PhaseShifterControlMode.class)),
+        new Parameter(ALWAYS_UPDATE_NETWORK_PARAM_NAME, ParameterType.BOOLEAN, "Update iidm network even if Newton-Raphson algorithm has diverged", ALWAYS_UPDATE_NETWORK_DEFAULT_VALUE)
```
annetill, line 215: Maybe "iidm" precision can be removed here.

jeandemanged (approved):

PR #757, opened 2023-03-20
Sensitivity values filtering

obrix:
Filter sensitivity values with new available parameters.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
@@ -7,6 +7,7 @@
 package com.powsybl.openloadflow.sensi;
 
 import com.powsybl.commons.PowsyblException;
+import com.powsybl.commons.reporter.Reporter;
 import com.powsybl.contingency.ContingencyContext;
 import com.powsybl.iidm.network.*;
 import com.powsybl.loadflow.LoadFlowParameters;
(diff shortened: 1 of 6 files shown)


src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -1178,4 +1182,51 @@ protected static int getLegNumber(SensitivityFunctionType type) {
...
+            case TRANSFORMER_PHASE_3:
+                return isFlowFunction(function) && Math.abs(value) < parameters.getAngleFlowSensitivityValueThreshold();
+            case BUS_TARGET_VOLTAGE:
+                return filterVoltageVariable(Math.abs(value), function, parameters);
+            default:
+                return false;
+        }
+    }
+
+    protected static boolean filterVoltageVariable(double value, SensitivityFunctionType function, SensitivityAnalysisParameters parameters) {
```
annetill, line 1206: Maybe: filterVoltageVariable could be more explicit. And maybe the Math.abs() could be in this method.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -1178,4 +1182,51 @@ protected static int getLegNumber(SensitivityFunctionType type) {
...
     }
+
+    public abstract void analyse(Network network, List<PropagatedContingency> contingencies, List<SensitivityVariableSet> variableSets, SensitivityFactorReader factorReader,
+                        SensitivityResultWriter resultWriter, Reporter reporter, Set<Switch> allSwitchesToOpen);
+
+    public static boolean filterSensitivityValue(double value, SensitivityVariableType variable, SensitivityFunctionType function, SensitivityAnalysisParameters parameters) {
+        switch (variable) {
+            case INJECTION_ACTIVE_POWER:
+            case HVDC_LINE_ACTIVE_POWER:
+                return isFlowFunction(function) && Math.abs(value) < parameters.getFlowFlowSensitivityValueThreshold();
```
geofjamg, line 1193: We should find a way to deal with missing cases. What about if we add a new function type for this variable and we don't have any filtering rule for it? Maybe we should in that case not filter but here we skip it?
Are we sure that we yet cover all combination of variable and function here?

obrix: All combinations documented on powsybl-core side should be covered, except the ones linked to reactive power as discussed with @annetill . Missing cases are handled with the case default "false" which mean unfiltered. If needed we can modify the switchs to explicitly return false on each unsupported cases.

src/test/java/com/powsybl/openloadflow/sensi/ac/AcSensitivityAnalysisTest.java
```diff
@@ -1119,6 +1168,7 @@ void testThreeWindingsTransformerAsFunction() {
     void testThreeWindingsTransformerAsVariable() {
         SensitivityAnalysisParameters sensiParameters = createParameters(false, "b1_vl_0", true);
         sensiParameters.getLoadFlowParameters().setBalanceType(LoadFlowParameters.BalanceType.PROPORTIONAL_TO_GENERATION_P_MAX);
+        sensiParameters.setAngleFlowSensitivityValueThreshold(0.1);
```
annetill, line 1171: Are your sure with the modification of the unit test that we lead to an uncovering somewhere else?

annetill (approved):

annetill (approved):

PR #761, opened 2023-03-26
Add most meshed slack bus max nominal voltage percentile parameter

geofjamg:

No

Feature

In case of most meshed slack bus selection, the nominal voltage used to select the slack bus is taken from the 90 percentile of all buses nominal voltage.

It is now possible to change this percentile value.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/MostMeshedSlackBusSelector.java
```diff
@@ -17,14 +17,26 @@
  */
 public class MostMeshedSlackBusSelector implements SlackBusSelector {
 
+    public static final double MAX_NOMINAL_VOLTAGE_PERCENTILE_DEFAULT_VALUE = 90;
```
geofjamg, line 20: @annetill ok to change it to 95?

annetill: 95 is okay for the UCTE files I have.

PR #762, opened 2023-03-27
Only use branches connected at both sides in most meshed slack bus

geofjamg:

No

Feature

In most meshed slack bus selection, we take into account branches connected at one side.

We only use branches connected at both sides (so the ones that can really flow active power).

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/MostMeshedSlackBusSelector.java
+++ b/src/main/java/com/powsybl/openloadflow/network/MostMeshedSlackBusSelector.java
@@ -29,6 +29,10 @@ public class MostMeshedSlackBusSelector implements SlackBusSelector {
         this.maxNominalVoltagePercentile = maxNominalVoltagePercentile;
     }
 
+    private static int getBranchCountConnectedAtBothSides(LfBus bus) {
+        return (int) bus.getBranches().stream().filter(LfBranch::isConnectedAtBothSides).count();
+    }
+
     @Override
     public SelectedSlackBus select(List<LfBus> buses, int limit) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/network/MostMeshedSlackBusSelector.java
```diff
@@ -41,7 +45,7 @@ public SelectedSlackBus select(List<LfBus> buses, int limit) {
         // select non-fictitious and most meshed bus among buses with the highest nominal voltage
         List<LfBus> slackBuses = buses.stream()
             .filter(bus -> !bus.isFictitious() && bus.getNominalV() == maxNominalV)
-            .sorted(Comparator.comparingInt((LfBus bus) -> bus.getBranches().size())
+            .sorted(Comparator.comparingInt(MostMeshedSlackBusSelector::getBranchCountConnectedAtBothSides)
```
EtienneLt, line 48: small question it counts lines and 2 windings transfo ? Should we add 3 windings transfo ? and maybe hvdc ?

annetill (approved):

src/main/java/com/powsybl/openloadflow/network/MostMeshedSlackBusSelector.java
annetill: I think that HVDC lines should be considered as generators or loads and not as branches in most cases. It is questionnable when the HVDC line is operated in AC emulation, but this operation mode is quite an exception at that time (France Spain, France Italy).

EtienneLt (approved):

PR #763, opened 2023-03-28
[WIP] Automatic DC voltage init selection

geofjamg:

No

Feature

When the network has a phase shifter with a very large angle (like 20° or more), it often end up with a NR divergence if voltage init mode is flat start.

Based on a phase shifter angle threshold, we automatically switch to DC voltage init.  

**Other information**:


src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -739,6 +748,18 @@ public OpenLoadFlowParameters setMostMeshedSlackBusSelectorMaxNominalVoltagePerc
         return this;
     }
 
+    public double getAutoDcInitPhaseShifterAngleThreshold() {
+        return autoDcInitPhaseShifterAngleThreshold;
+    }
+
+    public OpenLoadFlowParameters setAutoDcInitPhaseShifterAngleThreshold(double autoDcInitPhaseShifterAngleThreshold) {
+        if (autoDcInitPhaseShifterAngleThreshold <= 0 || autoDcInitPhaseShifterAngleThreshold >= 360) {
```
EtienneLt, line 756: we do not allow 0 ?

geofjamg:
I will retry it but in a different way.

PR #772, opened 2023-04-19
Fix disabled voltage control equations

annetill:

**Other information**:


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -224,7 +224,7 @@ static <T extends LfElement> void updateRemoteVoltageControlEquations(VoltageCon
                         .setActive(false);
                 equationSystem.getEquation(controllerElement.getNum(), ctrlEqType)
                         .orElseThrow()
-                        .setActive(false);
+                        .setActive(!controllerElement.isDisabled());
```
geofjamg, line 227: There is something, I don't understand: vl.isDisabled() check that controlled bus and ALL controller buses are disabled, so this should be always false and equivalent to actual code.

geofjamg: Understood, check should not be done on voltage control but just the controlled bus.
I pushed a fix.

PR #773, opened 2023-04-19
Replace dc boolean by an enum

geofjamg:

No

Refactoring

There is everywhere a "dc" boolean to get of we are in a DC ou AC LF.
This is not readable enough and there is a lot of code duplication in zero impedance branch management.

The boolean has been replaced by an enum and code has been simplified.

**Other information**:


geofjamg:
if someone has something better than `LoadFlowType`...

src/main/java/com/powsybl/openloadflow/network/LoadFlowType.java
```diff
@@ -0,0 +1,15 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.openloadflow.network;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public enum LoadFlowType {
```
flo-dup, line 12: I'm ok with that name, but just had a look to the literature anyway, for instance [this](https://ieeexplore.ieee.org/abstract/document/1451474) and [that](https://ieeexplore.ieee.org/abstract/document/4956966). Maybe you'd be more convinced with `LoadFlowModel` or `LoadFlowMethod`?

src/main/java/com/powsybl/openloadflow/network/LoadFlowType.java
```diff
@@ -0,0 +1,15 @@
...
+    AC,
+    DC
+}
```
flo-dup, line 15: Maybe add a small comment so that newcomers may understand AC / DC difference? Something like 
```java
public enum LoadFlowMethod {
    /** Conventional non-linear loadflow method */
    AC,
    /** Linear MW-only loadflow method */
    DC
}
```

flo-dup (approved):
Nice! :+1: 
Would be nice to even add this in `LoadFlowParameters` in fact...

src/main/java/com/powsybl/openloadflow/network/LoadFlowType.java
flo-dup: Well forget it, that's an internal enum in fact, it would not be very useful

src/main/java/com/powsybl/openloadflow/network/LoadFlowType.java
geofjamg: I really like power flow model!

geofjamg: "method" is more like the way to solve you model

PR #775, opened 2023-04-20
[WIP] Security analysis (generators + branches + bus bar sections) on real cases

annetill:
…ols excepted generators.

**Other information**:


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -271,6 +271,19 @@ static <T extends LfElement> void updateRemoteVoltageControlEquations(VoltageCon
                             .orElseThrow()
                             .setActive(false);
                 }
+
+                List<T> disabled = voltageControl.getMergedControllerElements()
```
annetill, line 275: Because we re-create distribution equation without taking into account the status of the controller elements, we have to add this in a way.

annetill:
Solved by #784 #785 and #779

PR #784, opened 2023-05-10
Fix isHidden() method and add allDisabled() method

annetill:

The `isHidden()` method could means that all the controller elements are disabled. In that case, the target V equation must be deactivated.
Furthermore, there is a mistake in the method to check if a voltage control is hidden or not. The unit test does not show the bug because the defaut behvior on main branch is 'false'. I suggest a more clear version.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
@@ -215,7 +215,7 @@ public class AcEquationSystemCreator {
                 .map(vc -> equationSystem.getEquation(vc.getControlledBus().getNum(), AcEquationType.BUS_TARGET_V).orElseThrow())
                 .collect(Collectors.toList());
 
-        if (controlledBus.isDisabled()) {
+        if (controlledBus.isDisabled() || voltageControl.allDisabled()) {
             // we disable all voltage control equations
             vEq.setActive(false);
             for (T controllerElement : controllerElements) {
@@ -224,7 +224,7 @@ public class AcEquationSystemCreator {
                         .setActive(false);
                 equationSystem.getEquation(controllerElement.getNum(), ctrlEqType)
                         .orElseThrow()
-                        .setActive(!controllerElement.isDisabled());
+                        .setActive(true);
             }
         } else {
             if (voltageControl.isHidden()) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -224,7 +224,7 @@ static <T extends LfElement> void updateRemoteVoltageControlEquations(VoltageCon
                         .setActive(false);
                 equationSystem.getEquation(controllerElement.getNum(), ctrlEqType)
                         .orElseThrow()
-                        .setActive(!controllerElement.isDisabled());
```
annetill, line 227: Indeed all controller elements are enabled at this stage, so we can directly put 'true'.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -328,7 +328,7 @@ public static void updateGeneratorVoltageControl(GeneratorVoltageControl voltage
         checkNotDependentVoltageControl(voltageControl);
         LfBus controlledBus = voltageControl.getControlledBus();
         if (voltageControl.isLocalControl()) {
-            if (voltageControl.isHidden()) {
+            if (voltageControl.isHidden()) { // FIXME
```
annetill, line 331: Need to discuss with you about what does it means for generator voltage control which priority is 0.

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
```diff
@@ -143,11 +143,19 @@ public boolean isHidden() {
         List<VoltageControl<?>> voltageControls = findVoltageControlsSortedByPriority(controlledBus);
         if (voltageControls.isEmpty()) {
             return true; // means all disabled
-        }
-        // we should normally have max 3 voltage controls (one of each type) because already merged
-        if (voltageControls.size() > 1) {
```
geofjamg, line 148: This code is totally equivalent to previous one. if only one voltage control in the list it is necessary "this" (because we start from it) and so on "voltageControls.get(0) != this" is false.
I would revert this change.

annetill: ´This’ (imagine generator voltage control) could have all its controllers disabled, but the controlled bus can have another voltage control (imagine with transformer) with enabled controllers. In that case, the list of voltage control is of size 1. ´This’ is hidden for me.

geofjamg: yes, agree

geofjamg (approved):

PR #785, opened 2023-05-10
Fix voltage control equation system updater: only main voltage control

annetill:

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemUpdater.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemUpdater.java
@@ -92,13 +92,13 @@ public class AcEquationSystemUpdater extends AbstractEquationSystemUpdater<AcVar
                 break;
             case BRANCH:
                 LfBranch branch = (LfBranch) element;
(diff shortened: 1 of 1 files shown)


src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalTransformerVoltageControlOuterLoop.java
```diff
@@ -213,7 +213,9 @@ public OuterLoopStatus check(OuterLoopContext context, Reporter reporter) {
                 double halfTargetDeadband = getHalfTargetDeadband(voltageControl);
                 if (Math.abs(diffV) > halfTargetDeadband) {
                     controlledBusesOutsideOfDeadband.add(controlledBus.getId());
-                    List<LfBranch> controllers = voltageControl.getMergedControllerElements();
+                    List<LfBranch> controllers = voltageControl.getMergedControllerElements().stream()
+                            .filter(b -> !b.isDisabled() && b.isVoltageController())
```
geofjamg, line 217: "b.isVoltageController()" is not necessary?

annetill: It was exactly my opinion but I put this because we have this everywhere else... I have to remove that condition for the other outer loops too?

geofjamg (approved):

PR #803, opened 2023-06-26
Fix switch update in fast mode

geofjamg:

No

Bug fix

When switch is connecting a new bus we get an error

We use connectivity calculation to correctly connect the right switch

**Other information**:


src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -175,6 +172,32 @@ private boolean onSwitchUpdate(String switchId, boolean open) {
             return found;
         }
 
+        private static void updateSwitch(boolean open, LfNetwork lfNetwork, LfBranch lfBranch) {
+            var connectivity = lfNetwork.getConnectivity();
+            connectivity.startTemporaryChanges();
+            try {
```
annetill, line 178: I think we must do something to mutualise the connectivity management. We have almost the same treatment in LfAction, and it is not the first time we forgot to update connectivity. Any idea?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -102,8 +103,15 @@ private static LoadFlowResult.ComponentResult.Status convertStatus(AcLoadFlowRes
...
 
+    private GraphConnectivityFactory<LfBus, LfBranch> getConnectivityFactory(OpenLoadFlowParameters parametersExt) {
+        return parametersExt.isNetworkCacheEnabled() && !parametersExt.getActionableSwitchesIds().isEmpty()
+                ? new NaiveGraphConnectivityFactory<>(LfBus::getNum)
+                : connectivityFactory;
+    }
+
     private LoadFlowResult runAc(Network network, LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt, Reporter reporter) {
-        AcLoadFlowParameters acParameters = OpenLoadFlowParameters.createAcParameters(network, parameters, parametersExt, matrixFactory, connectivityFactory);
+        GraphConnectivityFactory<LfBus, LfBranch> theConnectivityFactory = getConnectivityFactory(parametersExt);
```
annetill, line 113: What do you mean by "theConnectivityFactory" ?

geofjamg: this is just a different variable name to select either the factory provider in the constructor or force the naive one which is the only one that manage reconnection

geofjamg: to not confuse with the connectivityFactory class attribute

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -358,7 +370,7 @@ public Entry get(Network network, LoadFlowParameters parameters) {
             for (AcLoadFlowContext context : entry.getContexts()) {
                 AcLoadFlowResult result = context.getResult();
                 if (result != null && result.getNewtonRaphsonStatus() == NewtonRaphsonStatus.CONVERGED) {
-                    context.getParameters().setVoltageInitializer(new PreviousValueVoltageInitializer());
+                    context.getParameters().setVoltageInitializer(new PreviousValueVoltageInitializer(true));
```
annetill, line 373: This change is wanted?

geofjamg: yes, this is because in some cases, we can add bus to the main cmponent by closing a switch which was not previously been calculated so when need to fallback to flat start for theses buses

annetill (approved):

PR #808, opened 2023-06-28
Clean indexed terms when removing an equation

geofjamg:

No

Bug fix

When an equation is removed, corresponding indexed terms are not cleaned. 

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/equations/EquationSystem.java
+++ b/src/main/java/com/powsybl/openloadflow/equations/EquationSystem.java
@@ -152,6 +152,14 @@ public class EquationSystem<V extends Enum<V> & Quantity, E extends Enum<E> & Qu
         if (equation != null) {
             Pair<ElementType, Integer> element = Pair.of(type.getElementType(), num);
             equationsByElement.get(element).remove(equation);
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/equations/EquationSystem.java
```diff
@@ -152,6 +152,14 @@ public Equation<V, E> removeEquation(int num, E type) {
         if (equation != null) {
             Pair<ElementType, Integer> element = Pair.of(type.getElementType(), num);
             equationsByElement.get(element).remove(equation);
+            if (equationTermsByElement != null) {
```
annetill, line 155: This is not working on my case and I don't know exactly why. If I put 
````
equationTermsByElement = null;
indexAllTerms();
````

geofjamg: it is supposed to be equivalent but faster. The way you did it reindex everything which would be very slow.

geofjamg: looking at test case I did, does it seem ok?

annetill (approved):

PR #810, opened 2023-06-30
Fix voltage control hidden to disabled

annetill:

**Other information**:


src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
```diff
@@ -148,7 +148,7 @@ public boolean isHidden() {
             return true; // means all disabled
         } else {
             // we should normally have max 3 voltage controls (one of each type) because already merged
-            return voltageControls.get(0) != this;
+            return voltageControls.get(0) != this && voltageControls.get(0).getControlledBus() == this.getControlledBus();
```
geofjamg, line 151: But a voltage control, with a higher priority but not the same controlled bus BUT with a controlled bus connected to same zero impedance network (this is the only case where voltageControls.get(0).getControlledBus() == this.getControlledBus()) also hide this voltage control?

annetill: Yes, you are right.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -216,7 +216,8 @@ static <T extends LfElement> void updateRemoteVoltageControlEquations(VoltageCon
                 .map(vc -> equationSystem.getEquation(vc.getControlledBus().getNum(), AcEquationType.BUS_TARGET_V).orElseThrow())
                 .collect(Collectors.toList());
 
-        if (controlledBus.isDisabled() || voltageControl.isDisabledAndAlsoAllItsDependentVoltageControls()) {
```
annetill, line 219: I disagree with this change. All of this is just a question of having or not having a PV equation at controlled bus (or in the zero impedance network of the controlled bus).
- See the unit test I have added, we deactivate a PV equation. All the controllers are disabled, but we must keep the PV equation because we have a generator.
- Your change is really different than before because you only consider disabled voltage control with the same priority. But with `isDisabledAndAlsoAllItsDependentVoltageControls`, we are looking of all the voltage controls of the zero impedance network, with different priority.

annetill:
````
if (voltageControl.isHidden()) {
            if (condition) { // condition that we have to find together
                      vEq.setActive(false);
            }
            for (T controllerElement : controllerElements) {
                equationSystem.getEquation(controllerElement.getNum(), distrEqType)
                        .orElseThrow()
                        .setActive(false);
                        .orElseThrow()
                        .setActive(true); 
} else {
          List<T> enabledControllerElements = controllerElements.stream()
                        .filter(voltageControl::isControllerEnabled).collect(Collectors.toList());
          List<T> disabledControllerElements = controllerElements.stream()
                        .filter(Predicate.not(voltageControl::isControllerEnabled)).collect(Collectors.toList());

           // activate voltage control at controlled bus only if at least one controller element is enabled
           vEq.setActive(!enabledControllerElements.isEmpty());
}
````
Something like that is much more coherent with all the rework we are doing in outer loops #840 
What do you think ?

File: src/test/java/com/powsybl/openloadflow/network/LfNetworkTest.java [hunk 3]
```diff
@@ -204,4 +204,24 @@ void testGraphViz() throws IOException {
...
+        LfBus b1 = lfNetwork.getBusById("b1_vl_0");
+        assertEquals(VoltageControl.MergeStatus.MAIN, b1.getGeneratorVoltageControl().orElseThrow().getMergeStatus());
+        LfBus b2 = lfNetwork.getBusById("b2_vl_0");
+        assertEquals(VoltageControl.MergeStatus.DEPENDENT, b2.getGeneratorVoltageControl().orElseThrow().getMergeStatus());
+        LfBus b3 = lfNetwork.getBusById("b3_vl_0");
+        assertEquals(VoltageControl.MergeStatus.DEPENDENT, b3.getGeneratorVoltageControl().orElseThrow().getMergeStatus());
+        lfNetwork.getBusById("b01_vl_0").setDisabled(true); // only g1
+        assertFalse(b1.getGeneratorVoltageControl().orElseThrow().isDisabled());
+        assertFalse(b1.getGeneratorVoltageControl().orElseThrow().isHidden());
+        assertTrue(b2.getGeneratorVoltageControl().orElseThrow().isHidden()); // FIXME
```
annetill, line 224: Fix by 
`    public boolean isHidden() {
        // collect all voltage controls with the same controlled bus as this one and also all voltage controls coming
        // from merged ones
        List<VoltageControl<?>> voltageControls = findVoltageControlsSortedByPriority(controlledBus);
        if (voltageControls.isEmpty()) {
            return true; // means all disabled
        } else {
            // we should normally have max 3 voltage controls (one of each type) because already merged
            return voltageControls.get(0) != this && voltageControls.get(0).getMainVoltageControl() != this.getMainVoltageControl();
        }
    }`

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
```diff
@@ -73,11 +73,15 @@ public Type getType() {
         return type;
     }
 
+    /**
+     * Check is the merged voltage to which this voltage control belongs is disabled. Disabled means that there is no
+     * more controlled bus or no more controller element.
```
annetill, line 78: We must add in the comment that loosing a controlled bus in a merge voltage control is an open question. We can also imagine that loosing a controlled bus leads to the loose of its controllers and maybe a different management of the merge status.

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
```diff
@@ -120,46 +137,69 @@ private static void addVoltageControls(List<VoltageControl<?>> voltageControls,
...
         } else {
-            addVoltageControls(voltageControls, bus);
+            addMainVoltageControls(voltageControls, bus);
         }
         voltageControls.sort(Comparator.comparingInt(VoltageControl::getPriority));
         return voltageControls;
     }
 
     /**
+     * Check is the merged voltage to which this voltage control belongs is hidden by another one of a different type
```
annetill, line 159: typo: is -> if

src/main/java/com/powsybl/openloadflow/network/VoltageControl.java
```diff
@@ -120,46 +137,69 @@ private static void addVoltageControls(List<VoltageControl<?>> voltageControls,
...
         }
     }
 
-    public boolean isDisabledAndAlsoAllItsDependentVoltageControls() {
-        // collect all voltage controls with the same controlled bus as this one and also all voltage controls coming
-        // from merged ones
-        List<VoltageControl<?>> voltageControls = findVoltageControlsSortedByPriority(controlledBus);
-        if (voltageControls.isEmpty()) {
-            return true; // means all disabled
+    public boolean isVisible() {
```
annetill, line 181: If you don't need this method, I prefer to use not hidden. It is really clear to me. Visible is less.

annetill: Or make it private.

annetill (commented):
Approved!

PR #813, opened 2023-07-10
Fix updated Q limits in reactiveLimitsOuterLoop

caioluke:

No

Bug Fix

Adds an extra check for the reactive limits outer loop to verify that PQ buses' current minQ/maxQ's do not deviate too much from their generation targetQ, defined in the previous switch PV -> PQ. It could happens if the `targetP` of one of the generators connected to the bus has changed (e.g. because of distributed slack). It can also happen if the voltage of the bus has changed for static var compensator.

In some cases generator bus Q balance may not be zero

No


src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
```diff
@@ -37,6 +37,8 @@ public class ReactiveLimitsOuterLoop implements AcOuterLoop {
 
     public static final int MAX_SWITCH_PQ_PV = 3;
 
+    public static final double MAX_Q_LIMITS_DIFFERENCE = 1E-3;
```
jeandemanged, line 40: we could instead use the NR settings here, but that's maybe overkill

src/main/java/com/powsybl/openloadflow/network/LfBus.java
```diff
@@ -99,6 +104,10 @@ public interface LfBus extends LfElement {
 
     double getMaxQ();
 
+    Optional<QLimitType> getQLimitType();
+
+    void setQLimitType(QLimitType qLimitType);
```
jeandemanged, line 109: not sure if issue or not

we never do setQLimitType(null) when switching PQ->PV.

I do not think it would be required, Just we need to be aware e.g. during debug session that a bus may be PV now but still with an old QLimitType that was set from a previous outerloop.

jeandemanged (approved):

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/network/LfBus.java
geofjamg: Yes that would be more safe to set to null when switching PV

annetill (approved):

geofjamg (approved):

jeandemanged (approved):

PR #815, opened 2023-07-17
Fix calculated load Q when distributed slack on loads with constant power factor

caioluke:

Fix constant power factor when distributing slack on loads

Slack is not correctly distributed on some types of loads, when only the variableReactivePower is not 0.

Slack will always be correctly distributed

**Other information**:


caioluke:
Fix proposal is ok

annetill (approved):

PR #824, opened 2023-08-12
Fix dcUseTransformerRatio not taken into account in DC sensi

geofjamg:

No

Bug fix

When dcUseTransformerRatio option is to true, the post contingency reference flows are not consistent with the one calculated from a DC LF. This ratio seems not to have been taken into account in the alpha calculation that allow us to calculate post contingence reference flow from pre contingency reference flow.

**Other information**:


geofjamg:
Are sensitivity values wrong too and should not be the same depending on dcUseTransformerRatio value?

Hadrien-Godard (approved):

annetill (approved):

PR #826, opened 2023-08-12
Fix branch state saving and restore

geofjamg:

No

Bug fix

When saving a Pi model array, we do not save modified r1 and a1, so that restored state in case of a modified pi model array is not correct.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/BranchState.java
+++ b/src/main/java/com/powsybl/openloadflow/network/BranchState.java
@@ -11,8 +11,8 @@ package com.powsybl.openloadflow.network;
  */
 public class BranchState extends ElementState<LfBranch> {
 
-    private double a1 = Double.NaN;
-    private double r1 = Double.NaN;
+    private final double a1;
+    private final double r1;
     private final boolean phaseControlEnabled;
     private final boolean voltageControlEnabled;
     private Integer tapPosition;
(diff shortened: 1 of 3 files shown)


src/main/java/com/powsybl/openloadflow/network/BranchState.java
```diff
@@ -11,17 +11,21 @@
...
+    private final double r1;
     private final boolean phaseControlEnabled;
     private final boolean voltageControlEnabled;
     private Integer tapPosition;
 
     public BranchState(LfBranch branch) {
         super(branch);
         PiModel piModel = branch.getPiModel();
-        if (piModel instanceof PiModelArray) {
+        if (piModel instanceof PiModelArray piModelArray) {
```
annetill, line 23: What is this syntax?

geofjamg: New Java 17 syntax to avoid casting

src/main/java/com/powsybl/openloadflow/network/PiModelArray.java
```diff
@@ -101,6 +101,14 @@ public double getB2() {
         return getModel().getB2();
     }
 
+    public double getModifiedR1() {
```
annetill, line 104: Why do you need these two methods? It works with `getR1()` and `getA1()`.

geofjamg: I need to raw value of a1 and r1 to save it only if overridden. Look at getA1 and getR1, in case of nan it return value at tap position.

annetill: Maybe "continuous" instead of "modified"?

geofjamg: We already have a continuousR1 to store the continuous value of R1 even after rounding

annetill (approved):

PR #828, opened 2023-08-13
Fix small reference flow values instead of strict zero

geofjamg:

No

Bug fix

When computing the post contingency reference flow of a line which is in series with another line which is itself the contingency, we would expected to post contingency active power flow to be zero. Instead we have a very small value. 

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
@@ -55,7 +55,7 @@ import static com.powsybl.openloadflow.network.util.ParticipatingElement.normali
 public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis<DcVariableType, DcEquationType> {
 
     private static final double CONNECTIVITY_LOSS_THRESHOLD = 10e-7;
-    private static final double REFERENCE_FUNCTION_ZER0_THRESHOLD = 1e-13;
+    private static final double FUNCTION_REFERENCE_ZER0_THRESHOLD = 1e-13;
 
     static final class ComputedContingencyElement {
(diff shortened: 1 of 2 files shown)


geofjamg:
How to solve this? The way post contingency reference flows are calculated in a DC sensi analysis make it quite logic to not have strict zero values but a very small values (like 1e-13) because it is calculated from the pre-contingency reference flow substracted from contingency lost flow * sensi. So we cannot really end up to a perfect zero.
On the other end, for a user point of view (which is not supposed to be aware to how it is calculated), it does not really make sens to not have zero. 
What about, as a post processing, filtering very small values of post contingency reference flows and replace it by zero?

Hadrien-Godard (approved):

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -288,12 +289,26 @@ private void createBranchSensitivityValue(LfSensitivityFactor<DcVariableType, Dc
...
 
+        functionValue = fixZeroReferenceFlow(contingency, functionValue);
+
         double unscaledSensi = unscaleSensitivity(factor, sensitivityValue);
         if (!filterSensitivityValue(unscaledSensi, factor.getVariableType(), factor.getFunctionType(), parameters)) {
             resultWriter.writeSensitivityValue(factor.getIndex(), contingency != null ? contingency.getIndex() : -1, unscaledSensi, unscaleFunction(factor, functionValue));
         }
     }
 
+    /**
```
annetill, line 300: Just for coherence: the term "reference flow" was used in the previous API. Now, in DC sensitivity analysis, even if we have only active power flows as function types, we use in general "reference function".

PR #829, opened 2023-08-14
Improve performance of reference flow calculation in a DC sensi analysis

geofjamg:

No

Performance improvement

For reference flow calculation, we need to calculate pre-contingency state using DcLoadFlowEngine. The problem with current implementation is that we integrate the calculated state into the state vector and update the network which in consequence invalidate the Jacobian which will be needed to be recalculated (identically to before invalidation...).

We use a special version of the DcLoadFlowEngine that do not  update the state vector nor the network. Flow are calculated from equation terme using a special eval method that can take al alternative state vector.

In a real test case 10000 buses, 100 factors and 100000 contingencies, DC sensitivity calculation is now 4 times faster.

**Other information**:


Hadrien-Godard (approved):

annetill (approved):
Great PR, thanks a lot!

PR #834, opened 2023-08-24
Do not update state after UNSTABLE outer loop status

annetill:

No. After an UNSTABLE status for outerloop, we update the network. This is a new behavior after the separation of outerloops and NR in status and iterations.

After an end with UNSTABLE outerloop status, we don't update the network state. This could be a really issue in case of phase control outerloop for example.

**Other information**:


annetill:
Indeed, I wonder about our rules for AC security and sensitivity analysis. We only look at NR status...

geofjamg (approved):

PR #847, opened 2023-09-18
Full Sonar analysis

So-Fras:

Quality

An old Sonar profile is used, it is not up-to-date.

An up-to-date Sonar profile is used in powsybl-open-loadflow CI. New code smells are reported and should therefore be corrected.


annetill:
@geofjamg this is a proposal for making switch more lisible when needed (maybe not every time, as done in the PR). Just tell me what do think. Can we change the automatic indentation for mutliple case in a switch? Furthermore, I have an issue with `default` that are not covered when we switch on enum. Maybe I have to remove default in that case...

geofjamg:
> @geofjamg this is a proposal for making switch more lisible when needed (maybe not every time, as done in the PR). Just tell me what do think. Can we change the automatic indentation for mutliple case in a switch? Furthermore, I have an issue with `default` that are not covered when we switch on enum. Maybe I have to remove default in that case...

When new switch is used as an expression switch (so when we get the returned value from the switch), if not all cases are handled we get a compile time error. So we should in such a case safely remove the default case.
BUT be careful that this is not true, even with new switch syntax when it is used a switch statement instead of a switch expression.

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -204,39 +204,29 @@ public void onUpdate(Identifiable identifiable, String attribute, String variant
...
-                                && onShuntUpdate(shunt, attribute)) {
-                            done = true;
-                        }
-                    } else if (identifiable.getType() == IdentifiableType.SWITCH
-                            && attribute.equals("open")) {
-                        if (onSwitchUpdate(identifiable.getId(), (boolean) newValue)) {
-                            done = true;
-                        }
+
+            if (List.of("v", "angle", "p", "q", "p1", "q1", "p2", "q2").contains(attribute)) {
```
geofjamg, line 208: why this change? this is a sequential search worst in term of performance a a switch/case

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/NetworkCache.java
annetill: Sonar asks for an if loop here instead of a switch case. But I understand if we revert.

PR #852, opened 2023-09-23
New way to per unit lines

geofjamg:

No

Feature

We perunit lines with a different nominal voltage at both sides using a ratio correction.

We perunit lines with a different nominal voltage at both sides using either a ratio correction or an impedance correction.
Impedance correction seems to work better with voltage magnitude initialization.

**Other information**:


Hadrien-Godard (approved):

src/test/java/com/powsybl/openloadflow/Ieee14Test.java
```diff
@@ -98,18 +98,18 @@ void testDc() {
...
-        assertAngleEquals(-15.229412, network.getBusView().getBus("VL6_0"));
-        assertAngleEquals(-13.957433, network.getBusView().getBus("VL7_0"));
-        assertAngleEquals(-13.957433, network.getBusView().getBus("VL8_0"));
-        assertAngleEquals(-16.025846, network.getBusView().getBus("VL9_0"));
-        assertAngleEquals(-16.313486, network.getBusView().getBus("VL10_0"));
-        assertAngleEquals(-15.976867, network.getBusView().getBus("VL11_0"));
-        assertAngleEquals(-16.340761, network.getBusView().getBus("VL12_0"));
-        assertAngleEquals(-16.510537, network.getBusView().getBus("VL13_0"));
-        assertAngleEquals(-17.536792, network.getBusView().getBus("VL14_0"));
+        assertAngleEquals(-5.013434, network.getBusView().getBus("VL2_0"));
```
geofjamg, line 101: @zamarrenolm @marqueslanauja Is it expected to not have same result for DC calculation with no transformer ratio?

marqueslanauja: if `setDcUseTransformerRatio` is set to `false` the result for DC calculation will not be the same as the ratio must be considered to have the exact per unit conversion.

marqueslanauja (approved):

src/test/java/com/powsybl/openloadflow/Ieee14Test.java
JB-H: In the IEEE 14 example we have 2 lines with different Vnom:
---------- Line L7-8-1 has Vnom1 = 14.0 and Vnom2 = 20.0------------------------
---------- Line L7-9-1 has Vnom1 = 14.0 and Vnom2 = 12.0------------------------

JB-H: The two ways of per uniting are strictly equivalent for AC use.

If the option with no transformer ratio is active, I suppose that in the previous way of per uniting setR1(1 / Transformers.getRatioPerUnitBase(line)) is transformed into 1.
In the new way of per-uniting, the option has no effect since R1 is always one by construction.

JB-H: the new way of per uniting seems to keep more consistance between AC model and DC model approx

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -48,19 +47,63 @@ private static void createLineAsym(Line line, double zb, PiModel piModel, LfBran
...
+    private static LfBranchImpl createLine(Line line, LfNetwork network, LfBus bus1, LfBus bus2, LfNetworkParameters parameters) {
+        double r1;
+        double r;
+        double x;
+        double g1;
+        double b1;
+        double g2;
+        double b2;
+        double zb;
+        if (parameters.getLinePerUnitMode() == LinePerUnitMode.RATIO) {
```
annetill, line 63: Note for @phiedw: @geofjamg and @zamarrenolm have worked on a new way to per-unit lines. The idea is to improve management of lines with different nominal voltage at both ends, especially for initialisation. But, we introduce fatal differences in DC load flow with parameter `setDcUseTransformerRatio(false);` (not better not worst). So we have decided to introduce a parameter `LinePerUnitMode` for your transition. To be removed if everything is okay during next release train.

annetill (approved):

PR #853, opened 2023-09-23
Reduce the number of zero reactive key log error

geofjamg:

No

Feature

Each time we have a generator with a zero reactive key, we log an error.

We only warn about the number of generator with a zero reactive key, detailed logs are only at trace level.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
@@ -730,6 +730,11 @@ public class LfNetworkLoaderImpl implements LfNetworkLoader<Network> {
                     lfNetwork, report.generatorsWithInconsistentTargetVoltage);
         }
 
+        if (report.generatorsWithZeroRemoteVoltageControlReactivePowerKey > 0) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -730,6 +730,11 @@ private LfNetwork create(int numCC, int numSC, Network network, List<Bus> buses,
                     lfNetwork, report.generatorsWithInconsistentTargetVoltage);
         }
 
+        if (report.generatorsWithZeroRemoteVoltageControlReactivePowerKey > 0) {
+            LOGGER.warn("Network {}: {} generators have an zero remote voltage control reactive power key",
```
annetill, line 734: Typo: a zero remote...

annetill (approved):
Just a little typo!

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
geofjamg: Fixed

PR #862, opened 2023-09-28
Fix PTC action support and support of RTC action

Hadrien-Godard:

No.

Bugfix

Currently, PTC and RTC PI model does not include tap positions if regulation is deactivated. This is an issue if the PST or RTC is involved in operator strategies / actions.

This PR allows PTC and RTC to be retained also when they are involved in operator strategies / actions.
With this PR, PTC and RTC pimodel contains tap positions if the PTC and RTC is involved not only in regulation but also in at least one action in a security analysis run. 

No

**Other information**:

none


src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -195,7 +192,7 @@ public AcLoadFlowResult run() {
     }
 
     public static <T> List<AcLoadFlowResult> run(T network, LfNetworkLoader<T> networkLoader, AcLoadFlowParameters parameters, Reporter reporter) {
-        return LfNetwork.load(network, networkLoader, parameters.getNetworkParameters(), reporter)
+        return LfNetwork.load(network, networkLoader, new ArrayList<>(), new ArrayList<>(), parameters.getNetworkParameters(), reporter)
```
annetill, line 195: You must do a load method without this two lists because mort of the time we don't need it.

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -523,15 +524,15 @@ public static <T> List<LfNetwork> load(T network, LfNetworkLoader<T> networkLoad
         return load(network, networkLoader, parameters, Reporter.NO_OP);
     }
 
-    public static <T> List<LfNetwork> load(T network, LfNetworkLoader<T> networkLoader, SlackBusSelector slackBusSelector, Reporter reporter) {
```
annetill, line 526: This removal is okay for me, to be confirmed by @geofjamg

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
```diff
@@ -1139,4 +1133,62 @@ void testPhaseTapChangerActionThreeWindingsTransformer() {
                 operatorStrategies1, actions1, Reporter.NO_OP));
         assertEquals("Phase tap changer tap connection action: only one tap in branch PS1_leg_1", exception.getCause().getMessage());
     }
+
+    @Test
+    void testActionOnRetainedPst() {
```
annetill, line 1138: @Hadrien-Godard you need a test for RTC...

annetill (approved):

src/main/java/com/powsybl/openloadflow/network/LfNetworkLoader.java
```diff
@@ -7,6 +7,7 @@
 package com.powsybl.openloadflow.network;
 
 import com.powsybl.commons.reporter.Reporter;
+import com.powsybl.openloadflow.network.impl.LfTopoConfig;
```
geofjamg, line 10: We have a dependency in the wring direction, we maybe should move LfTopoConfig in the network package

src/main/java/com/powsybl/openloadflow/network/impl/LfTopoConfig.java
```diff
@@ -22,6 +22,10 @@ public class LfTopoConfig {
 
     private final Set<String> busIdsToLose = new HashSet<>();
 
+    private final Set<String> branchIdsWithPtcToRetain = new HashSet<>();
```
geofjamg, line 25: For later, but maybe we should find a better name for LfTopoConfig once this will be added?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -268,6 +270,30 @@ protected static void findAllSwitchesToOperate(Network network, List<Action> act
                 });
     }
 
+    protected static void findAllPtcToOperate(List<Action> actions, LfTopoConfig topoConfig) {
+        for (Action action : actions) {
+            if (action.getType().equals("PHASE_TAP_CHANGER_TAP_POSITION")) {
```
geofjamg, line 275: We have a constant for that  PhaseTapChangerTapPositionAction.NAME

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -268,6 +270,30 @@ protected static void findAllSwitchesToOperate(Network network, List<Action> act
...
+                        side -> topoConfig.addBranchIdsWithPtcToRetain(LfLegBranch.getId(side, ptcAction.getTransformerId())), // T3WT
+                        () -> topoConfig.addBranchIdsWithPtcToRetain(ptcAction.getTransformerId()) // T2WT
+                );
+            }
+        }
+    }
+
+    protected static void findAllRtcToOperate(List<Action> actions, LfTopoConfig topoConfig) {
+        for (Action action : actions) {
+            if (action.getType().equals("RATIO_TAP_CHANGER_TAP_POSITION")) {
```
geofjamg, line 287: Same

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/network/impl/LfTopoConfig.java
annetill: We can do it now, shunt compensator action is coming just after, what do you suggest ?

geofjamg: Could be a `LfNetworkLoadingParameters`

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
geofjamg: yes

src/main/java/com/powsybl/openloadflow/network/impl/LfTopoConfig.java
annetill: I think we can leave it like this and re-think about it with shunt compensator action.

geofjamg (approved):

Issue #818, opened 2023-07-26
Exclude generators below minP from slack distribution


jeandemanged:
### Describe the current behavior

generators below minP are excluded from slack distribution from slack distribution only if minP > 0, see [here](https://github.com/powsybl/powsybl-open-loadflow/blob/v1.2.3/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java#L339)

In case of a pumping generator below minP with minP zero or negative, OLF with slack distribution enabled sets the generator at minP, the difference is then distributed to other generators.

### Describe the expected behavior

generators below minP should be excluded from slack distribution always and left unchanged at targetP

### Describe the motivation

Not breaking initial generation pattern

### Extra Information

OLF v1.2.3

PR #868, opened 2023-09-29
Exclude generators below zero or negative minP from slack distribution

jeandemanged:

#818 

No


geofjamg:
@jeandemanged I think we should add a parameter to configure this behaviour. This ugly heuristic I guess is here to get same result as Hades2 (@annetill to confirm?)

annetill (approved):
Approved without any parameter.

geofjamg:
> @jeandemanged I think we should add a parameter to configure this behaviour. This ugly heuristic I guess is here to get same result as Hades2 (@annetill to confirm?)

It looks like it is not :-)

geofjamg (approved):

PR #869, opened 2023-09-29
[WIP] Option to ignore active power limits in slack distribution

jeandemanged:

No

Feature

With slack distribution enabled, with distribution on generators
- generators with targetP outside minP/maxP are always excluded from slack distribution
- generators minP/maxP are always enforced for generators participating in slack distribution

Add a new option `useActiveLimits`, similar to `useReactiveLimits`.
When set to `true` (default), apply current behavior.
When set to `false`:
- generators with targetP outside minP/maxP will be allowed to participate in slack distribution
- generators minP/maxP are ignored in slack distribution

No

**Other information**:

None


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
```diff
@@ -301,40 +301,47 @@ public void setParticipating(boolean participating) {
...
         if (Math.abs(targetP) < POWER_EPSILON_SI) {
-            LOGGER.trace("Discard generator '{}' from active power control because targetP ({}) equals 0",
+            LOGGER.trace("Discard generator '{}' from active power control because targetP ({} MW) equals 0",
                     generatorId, targetP);
             if (report != null) {
                 report.generatorsDiscardedFromActivePowerControlBecauseTargetEqualsToZero++;
             }
             participating = false;
         }
+        if (maxP > plausibleActivePowerLimit) {
```
annetill, line 314: I understand your point of view and why you prefer to keep this check even if limits are disabled but... it makes the feature less comprehensive.

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowParameters.java
```diff
@@ -23,6 +23,8 @@ public class DcLoadFlowParameters extends AbstractLoadFlowParameters {
 
     private final boolean distributedSlack;
 
+    private final boolean useActiveLimits;
```
geofjamg, line 26: You don't need to have it here, as it is already in `LfNetworkParameters` which is embedded in DcLoadLowParameter (though the `AbstractLoadFlowParameters`)

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
geofjamg: I think we need this check otherwise some active power distribution mode based on pmax won't work anymore. An alternative would be to forbid these active power distribution modes in case of useActiveLimits to false but this was not the idea of this feature.

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowParameters.java
jeandemanged: wow I did completely messed up here ... thanks

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
annetill: @jeandemanged why do you say "factors for fictitious elements."? Ficticious?

geofjamg (approved):

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
jeandemanged: Some TSOs provide provide generators with very high Pmax in their IGMs, equivalents would be a better term instead of fictitous

PR #871, opened 2023-09-29
Fix reporting of distributed slack active power

jeandemanged:

No

Bugfix

Distributed active power is currently reported as the difference of slack injection between first NR and last NR. This works only approximately because of change in losses that can occur between first and last NR. This is especially visible when slack distribution is disabled, we get a non-zero distributed active power reported in the results - while no distribution was done.

Distributed slack outer loop now keeps track of the distributed active power.

No

**Other information**:

None


src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -123,11 +122,15 @@ public AcLoadFlowResult run() {
                 .map(outerLoop -> Pair.of(outerLoop, new AcOuterLoopContext(context.getNetwork())))
                 .collect(Collectors.toList());
 
+        Optional<DistributedSlackContextData> distributedSlackContextDataOptional = Optional.empty();
         // outer loops initialization
         for (var outerLoopAndContext : outerLoopsAndContexts) {
             var outerLoop = outerLoopAndContext.getLeft();
             var outerLoopContext = outerLoopAndContext.getRight();
             outerLoop.initialize(outerLoopContext);
+            if (outerLoop instanceof DistributedSlackOuterLoop) {
```
geofjamg, line 131: Maybe it would be simpler to do this after the NR run when we iterate over outer loops for calling cleanup method ?

jeandemanged: right, thanks

geofjamg (approved):

PR #872, opened 2023-10-01
Secondary voltage control, coupled version

geofjamg:

No

Feature

This the the "coupled" implementation where we try to reach the pilot point target and align generators reactive power in a single calculation. The aim is to reduce the number iterations needed to converge.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/LfSecondaryVoltageControl.java
```diff
@@ -41,7 +42,45 @@ public LfBus getPilotBus() {
         return pilotBus;
     }
 
-    public Set<LfBus> getControlledBuses() {
-        return controlledBuses;
+    public Set<GeneratorVoltageControl> getGeneratorVoltageControls() {
+        return generatorVoltageControls;
+    }
+
+    private static List<LfBus> findControllerBuses(LfBus controlledBus) {
```
annetill, line 49: Do we have forgotten `MAIN` status filtering?

src/main/java/com/powsybl/openloadflow/ac/outerloop/SecondaryVoltageControlOuterLoop.java
```diff
@@ -256,85 +217,93 @@ private static double calculatePilotPointDv(LfSecondaryVoltageControl secondaryV
...
             }
+        }
 
-            var controllerBusIndex = buildBusIndex(controllerBuses);
+        if (adjustedZoneNames.isEmpty()) {
+            return Collections.emptyList();
+        }
 
-            DenseMatrix a = createA(controllerBuses, controllerBusIndex);
+        List<LfBus> allControllerBuses = secondaryVoltageControls.stream()
```
annetill, line 243: This is not clear why you leave that here... and why you don't put this method in the `LfSecondayVoltageControl` class.

src/main/java/com/powsybl/openloadflow/network/LfSecondaryVoltageControl.java
geofjamg: I think the check is done when listing the controlled buses

annetill (approved):

PR #879, opened 2023-10-09
Fix incremental outer loops

annetill:

No.

We consider hidden voltage controls in shunt and transformer incremental outer loops. We introduce a dedicated `isHidden` for the outer loops only.

To be precise, shunt and transformer voltage controls can be considered even if hidden by a generator (or a transformer).
In case of multiple voltage control types, we use as targetV, the targetV of the highest priority voltage control.

If we have heterogeneous voltage controls for the same bus, we only take into account the one with highest priority.

Now heterogeneous voltage controls are taken into account for a single bus if incremental shunt or transformer voltage control mode is used.

No

**Other information**:

n/a


Hadrien-Godard (approved):
Do not forget to delete test and changes in pom.xml

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalTransformerVoltageControlOuterLoop.java
```diff
@@ -56,6 +57,24 @@ public String getName() {
...
+    public static List<LfBranch> getControllerElements(LfNetwork network) {
+        return network.getBuses().stream()
+                .filter(bus -> bus.isVoltageControlled(VoltageControl.Type.TRANSFORMER))
+                .filter(bus -> bus.getVoltageControl(VoltageControl.Type.TRANSFORMER).orElseThrow().getMergeStatus() == VoltageControl.MergeStatus.MAIN)
+                .filter(bus -> !bus.getVoltageControl(VoltageControl.Type.TRANSFORMER).orElseThrow().isHidden(true))
+                .flatMap(bus -> bus.getTransformerVoltageControl().orElseThrow().getMergedControllerElements().stream())
+                .filter(Predicate.not(LfBranch::isDisabled))
+                .collect(Collectors.toList());
+    }
+
```
annetill, line 77: You have forgotten to change the list of controller elements to build the context.

Hadrien-Godard (approved):

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalShuntVoltageControlOuterLoop.java
```diff
@@ -47,13 +48,31 @@ public String getName() {
...
+
+    public static List<LfShunt> getControllerElements(LfNetwork network) {
+        return network.getBuses().stream()
+                .filter(bus -> bus.isVoltageControlled(VoltageControl.Type.SHUNT))
+                .filter(bus -> bus.getVoltageControl(VoltageControl.Type.SHUNT).orElseThrow().getMergeStatus() == VoltageControl.MergeStatus.MAIN)
+                .filter(bus -> !bus.getVoltageControl(VoltageControl.Type.SHUNT).orElseThrow().isHidden(true))
+                .flatMap(bus -> bus.getShuntVoltageControl().orElseThrow().getMergedControllerElements().stream())
+                .filter(Predicate.not(LfShunt::isDisabled))
+                .collect(Collectors.toList());
+    }
```
jeandemanged, line 67: These two methods `getControlledBuses `and `getControllerElements` are scanning all buses, I am a bit worried about performance consequences for large networks, because they are called in the outerloop `check `method.

I might be better to store the "static/unchanging" part of this data in the outerloop context at outerloop initialization time, i.e. store the list of shunt voltage controlled buses in the outerloop context.

Then at `check` time, use this unchanging list and apply additional filters with the "dynamic" part: isHidden & isDisabled.


to be fair: not sure if an issue, it was already the case before this PR...

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalTransformerVoltageControlOuterLoop.java
```diff
@@ -56,14 +57,32 @@ public String getName() {
...
+    public static List<LfBranch> getControllerElements(LfNetwork network) {
+        return network.getBuses().stream()
+                .filter(bus -> bus.isVoltageControlled(VoltageControl.Type.TRANSFORMER))
+                .filter(bus -> bus.getVoltageControl(VoltageControl.Type.TRANSFORMER).orElseThrow().getMergeStatus() == VoltageControl.MergeStatus.MAIN)
+                .filter(bus -> !bus.getVoltageControl(VoltageControl.Type.TRANSFORMER).orElseThrow().isHidden(true))
+                .flatMap(bus -> bus.getTransformerVoltageControl().orElseThrow().getMergedControllerElements().stream())
+                .filter(Predicate.not(LfBranch::isDisabled))
+                .collect(Collectors.toList());
+    }
```
jeandemanged, line 76: Same comment as for `IncrementalShuntVoltageControlOuterLoop `.

Also some code mutualization between `IncrementalShuntVoltageControlOuterLoop` 
 and `IncrementalTransformerVoltageControlOuterLoop ` would be welcomed here.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
jeandemanged, line 1: I would have expected a test similar to `AcLoadFlowShuntTest#testIncrementalVoltageControlWithGenerator` here. E.g. a network with a generator and it's step up transformer both regulating the same HV bus.

src/test/java/com/powsybl/openloadflow/network/ShuntNetworkFactory.java
```diff
@@ -125,4 +125,47 @@ public static Network createWithTwoShuntCompensators() {
...
+                .setTargetP(0)
+                .setTargetV(393)
+                .setMinP(0)
+                .setMaxP(150)
+                .setVoltageRegulatorOn(true)
+                .add();
+        return network;
+    }
+
+    public static Network createWithGeneratorAndShuntNonImpedant() {
```
jeandemanged, line 145: Not really related to this PR, but this network is causing the following ERROR log to be issued:
`ERROR c.p.o.n.i.LfNetworkLoaderImpl | Controller shunt vl3_0_controller_shunt_compensators is already in a shunt voltage control. The second controlled bus vl3_0 is ignored`
Which is imo a false positive that should be fixed.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -14,9 +14,7 @@
 import com.powsybl.math.matrix.DenseMatrixFactory;
 import com.powsybl.openloadflow.OpenLoadFlowParameters;
 import com.powsybl.openloadflow.OpenLoadFlowProvider;
-import com.powsybl.openloadflow.network.HvdcNetworkFactory;
-import com.powsybl.openloadflow.network.SlackBusSelectionMode;
-import com.powsybl.openloadflow.network.VoltageControlNetworkFactory;
+import com.powsybl.openloadflow.network.*;
```
jeandemanged, line 17: ?

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalShuntVoltageControlOuterLoop.java
annetill: Note that the MAIN status is dynamic too and cannot be computed in the beginning.

src/test/java/com/powsybl/openloadflow/network/ShuntNetworkFactory.java
annetill: "imo"? Before the security analysis where we create 2 shunt voltage controls, a MAIN one and a DEPENDENT one, we run 2 load flows. In these two load flows, a single shunt voltage control with controlled bus b3 is created. This control has two shunt references in it.

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalShuntVoltageControlOuterLoop.java
jeandemanged: thanks, that's still not that bad since less that half of the buses will have controls of given type

jeandemanged (approved):

PR #888, opened 2023-10-28
Refactor propagated contingency

geofjamg:

No

Refactoring

Propagated contingency API is badly designed because we need 2 phases, one to create and another one to complete the data knowing the breaker mode.

We don't need anymore the breaker mode and we can remove the complete phase.
 
**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingencyCreationParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingencyCreationParameters.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) 2023, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.openloadflow.network.impl;
+
+/**
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ */
+public class PropagatedContingencyCreationParameters {
+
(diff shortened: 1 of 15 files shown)


src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
```diff
@@ -251,11 +251,6 @@ private List<OperatorStrategyResult> createOperatorStrategyResults(DcSecurityAna
                 .setCacheEnabled(false); // force not caching as not supported in secu analysis
 
         try (LfNetworkList lfNetworks = Networks.load(network, dcParameters.getNetworkParameters(), topoConfig, Reporter.NO_OP)) {
-
```
annetill, line 254: I am worried about your change using `LoadFlowParameters` here, and for DC sensitivity analysis. Because we force for these use cases no shunt voltage control and no HVDC AC emulation. But here you can configure `true` without any issue for a DC load flow. But for the creation of the propagated contingency, it will be wrong.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingencyCreationParameters.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.powsybl.loadflow.LoadFlowParameters;
+import com.powsybl.openloadflow.network.LfNetworkParameters;
+
+/**
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ */
+public class PropagatedContingencyCreationParameters {
+
+    public static final boolean CONTINGENCY_PROPAGATION_DEFAULT_VALUE = false;
```
annetill, line 17: The default value is in `OpenSecurityAnalysisParameters` and is set to `true`.

annetill (approved):

PR #895, opened 2023-11-08
Improve performance of incremental outer loops

caioluke:

Performance enhancement

We currently calculate sensitivity for all controllers of the network, even though we won't need them all.

Sensitivities are only calculated for the necessary controllers (outside target/deadband).

No

**Other information**:


src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalShuntVoltageControlOuterLoop.java
```diff
@@ -86,7 +88,27 @@ private static int[] createControllerShuntIndex(LfNetwork network, List<LfShunt>
         private static DenseMatrix calculateSensitivityValues(List<LfShunt> controllerShunts, int[] controllerShuntIndex,
                                                               EquationSystem<AcVariableType, AcEquationType> equationSystem,
                                                               JacobianMatrix<AcVariableType, AcEquationType> j) {
-            DenseMatrix rhs = new DenseMatrix(equationSystem.getIndex().getSortedEquationsToSolve().size(), controllerShunts.size());
+            int nRows = equationSystem.getIndex().getSortedEquationsToSolve().size();
+            int nColumns = controllerShunts.size();
+
+            if (nColumns == 1) {
```
geofjamg, line 94: What is the purpose of this special case ? It does not optimize anything. Calling solve on a vector or on a matrix with one column is exactly the same.

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalTransformerVoltageControlOuterLoop.java
```diff
@@ -85,7 +83,27 @@ public SensitivityContext(LfNetwork network, List<LfBranch> controllerBranches,
         private static DenseMatrix calculateSensitivityValues(List<LfBranch> controllerBranches, int[] controllerBranchIndex,
                                                               EquationSystem<AcVariableType, AcEquationType> equationSystem,
                                                               JacobianMatrix<AcVariableType, AcEquationType> j) {
-            DenseMatrix rhs = new DenseMatrix(equationSystem.getIndex().getSortedEquationsToSolve().size(), controllerBranches.size());
+            int nRows = equationSystem.getIndex().getSortedEquationsToSolve().size();
+            int nColumns = controllerBranches.size();
+
+            if (nColumns == 1) {
```
geofjamg, line 89: Same

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalShuntVoltageControlOuterLoop.java
caioluke: That's true, just fixed it

geofjamg:
@caioluke is this optimization based on some performance profiles? Do you have observed in a real case a significant performance improvement?

jeandemanged:
@geofjamg yes we observed the issue on large networks.
Didn't check code yet but just tested it today.
Performance improvement is quite significative.

Flamegraph before (outerloop is taking more time than the NR ...):
image

Flamegraph after:
image

solution "path" unchanged but much faster (left is before fix, right after fix):
image

geofjamg:
Indeed.
Do you have the "solve2" individual execution time (before) and how much time it is called?
Also are you able with your profiler to get JNI native function profiling (IntelliJ profiler can do it by changing the option 'Collect native calls") ?

geofjamg:
I summarize the discussion I had with @caioluke:
 - it is indeed a good idea to pre-filter controllers that are outside of the dead band
   - it really accelerates the performance, because the more the incremental outerloops are running less there is controller outside of dead band. So firsts outer loops are slow and last are very fast.
   - Even if faster, we still see the sensi calculation of incremental outer loops in profiling so there is room to more optimization.
 -  The REAL PROBLEM which is the one to solve is that we are solving a very very sparse second member. Second member is made with a matrix that has only one non zero value per column (and the number of row is the size of the system). KLU library only supports dense second member (this is the normal use so it makes sense). So in our very special way to use KLU for sensi analysis, the KLU_solve and KLU_tsolve functions are very inefficients. Thanks @caioluke @jeandemanged for pointing this I was not aware of it.
 - A solution would be to re-implement our own KLU_solve and KLU_tsolve function. Looking at https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/dev/KLU/Source/klu_sort.c for instance, the code is quite simple and short and could maybe be copy/pasted to work on a sparse second member (instead of a B[], a pair B_values[] and B_Index[]). Performance would be much faster, only processing non zero values. 

@jeandemanged @Hadrien-Godard what do you think?

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalShuntVoltageControlOuterLoop.java
```diff
@@ -156,20 +158,41 @@ public OuterLoopStatus check(AcOuterLoopContext context, Reporter reporter) {
...
+
+        // check which shunts are not within their deadbands
+        List<LfShunt> controllerShuntsOutsideOfDeadband = new ArrayList<>();
+        List<LfBus> controlledBusesOutsideOfDeadband = new ArrayList<>();
+        controlledBuses.forEach(controlledBus -> {
+            ShuntVoltageControl voltageControl = controlledBus.getShuntVoltageControl().orElseThrow();
+            double diffV = controlledBus.getHighestPriorityMainVoltageControl().orElseThrow().getTargetValue() - voltageControl.getControlledBus().getV();
+            double halfTargetDeadband = getHalfTargetDeadband(voltageControl);
+            if (Math.abs(diffV) > halfTargetDeadband) {
+                List<LfShunt> controllers = voltageControl.getMergedControllerElements().stream()
```
geofjamg, line 171: We have a difference here in the way we get controllers from a controlled. In previous code, we use `IncrementalContextData.getControllerElements` which has a filter on voltage control type. Is it equivalent?

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalTransformerVoltageControlOuterLoop.java
```diff
@@ -204,43 +205,64 @@ public OuterLoopStatus check(AcOuterLoopContext context, Reporter reporter) {
...
+        List<String> controlledBusesWithAllItsControllersToLimit = new ArrayList<>();
+
+        controlledBusesOutsideOfDeadband.forEach(controlledBus -> {
+            TransformerVoltageControl voltageControl = controlledBus.getTransformerVoltageControl().orElseThrow();
+            double diffV = getDiffV(voltageControl);
+            double halfTargetDeadband = getHalfTargetDeadband(voltageControl);
+            if (Math.abs(diffV) <= halfTargetDeadband) { // buses have already been filtered
+                throw new PowsyblException("Bus should have been outside of deadband");
+            }
+            List<LfBranch> controllers = voltageControl.getMergedControllerElements().stream()
```
geofjamg, line 249: Same

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalShuntVoltageControlOuterLoop.java
caioluke: Yes, it is equivalent. I directly call the method for the specific type of voltage control: `getShuntVoltageControl` or `getTransformerVoltageControl`. And I also filter out the disabled controllers.

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalTransformerVoltageControlOuterLoop.java
caioluke: Same

src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalTransformerVoltageControlOuterLoop.java
```diff
@@ -191,11 +201,26 @@ private boolean adjustWithSeveralControllers(List<LfBranch> controllerBranches,
     }
 
     private static double getDiffV(TransformerVoltageControl voltageControl) {
-        double targetV = voltageControl.getTargetValue();
+        double targetV = voltageControl.getControlledBus().getHighestPriorityMainVoltageControl().orElseThrow().getTargetValue();
```
annetill, line 204: @jeandemanged and @geofjamg bug fix here. We have to check the voltage difference against the voltage control of highest priority, most of the time those of the generator and not those of the transformer.

annetill (approved):

annetill:
Maybe a final performance test is needed before merging and a unit test that shows the bug.

jeandemanged (approved):

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerControlTest.java
```diff
@@ -363,6 +363,85 @@ void remoteVoltageControlT2wtTest2() {
...
+
+        parameters.setTransformerVoltageControlOn(true);
+        parametersExt.setTransformerVoltageControlMode(OpenLoadFlowParameters.TransformerVoltageControlMode.INCREMENTAL_VOLTAGE_CONTROL);
+
+        t2wt.getRatioTapChanger()
+                .setTargetDeadband(0)
+                .setRegulating(true)
+                .setTapPosition(0)
+                .setRegulationTerminal(line34.getTerminal2())
+                .setTargetV(33.0);
```
annetill, line 408: Change this targetV to 30.0 to show the benefit of the bug fix.

geofjamg (approved):

jeandemanged (approved):
all good, thanks

other improvements (see [here](https://github.com/powsybl/powsybl-open-loadflow/pull/895#issuecomment-1807143945)) would be in other PRs

PR #899, opened 2023-11-16
Generator reference priorities

jeandemanged:

No

Feature

Angle reference bus is always the (first) slack bus.

Leverage https://github.com/powsybl/powsybl-core/pull/2763

- Only for generator terminals as a first step
- Only in LoadFlow, not used in SA nor sensi
- new OLF parameters:
  - `referenceBusSelectionMode` with possible values
    - `FIRST_SLACK` (default, =current behavior): angle reference bus selected as the first slack bus among potentially multiple slacks
    - `GENERATOR_REFERENCE_PRIORITY`: angle reference bus selected from generator reference priorities
  - `writeReferenceTerminals` boolean to have reference generator(s) terminals written to the ReferenceTerminals iIDM extension

No

**Other information**:


geofjamg (approved):

geofjamg (approved):
@jeandemanged we can see in your PR that this feature is implemented for generators. It would be feasible to do it also for any injections and busbar sections but very hard for branches and probably useless. Why not in that case, restrict IIDM extension `ReferencePriority` and `ReferenceTerminal` to only terminal types that we want to support?

jeandemanged:
> @jeandemanged we can see in your PR that this feature is implemented for generators. It would be feasible to do it also for any injections and busbar sections but very hard for branches and probably useless. Why not in that case, restrict IIDM extension `ReferencePriority` and `ReferenceTerminal` to only terminal types that we want to support?

Fully agree, we can indeed restrict in core

PR #903, opened 2023-11-21
Fix utility method to create generator in unit tests 

annetill:

No, just in my mind... we have a method to create generators for our small network that automatically set `maxP` to `targetP`. The consequence is that we often fall in a failure to distribute slack. Most of the time, there is a "fix" in the unit test, but it has been forgotten. A relevant unit test must have it slack all distributed or all keep in slack.

I have put `maxP` equal to 2 * `targetP`. Note that sometimes, it is not enough to have a good slack distribution (when we loose generators in contingencies for example).

**Other information**:

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/network/AbstractLoadFlowNetworkFactory.java
+++ b/src/test/java/com/powsybl/openloadflow/network/AbstractLoadFlowNetworkFactory.java
@@ -63,7 +63,7 @@ public abstract class AbstractLoadFlowNetworkFactory {
                 .setConnectableBus(b.getId())
                 .setEnergySource(EnergySource.OTHER)
                 .setMinP(0)
-                .setMaxP(p)
+                .setMaxP(2 * p)
                 .setTargetP(p)
                 .setTargetV(v)
                 .setVoltageRegulatorOn(true)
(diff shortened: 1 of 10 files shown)


src/test/java/com/powsybl/openloadflow/network/AbstractLoadFlowNetworkFactory.java
```diff
@@ -63,7 +63,7 @@ protected static Generator createGenerator(Bus b, String id, double p, double v)
                 .setConnectableBus(b.getId())
                 .setEnergySource(EnergySource.OTHER)
                 .setMinP(0)
-                .setMaxP(p)
+                .setMaxP(2 * p)
```
caioluke, line 66: Why 2 times p and not... 10 times p?
If we wanted to be sure that slack will be distributed, couldn't we set maxP after the whole network is built to something like maxP = sum(targetPs + loadP0s)?

caioluke (approved):

src/test/java/com/powsybl/openloadflow/network/AbstractLoadFlowNetworkFactory.java
annetill: Yes, it is a choice and I don't know if it is relevant. As you can see, in some tests it is not enough. @jeandemanged any opinion?

jeandemanged: 2* is enough, I'd say it is the unit test's author responsibility to verify it's test is **really** a **unit** test, i.e. not relying too much of many OLF features within a single test.

jeandemanged (approved):
👍

jeandemanged (approved):

PR #919, opened 2023-12-02
Improve logging, reporting and LF results for components without generators

jeandemanged:

No

Feature

Synchronous Components (islands) without any generator connected are causing many ERROR logs to be raised, and many report entries created which are just noise.
Hundreds of errors are logged & reported on real case large networks when OLF configured to solve all components which is quite inconvenient, and not really an error.

Network post-loading validation now:
- checks for islands without any generator,
- and for AC models, checks additionally for islands with generators but without voltage control enabled.
Only relevant logs & reports are now produced.

For islands without any generator, only one log & one report entry is produced reporting the total number of islands where no calculation will be made.

With new LoadFlowResult API (see https://github.com/powsybl/powsybl-core/pull/2767) and #918 we now are able to report new NO_CALCULATION status for "dead" islands without any generator.

> ⚠️ **Important Notes**:
> - **Functional change:** DC Load Flow with this change will **not** solve islands without generators anymore. It was solving them before - in my opinion this is not a regular use case.
> - AC Load Flow: islands with generators but no voltage regulation are **not** considered dead, a FAILED status is reported in the case of generators present but no voltage regulation - no change of functionality here.

- [x] No

n/a

n/a

**Other information**:

n/a


src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -619,17 +651,29 @@ public static <T> List<LfNetwork> load(T network, LfNetworkLoader<T> networkLoad
...
+                    LOGGER.info("Network {} is invalid, no calculation will be done", lfNetwork);
+                    // we want to report this
+                    lfNetwork.setReportNode(Reports.createLfNetworkReportNode(reportNode, lfNetwork.getReportNode(), lfNetwork.getNumCC(), lfNetwork.getNumCC()));
+                }
+                case INVALID_NO_GENERATOR -> deadComponentsCount++; // will be reported later on altogether
             }
         }
+        if (deadComponentsCount > 0) {
+            Reports.reportComponentsWithoutGenerators(reportNode, deadComponentsCount);
+            LOGGER.info("No calculation will be done on {} network(s) that have have no generators", deadComponentsCount);
```
caioluke, line 675: "have" is written twice

caioluke (changes requested):
Minor typo

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -580,6 +597,21 @@ public void fix(boolean minImpedance, double lowImpedanceThreshold) {
     }
 
     private void validateBuses(LoadFlowModel loadFlowModel, ReportNode reportNode) {
+        // DC or AC, if no generator, network is dead
```
jeandemanged, line 600: we might want to add an option to allow computing islands without generators in DC

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -248,12 +256,22 @@ public static ReportNode createLoadFlowReporter(ReportNode reportNode, String ne
...
+    }
+
+    public static ReportNode createLfNetworkReportNode(ReportNode reportNode, ReportNode lfNetworkReportNode, int networkNumCc, int networkNumSc) {
+        ReportNode newReportNode = reportNode.newReportNode()
                 .withMessageTemplate("lfNetwork", "Network CC${networkNumCc} SC${networkNumSc}")
                 .withUntypedValue(NETWORK_NUM_CC, networkNumCc)
                 .withUntypedValue(NETWORK_NUM_SC, networkNumSc)
                 .add();
+        newReportNode.include(lfNetworkReportNode);
+        return newReportNode;
```
jeandemanged, line 274: this will need to be revisited at next powsybl-core bump because of https://github.com/powsybl/powsybl-core/pull/3007

olperr1: This is done in #1034.

PR #921, opened 2023-12-04
Fix NPE on VSC without a HVDC line

geofjamg:

No

Bug fix

We get a NPE when HVDC emulation is activated and we have a VSC or LCC converter station without a HVDC line.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfVscConverterStationImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfVscConverterStationImpl.java
@@ -6,6 +6,7 @@
  */
 package com.powsybl.openloadflow.network.impl;
 
+import com.powsybl.iidm.network.HvdcLine;
 import com.powsybl.iidm.network.ReactiveLimits;
 import com.powsybl.iidm.network.VscConverterStation;
 import com.powsybl.iidm.network.util.HvdcUtils;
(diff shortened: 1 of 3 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfVscConverterStationImpl.java
```diff
@@ -78,12 +79,14 @@ public double getTargetQ() {
 
     @Override
     public double getMinP() {
-        return -getStation().getHvdcLine().getMaxP() / PerUnit.SB;
+        HvdcLine hvdcLine = getStation().getHvdcLine();
+        return hvdcLine != null ? -hvdcLine.getMaxP() / PerUnit.SB : -Double.MAX_VALUE;
```
annetill, line 83: It is not needed as you cannot have a VSC without any HVDC line in network loading.

geofjamg: I had a failing test without adding it.

annetill: ok, see #691 for a better solution about hvdc line.

annetill (approved):

PR #937, opened 2023-12-17
Fix contingency propagation

geofjamg:

No

Bug fix

- [ ] Yes
- [ ] No

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -450,10 +460,10 @@ public Optional<LfContingency> toLfContingency(LfNetwork network) {
                         boolean otherSideConnected;
                         if (branch.getBus1() == busToLost) {
                             otherSideBus = branch.getBus2();
-                            otherSideConnected = branch.isConnectedSide2();
+                            otherSideConnected = branch.isConnectedSide2() && isConnectedAfterContingencySide2(branchesToOpen, branch);
```
rosiereflo, line 463: Why don't we update the network g=following the contingency?

annetill: This is done in the `apply` method of  `LfContingency`.

annetill: Yes, exactly. Note that this method `apply` is not called always in some DC sensitivity analyses, that is why we have a lot of treatments in the `PropagatedContingency` class.

annetill:
Not really easy to add a unit test, no clear issue (throw for example) in real cases I tested since yesterday.

annetill (approved):
Approved

annetill (approved):

PR #941, opened 2024-01-02
Fix sharing of reactive power with for multiple generators with ReactivePowerControl

caioluke:

Bug fix

Reactive power is always equally split between generators on a bus controlling the reactive power on a branch. Thus, ignoring the reactive keys. 

The reactive keys will not only work for splitting the reactive power between different buses but also for different generators on the same bus.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
@@ -518,10 +518,13 @@ public abstract class AbstractLfBus extends AbstractElement implements LfBus {
     }
 
     private static ToDoubleFunction<String> splitDispatchQWithReactiveKeys(List<LfGenerator> generatorsWithControl, double qToDispatch) {
-        double sumQkeys = 0.0;
+        double sumQkeys = 0;
         for (LfGenerator generator : generatorsWithControl) {
             double qKey = generator.getRemoteControlReactiveKey().orElseThrow();
             sumQkeys += qKey;
+        }
+        if (sumQkeys == 0) { // to avoid division by zero
+            sumQkeys = 1;
         }
 
         Map<String, Double> qToDispatchByGeneratorId = new HashMap<>(generatorsWithControl.size());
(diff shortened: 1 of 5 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -538,6 +586,11 @@ protected static double dispatchQ(List<LfGenerator> generatorsWithControl, boole
             case K_EQUAL_PROPORTION -> allGeneratorsHavePlausibleReactiveLimits(generatorsWithControl)
                     ? splitDispatchQWithEqualProportionOfK(generatorsWithControl, qToDispatch)
                     : splitDispatchQ(generatorsWithControl, qToDispatch); // fallback to q dispatch
+            case REACTIVE_KEYS -> allGeneratorsHaveReactiveKeys(generatorsWithControl)
```
annetill, line 589: I am not convinced by your fix. Indeed, during calculation, if reactive keys are present, we take them (`createReactiveKeys`). If not, we have two fall backs. `K_EQUAL_PROPORTION` is a bit special and must not be changed, but I think that the good solution is to adapt the `Q_EQUAL_PROPORTION` that seems to be too simple compared to what we do for calculation. To be confirmed by @geofjamg, but I think that you have to follow the same logic as the logic we have for creation.

caioluke: I agree with you, there is definetely something to discuss about my fix.
The thing is that reactive keys for calculations created by `createReactiveKeys` are used to split reactive power between **buses**. Whereas here `reactivePowerDispatchMode` controls how reactive power is split within a bus, between **generators**. I don't know if there are cases which we would like to split reactive power between buses in a certain way and in another way between generators...

annetill: Yes of course, but as you said, we must be coherent...

annetill (approved):
Thanks a lot!

PR #943, opened 2024-01-03
Added support of action to reconnect branch at both side action 

annetill:

No. Only line opening at both sides is supported as remedial action in main branch. An error message throws if a user wants to simulate a one side modification or a line re-connection at both sides. During this feature supports, I rework on the action in `powsybl-core` because it is not adapted to open/close at one side only (https://github.com/powsybl/powsybl-core/pull/2843/).

For the moment, only the support of a remedial action that connects at both side a line that was totally disconnected.

- [ ] Yes
- [ ] No

**Other information**:


src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -290,6 +290,20 @@ protected static void findAllRtcToOperate(List<Action> actions, LfTopoConfig top
         }
     }
 
+    protected static void findAllBranchToOpen(Network network, List<Action> actions, LfTopoConfig topoConfig) {
```
rosiereflo, line 293: If I understand well, should be named findAllBranchToClose?

annetill: Ah yes, thanks!

src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
```diff
@@ -124,15 +124,23 @@ private static void retainAndCloseNecessarySwitches(Network network, LfTopoConfi
                 .forEach(sw -> sw.setRetained(true));
 
         topoConfig.getSwitchesToClose().forEach(sw -> sw.setOpen(false)); // in order to be present in the network.
+        topoConfig.getBranchIdsToClose().stream().map(id -> network.getBranch(id)).forEach(branch -> {
```
obrix, line 127: .map(id -> network.getBranch(id)) ---> .map(network::getBranch) (SONAR)

obrix (commented):
Looks good to me, just a small fix needed for sonar.

rosiereflo (approved):

jeandemanged (approved):
I'd just rename the PR to "Added support of branch re-connection action" if you agree

PR #961, opened 2024-01-22
Loss of a VSC converter station after contingency

annetill:

Yes, with #957. This PR follows the analysis made in that PR and extracts a smaller part. The aim of this current PR is to handle the loss of a VSC converter station. As there is no VSC converter station contingency, this happens when we loose a bus that contains a VSC converter station. Two cases:
- The other converter station is not in the `LfNetwork`: the component with the lost VSC converter station is isolated (and not computed), so is the other side. This case if supported in main branch. Note that it could not be in AC emulation (both converter stations must be in the same `LfNetwork`).
- The other converter station is also in the `LfNetwork`: this case is not supported well. 1) In case of an AC emulation operation, as one side bus is isolated (and not computed), we must disabled the AC emulation equations. 2) In case of a set point operation, in main branch, the other converter station continues to produce or consumes active power.

Solution proposal:

1) We disable the AC emulation equations and apply a set point of zero on the hvdc line. The other converter station keeps its voltage control capabilities. Putting a set point to zero can be discuss.
2) We put a set point of zero on the hvdc line. The other converter station keeps its voltage control capabilities. Putting a set point to zero can be discuss. Indeed, if the lost component is big enough with generators and loads, depending on rectifier and inverter configuration, active power flow can flow. Update: the simple criteria of loosing only one bus is kept.

Note that in that proposal, we change something important: if both VSC converter stations are in the `LfNetwork` at loading, we create a `LfHvdc`, even if not operated in AC emulation.    

LCC converter stations are not supported in that fix.

N case, where for an hvdc line, both converter stations are not in the same synchronous/connected component is not handled. Active power flow is not fixed. Note that in that case, the criteria is not obvious.

Update: what is not supported is in case of bus contingency or bus bar section contingency in DC sensitivity analysis, in the case of the lost bus has a VSC converter station.


src/main/java/com/powsybl/openloadflow/network/impl/LfVscConverterStationImpl.java
vidaldid-rte, line 1: The comment line 31 ("set only when AC emulation is activated") should be updated. The hvdc object is now set if both LfBus are in the LfNetwork.
@annetill

geofjamg (approved):

vidaldid-rte (approved):

annetill:
image

PR #970, opened 2024-01-30
Bump to powsybl-core 6.2.0

jeandemanged:

No

Feature

we use powsybl-core:6.2.0-RC1

we use powsybl-core:6.2.0

- [x] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,17 +90,17 @@ dependencies to respectively have access to network model, IEEE test networks an
 <dependency>
     <groupId>com.powsybl</groupId>
     <artifactId>powsybl-iidm-impl</artifactId>
-    <version>6.1.1</version>
+    <version>6.2.0</version>
 </dependency>
 <dependency>
     <groupId>com.powsybl</groupId>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,17 +90,17 @@ dependencies to respectively have access to network model, IEEE test networks an
...
     <groupId>com.powsybl</groupId>
     <artifactId>powsybl-ieee-cdf-converter</artifactId>
-    <version>6.1.1</version>
+    <version>6.2.0</version>
 </dependency>
 <dependency>
     <groupId>org.slf4j</groupId>
     <artifactId>slf4j-simple</artifactId>
-    <version>2.0.9</version>
+    <version>2.0.11</version>
```
olperr1, line 103: In powsybl-core 6.2.0, we still use slf4j-simple 2.0.9.

Thus, if someone defines a project with both powsybl-core 6.2.0 (without exclusions) and slf4j-simple 2.0.11, they will have both slf4j-simple versions in their dependency tree, which may lead to problems.

olperr1 (approved):

PR #976, opened 2024-02-07
Fix Static Var Compensator slope not simulated

jeandemanged:

No

Bugfix

Under certain condition an SVC slope might be ignored while it shouldn't.

With following options:
- voltagePerReactivePowerControl true
- svcVoltageMonitoring **false**

And an SVC with:
- VoltagePerReactivePowerControl extension defined (slope)
- StandbyAutomaton extension **not** defined

In this case an incorrect warning is raised and the slope not simulated.

`Static var compensator <id> has VoltagePerReactivePowerControl and StandbyAutomaton extensions, VoltagePerReactivePowerControl extension ignored`

SVC slope simulated as expected.

The unit test was relying too much on default `true` value of the `svcVoltageMonitoring` option, this has been fixed.

- [x] No

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowSvcTest.java
+++ b/src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowSvcTest.java
@@ -25,13 +25,13 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * SVC test case.
- *
+ *<pre>
  * g1        ld1
  * |          |
  * b1---------b2
  *      l1    |
  *           svc1
- *
+ *</pre>
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
 class AcLoadFlowSvcTest {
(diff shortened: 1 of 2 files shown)


annetill (approved):

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowSvcTest.java
```diff
@@ -272,12 +276,38 @@ void testStandByAutomaton() {
...
         LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        assertTrue(result.isFullyConverged());
 
         assertReactivePowerEquals(150.091, svc1.getTerminal());
         assertVoltageEquals(387.415, bus2);
     }
 
+    @Test
+    void testStandByAutomatonAndSlope() {
+        svc1.setVoltageSetpoint(385)
```
rosiereflo, line 289: Is there a test in which the StandByAutomaton is not defined while the VoltagePerReactivePowerControl is?

jeandemanged: Yes, see the testSvcWithSlope till testSvcWithSlope5 tests (were already existing)
Same, the tests testStandByAutomaton till testStandByAutomaton5 were already existing.

I added this test for the case where both extensions are defined - did not exist before. Today's behavior is that StandByAutomaton takes priority.

rosiereflo: ok thanks!

PR #981, opened 2024-02-21
Draft : SA : output a result for contingencies with no impact

vmouradian:

feature

When a contingency is converted to lfContingency befor beeing simulated three options are possible : 
* The contingency leads to an isolated slack bus -> contingency skipped and no result is provided
* The contingency has no impact -> contingency skipped and no result is provided
* The contingency is ok -> converted to lfContingency and simulated

* The contingency leads to an isolated slack bus -> contingency skipped and no result is provided
*  The contingency has no impact -> **return empty result with ``PostContingencyComputationStatus.NO_IMPACT``**
* The contingency is ok -> converted to lfContingency and simulated

This MR is a draft for the moment, the three cases are separated by a new enum to look more clear. But maybe it is a bit overkill.

New unit tests should still be added.

Another option could be to still create the lfContingency if it has no impact and add a hasNoImpact() method to the class (which won't be the same as hasNoImpact() from propagatedContingency, it is a deeper check) 

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
@@ -441,7 +441,17 @@ public class PropagatedContingency {
         return status == null || status == DisabledBranchStatus.SIDE_1;
     }
 
+    public enum ToLfContingencyConversionStatus {
+        OK,
+        ISOLATED_SLACK_BUS,
+        NO_IMPACT;
+    }
+
     public Optional<LfContingency> toLfContingency(LfNetwork network) {
+        return toLfContingencyWithStatus(network).getKey();
+    }
+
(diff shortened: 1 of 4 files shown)


obrix:
@vmouradian I am not sure we can merge as it is right now :

* I think we can probably avoid creating another enum for status
* This means maybe we lack ISOLATED_SLACK_BUS in post contingency computation status
* Which lead me to think we should maybe do things differently here : PostContingencyComputationStatus as its name indicates is a status after computation. Converged, max_iteration_reached and solver failed refers to the load flow evaluation while no_impact and isolated_slack_bus refers to the application of the contingency.

I think we should probably discussed with @annetill to see what is the most generic approach here.

vmouradian:
@obrix 
I agree with you on this, at the time I made the PR the goal was just to show that we have 3 separated cases where two were leading to no result. 
I created the enum to make it more visual but I also think that it is overkill for a final solution

PR #986, opened 2024-03-07
Generation action generator change must stay within Pmin and Pmax

vmouradian:

bugfix

During SA, when applying a lfAction that changes generator's Target P, it can set a target that is outside of the generator's Pmin and Pmax.

The new targetP value is limited by Pmin and Pmax

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfAction.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfAction.java
@@ -328,6 +328,26 @@ public final class LfAction {
         }
     }
 
+        if (generatorChange != null) {
+            LfGenerator generator = generatorChange.generator();
+            if (!generator.isDisabled()) {
+                double deltaTargetP = generatorChange.deltaTargetP();
+                double oldTargetP = generator.getTargetP();
+                double newTargetP = oldTargetP + deltaTargetP;
+                // if useActiveLimits is false or if the generator target is already outside its [Pmin;Pmax], the limits are ignored
+                if (networkParameters.isUseActiveLimits() && oldTargetP >= generator.getMinP() && oldTargetP <= generator.getMaxP()) {
+                    newTargetP = Math.max(Math.min(newTargetP, generator.getMaxP()), generator.getMinP());
+                    if (newTargetP == generator.getMaxP() || newTargetP == generator.getMinP()) {
+                        LOGGER.warn("Action on generator {} restricted to its active power limits [{}MW;{}MW], new target P set to {}MW", generator.getId(), generator.getMinP() * PerUnit.SB, generator.getMaxP() * PerUnit.SB, newTargetP * PerUnit.SB);
+                    }
+                }
+                generator.setTargetP(newTargetP);
+                if (!AbstractLfGenerator.checkActivePowerControl(generator.getId(), generator.getTargetP(), generator.getMinP(), generator.getMaxP(),
+                        networkParameters.getPlausibleActivePowerLimit(), networkParameters.isUseActiveLimits(), null)) {
+                    generator.setParticipating(false);
+                }
+            }
+        }
     private void applyGeneratorChange(LfNetworkParameters networkParameters) {
         LfGenerator generator = generatorChange.generator();
         if (!generator.isDisabled()) {
--- a/src/main/java/com/powsybl/openloadflow/network/LfAction.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfAction.java
@@ -328,36 +328,25 @@ public final class LfAction {
         }
     }
 
-        if (generatorChange != null) {
-            LfGenerator generator = generatorChange.generator();
-            if (!generator.isDisabled()) {
-                double changeTargetP = generatorChange.changeTargetP();
-                double oldTargetP = generator.getTargetP();
-                double newTargetP;
-                if (generatorChange.isRelative()) {
-                    newTargetP = oldTargetP + changeTargetP;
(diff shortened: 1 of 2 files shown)


Hadrien-Godard (approved):

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -335,7 +335,9 @@ public void apply(LfNetworkParameters networkParameters) {
         if (generatorChange != null) {
             LfGenerator generator = generatorChange.generator();
             if (!generator.isDisabled()) {
-                generator.setTargetP(generator.getTargetP() + generatorChange.deltaTargetP());
+                double deltaTargetP = generatorChange.deltaTargetP();
```
annetill, line 338: First, thank you for the bug fix. Second, your fix proposal is not very adapted for long-term support and users. I think that you have to do if loops in order to add a log message to explain why we don't want to modify a target P outside of the generator active limits. Because the fact is that you are not going to apply the remedial action asked by the user... we can change the remedial action and log, or throw...

Hadrien-Godard: I agree to add a log to explain that the whole action has not been applied (and why). Throwing an error seems problematic to me. In case of relative redispatch action, with big slack distribution in pre-contingency case, this case could happen a lot.

jeandemanged: need indeed a log

should also probably make use of option `useActiveLimit` of LfNetworkParameters, as follows:
- `useActiveLimit` set to `false`: ignore limits
- `useActiveLimit` set to `true`:
  - generator is within [Pmin;Pmax] before the action: restrict the change and log warning
  - generator is already outside [Pmin;Pmax]: ignore limits

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
```diff
@@ -892,34 +892,34 @@ private void testGeneratorAction(boolean dc, LoadFlowParameters.BalanceType bala
...
         assertEquals(network.getLine("l14").getTerminal1().getP(), getOperatorStrategyResult(result, "strategyG3").getNetworkResult().getBranchResult("l14").getP1(), LoadFlowAssert.DELTA_POWER);
         assertEquals(network.getLine("l23").getTerminal1().getP(), getOperatorStrategyResult(result, "strategyG3").getNetworkResult().getBranchResult("l23").getP1(), LoadFlowAssert.DELTA_POWER);
         assertEquals(network.getLine("l34").getTerminal1().getP(), getOperatorStrategyResult(result, "strategyG3").getNetworkResult().getBranchResult("l34").getP1(), LoadFlowAssert.DELTA_POWER);
 
         // reverse action and apply fourth remedial action
-        network.getGenerator(g2).setTargetP(g2PostContingencyTargetP + deltaG2);
-        network.getGenerator(g1).setTargetP(g1PostContingencyTargetP + deltaG1);
+        setTargetPWithinLimits(network.getGenerator(g2), g2PostContingencyTargetP + deltaG2, openLoadFlowParameters);
+        setTargetPWithinLimits(network.getGenerator(g1), g1PostContingencyTargetP + deltaG1, openLoadFlowParameters);
         loadFlowRunner.run(network, parameters);
```
rosiereflo, line 923: Is there a test for which the P is outside the limits when useActiveLimit is enabled, and one with useActiveLimit disabled? Thanks

vmouradian: Not for the moment (which caused unsufficient code coverage by the way), I am currently adding one

annetill (changes requested):
The requirements are not clear enough to provide a good fix. The remedial action provides two ways to change the generator targetP:
- In relative value: the change (given by the user) happens just after contingency and slack distribution.
- In absolue value: in that case, the implementation in main branch is questionnable because it follow the same logic as the relative remedial action and it could be confusing by the user that the generator in question after remedial action is not at the set point defined in the remedial action.

So first, we have to clarify these two points. Secondly: for open RAO, we need to be able to set a targetP to zero and if I understand well what you are doing, we cannot anymore. Maybe a good solution could be (and has to be discuss) :
- In relative value, we keep what you have done with the minP/maxP check.
- In absolute value, maybe we have to apply what is given by the user, whatever he gives. So maybe no minP/maxP check...  

Can we clarify what we want first, add a unit test to force a targetP to zero and then provides the good fix?

vmouradian:
Yes of course, I have to admit that I was not aware of all that was at stake here.
From my comprehension, what you suggest would also need a small refactor of LfAction (GeneratorChange more specifically) and give it the boolean value of activePowerRelativeValue from the GeneratorAction class, is that correct ?

annetill:
Can I close this PR?

annetill:
@vmouradian can you rebase your PR?

annetill:
After a talk with @Hadrien-Godard , it seams impossible to make something coherent everywhere with `useActiveLimits` parameter and it is not the initial need that is only related to a generator action with relative value change. We think that:
- First we have to check if information inside `GeneratorAction` is enough ;
- Then, if not, we can imagine a dedicated action for counter trading and re-dispatching in order to correctly deal with slack distribution, respect of active power limits and respect of the volumes asked by the user.

PR #990, opened 2024-03-08
Always update network with a reference terminal

annetill:

No, but this PR (https://github.com/powsybl/powsybl-core/pull/2921) is going to rely on a reference terminal extension and no more on the slack terminal extension.

- We always update the network with a reference terminal. Note that it will be the same for slack terminal (https://github.com/powsybl/powsybl-core/pull/2921).
- In case of  `FIRST_SLACK` mode, we update with the first connected terminal of the bus. 

- [ ] Yes
- [x] No

**Other information**:


jeandemanged:
excellent, just a very minor sonar issue to solve

PR #994, opened 2024-03-12
Network slack bus selector multi-variant support/bug

annetill:

Working with variants, we have a null slackTerminal.getTerminal(). 

Bug fix

It is valid that a VoltageLevel has a `SlackTerminal` extension present and at the same time the `extension.getTerminal()` returns `null`: this happens if the VoltageLevel does not hold any slack for the current variant, but for another variant.

Currently this causes an exception in OLF.

Proper handling of the case of a SlackTerminal extension present and at the same time extension.getTerminal() returns null for the current network variant.

- [x] No

n/a

**Other information**:

n/a


src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowEurostagTutorialExample1Test.java
```diff
@@ -465,4 +465,26 @@ void maxOuterLoopIterationTest() {
         assertFalse(result.isFullyConverged());
         assertEquals(LoadFlowResult.ComponentResult.Status.MAX_ITERATION_REACHED, result.getComponentResults().get(0).getStatus());
     }
+
+    @Test
+    void testWriteReadSlackBus() {
+        Network network = EurostagFactory.fix(EurostagTutorialExample1Factory.create());
+        network.getVariantManager().cloneVariant(network.getVariantManager().getWorkingVariantId(), "newVariant");
+        LoadFlowParameters parameters = new LoadFlowParameters();
+        parameters.setWriteSlackBus(true);
```
jeandemanged, line 474: you could use method chaining, and also setReadSlackBus(true) to better emphasize the test intention

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowEurostagTutorialExample1Test.java
```diff
@@ -465,4 +465,26 @@ void maxOuterLoopIterationTest() {
...
+        OpenLoadFlowParameters openLoadFlowParameters =
+                OpenLoadFlowParameters.create(parameters).setSlackBusSelectionMode(SlackBusSelectionMode.FIRST);
+        LoadFlowResult result = LoadFlow.run(network, parameters);
+        assertTrue(result.isFullyConverged());
+        SlackTerminal slackTerminal = network.getVoltageLevel("VLGEN").getExtension(SlackTerminal.class);
+        assertNotNull(slackTerminal);
+        assertNotNull(slackTerminal.getTerminal());
```
jeandemanged, line 481: you could also demonstrate, here, or at the end of the test, via assertions, that slackTerminal.getTerminal() is null in the `newVariant` variant, this would make the test purpose much more explicit.

jeandemanged (changes requested):
i suggest make the test purpose more explicit

jeandemanged (approved):
ty

PR #999, opened 2024-03-15
NR mismatch reporter: add unit in mismatches

caioluke:

Bug fix

Currently reported NR mismatches are unitless

Add units for reported NR mismatches

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/solver/NewtonRaphson.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/solver/NewtonRaphson.java
@@ -85,8 +85,8 @@ public class NewtonRaphson extends AbstractAcSolver {
                         double busPhi = equationSystem.getStateVector().get(busPhiRow);
                         LfBus bus = network.getBus(elementNum);
                         double busNominalV = bus.getNominalV();
-                        double busInjectionP = bus.getTargetP() * PerUnit.SB;
-                        double busInjectionQ = bus.getTargetQ() * PerUnit.SB;
+                        double busSumP = bus.getP().eval() * PerUnit.SB;
+                        double busSumQ = bus.getQ().eval() * PerUnit.SB;
 
                         if (LOGGER.isTraceEnabled()) {
                             LOGGER.trace("Largest mismatch on {}: {}", getEquationTypeDescription(acEquationType), equationMismatch);
@@ -95,7 +95,7 @@ public class NewtonRaphson extends AbstractAcSolver {
                         }
 
                         if (reportNode != null) {
-                            Reports.BusReport busReport = new Reports.BusReport(elementId, equationMismatch, busNominalV, busV, busPhi, busInjectionP, busInjectionQ);
+                            Reports.BusReport busReport = new Reports.BusReport(elementId, equationMismatch, busNominalV, busV, busPhi, busSumP, busSumQ);
                             Reports.reportNewtonRaphsonLargestMismatches(reportNode, getEquationTypeDescription(acEquationType), busReport);
                         }
                     });
--- a/src/main/java/com/powsybl/openloadflow/ac/solver/NewtonRaphson.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/solver/NewtonRaphson.java
@@ -85,8 +85,8 @@ public class NewtonRaphson extends AbstractAcSolver {
                         double busPhi = equationSystem.getStateVector().get(busPhiRow);
                         LfBus bus = network.getBus(elementNum);
                         double busNominalV = bus.getNominalV();
-                        double busSumP = bus.getP().eval() * PerUnit.SB;
-                        double busSumQ = bus.getQ().eval() * PerUnit.SB;
+                        double busInjectionP = bus.getTargetP() * PerUnit.SB;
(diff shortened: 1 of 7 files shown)


src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -392,12 +388,33 @@ public static void reportNewtonRaphsonLargestMismatches(Reporter reporter, Strin
...
+                mismatchUnitConverter = PerUnit.SB;
+            }
+            case "Q" -> {
+                mismatchUnit = "MVar";
+                mismatchUnitConverter = PerUnit.SB;
+            }
+            case "V" -> {
+                mismatchUnit = "kV";
+                mismatchUnitConverter = busReport.nominalV();
+            }
```
jeandemanged, line 410: leave in p.u. for voltage equations.

jeandemanged (changes requested):
also a sonar issue to solve

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -392,12 +388,33 @@ public static void reportNewtonRaphsonLargestMismatches(Reporter reporter, Strin
...
+                mismatchUnit = "p.u.";
+                mismatchUnitConverter = 1.0;
+            }
+        }
+
+        Map<String, TypedValue> subReporterMap = new HashMap<>();
+        subReporterMap.put("equationType", new TypedValue(acEquationType, TypedValue.UNTYPED));
+        subReporterMap.put("mismatch", new TypedValue(mismatchUnitConverter * busReport.mismatch(), OpenLoadFlowReportConstants.MISMATCH_TYPED_VALUE));
+        subReporterMap.put("mismatchUnit", new TypedValue(mismatchUnit, TypedValue.UNTYPED));
+        Reporter subReporter = reporter.createSubReporter("NRMismatch" + acEquationType, "Largest ${equationType} mismatch: ${mismatch} ${mismatchUnit}", subReporterMap);
```
jeandemanged, line 417: I cannot understand the reason of the change of key from just `"NRMismatch"` to now `"NRMismatch" + acEquationType` ? Why do you need a different key, while the formatting message is always the same ?

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/util/Reports.java
caioluke: You're right. I guess I thought that since the units would be different, so should the reporter name. Just fixed it.

jeandemanged (approved):

src/main/java/com/powsybl/openloadflow/ac/solver/NewtonRaphson.java
```diff
@@ -85,8 +85,8 @@ public void reportAndLogLargestMismatchByAcEquationType(ReportNode reportNode, E
                         double busPhi = equationSystem.getStateVector().get(busPhiRow);
                         LfBus bus = network.getBus(elementNum);
                         double busNominalV = bus.getNominalV();
-                        double busSumP = bus.getP().eval() * PerUnit.SB;
-                        double busSumQ = bus.getQ().eval() * PerUnit.SB;
+                        double busInjectionP = bus.getTargetP() * PerUnit.SB;
+                        double busInjectionQ = bus.getTargetQ() * PerUnit.SB;
```
vidaldid-rte, line 90: @caioluke not sure why you call this quantity a busInjectionQ. For a PV node the targetQ is irrelevant.
You may also miss some reactive terms that are usually considered as bus injection like shunt and static var compensator.

caioluke: @vidaldid-rte what about reporting the targetQ for PQ buses and bus.getQ().eval() for PV buses?
How would I get all these other reactive terms?

vidaldid-rte: IM' not sure what is expected  in the report and if it is really possible do derive a bus injection before the NR converges.

For something named bus injection, I would understand if as all the injections that do not come from the branches. 
I would define it as follow for a Q or P equation
rhs - sum of all active equation terms in the equation for type ElementType.BRANCH
although it has no physical meaning before the NR converges because it includes the equation mismatch
The above expression would include
    The equation mismatch:
    For P (if inactive) , the slack at slack bus
    For Q, if PV node and Q equatin inactive,  the reactive power to distribute to the generators controlling tension
    The shunt terms
    The HVDC station terms (AC emulation mode)
    The static injection (targetQ ...)

rosiereflo (approved):

src/main/java/com/powsybl/openloadflow/ac/solver/NewtonRaphson.java
vidaldid-rte: OK. I agree that the report is misleading because it labels the quantity "bus injection".
It is hard to give a user friendly name for that quantity. In the context of Newton Raphson it is the Left Hand Side of the equation (the computed part) that is supposed when the equation is active to match the static part, the right hand side.

vidaldid-rte (approved):

jeandemanged (approved):

PR #1003, opened 2024-03-25
Fix restoreInitialTopology in SA with multiple SCs

caioluke:

Bug fix

When restoring the initial topology in SA, we try to close switches/branches in all LfNetworks. When we have more than 1 LfNetwork, the code crashes.

We will be able to restore the initial topology in SA with multiple SCs.

- [ ] Yes
- [X] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
@@ -136,18 +136,18 @@ public final class Networks {
         }); // in order to be present in the network.
     }
 
-    private static Set<String> restoreInitialTopology(LfNetwork network, Set<String> switchAndBranchIdsToClose) {
-        Set<String> switchAndBranchIdsLeftToClose = new HashSet<>();
+    private static void restoreInitialTopology(LfNetwork network, Set<String> switchAndBranchIdsToClose) {
         var connectivity = network.getConnectivity();
         connectivity.startTemporaryChanges();
+        Set<String> toRemove = new HashSet<>();
         switchAndBranchIdsToClose.forEach(id -> {
             LfBranch branch = network.getBranchById(id);
             if (branch != null) {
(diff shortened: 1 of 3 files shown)


obrix:
The missing coverage are edge cases very hard to test in a unit test. I feel we can merge as is.

obrix (approved):

src/main/java/com/powsybl/openloadflow/network/impl/Networks.java
```diff
@@ -222,9 +224,21 @@ public static LfNetworkList load(Network network, LfNetworkParameters networkPar
...
                 for (LfNetwork lfNetwork : lfNetworks) {
+                    // all switches and branches were closed
+                    if (switchAndBranchIdsLeftToClose.isEmpty()) {
+                        break;
+                    }
                     // disable all buses and branches not connected to main component (because of switch to close)
-                    restoreInitialTopology(lfNetwork, modifiedTopoConfig.getSwitchesToClose(), modifiedTopoConfig.getBranchIdsToClose());
+                    restoreInitialTopology(lfNetwork, switchAndBranchIdsLeftToClose);
+                }
+                if (!switchAndBranchIdsLeftToClose.isEmpty()) {
```
annetill, line 240: I don't understand why we could be in that case (with the exception). If we are in that case, it means that we have a bug in the code. Any advice @geofjamg ?

geofjamg: Either:
  - we are able to add a test case for this case because it is a real case
  - we remove the code because it cannot happen

caioluke: It is not a real case, I just put it to be sure that we would catch a bug if it happened. I will remove this bit of code.

annetill (approved):

geofjamg (approved):

PR #1005, opened 2024-03-26
Honnor HVDCOperatorActivePowerRange as well as Pmax of hvdcLine in AC emulation

vidaldid-rte:

Take into account active power limits of HVDC in AC emulations
They can be defined by
  maxP in the HVDC line
Or in the  HvdcOperatorActivePowerRange extension that provides a limit in each direction

Limits are ignored in AC Emulation

Limits are honoured.
If P0+KdeltaPhi out of range, P is capped to a bound of the range?

- [ ] Yes
- [ X] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfHvdcImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfHvdcImpl.java
@@ -46,12 +46,7 @@ public class LfHvdcImpl extends AbstractElement implements LfHvdc {
 
     private final double pMaxFromCS2toCS1;
 
-    public LfHvdcImpl(String id,
-                      LfBus bus1,
-                      LfBus bus2,
-                      LfNetwork network,
-                      HvdcLine hvdcLine,
-                      boolean acEmulation) {
+    public LfHvdcImpl(String id, LfBus bus1, LfBus bus2, LfNetwork network, HvdcLine hvdcLine, boolean acEmulation) {
         super(network);
         this.id = Objects.requireNonNull(id);
         this.bus1 = bus1;
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfHvdcImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfHvdcImpl.java
@@ -6,7 +6,9 @@
  */
 package com.powsybl.openloadflow.network.impl;
 
+import com.powsybl.iidm.network.HvdcLine;
 import com.powsybl.iidm.network.extensions.HvdcAngleDroopActivePowerControl;
+import com.powsybl.iidm.network.extensions.HvdcOperatorActivePowerRange;
 import com.powsybl.openloadflow.network.*;
 import com.powsybl.openloadflow.util.Evaluable;
 import com.powsybl.openloadflow.util.PerUnit;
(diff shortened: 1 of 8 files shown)


src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowVscTest.java
```diff
@@ -468,4 +469,64 @@ void testVscVoltageControlWithOneSideDisconnected() {
...
+
+        LoadFlow.Runner loadFlowRunner = new LoadFlow.Runner(new OpenLoadFlowProvider(new DenseMatrixFactory()));
+        LoadFlowParameters p = new LoadFlowParameters();
+        p.setHvdcAcEmulation(true);
+        LoadFlowResult result = loadFlowRunner.run(network, p);
+
+        assertTrue(result.isFullyConverged());
+
+        // Active flow capped at limit. Output has losses (due to VSC stations)
+        assertEquals(170, network.getHvdcConverterStation("cs2").getTerminal().getP(), DELTA_POWER);
```
annetill, line 491: Can you used dedicated method for voltage or power assertions `assertActivePowerEquals` and `assertVoltageEquals`?

src/main/java/com/powsybl/openloadflow/network/impl/LfHvdcImpl.java
```diff
@@ -40,16 +42,33 @@ public class LfHvdcImpl extends AbstractElement implements LfHvdc {
 
     private boolean acEmulation;
 
-    public LfHvdcImpl(String id, LfBus bus1, LfBus bus2, LfNetwork network, HvdcAngleDroopActivePowerControl control,
+    private final double pMaxFromCS1toCS2;
+
+    private final double pMaxFromCS2toCS1;
+
+    public LfHvdcImpl(String id,
```
annetill, line 49: Just a detail, can you keep the format used in the rest of the code? We put in line. It is just to have an homogeneous way to code.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowVscTest.java
vidaldid-rte: Done

src/main/java/com/powsybl/openloadflow/network/impl/LfHvdcImpl.java
vidaldid-rte: Done. But I do not like this layout for long list of arguments (larger than a laptop screen).
It's OK here.

annetill (approved):

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowVscTest.java
```diff
@@ -468,4 +469,64 @@ void testVscVoltageControlWithOneSideDisconnected() {
...
+        network.getHvdcLine("hvdc23")
+                .newExtension(HvdcOperatorActivePowerRangeAdder.class)
+                .withOprFromCS2toCS1(180)
+                .withOprFromCS1toCS2(170)
+                .add();
+
+        LoadFlow.Runner loadFlowRunner = new LoadFlow.Runner(new OpenLoadFlowProvider(new DenseMatrixFactory()));
+        LoadFlowParameters p = new LoadFlowParameters();
+        p.setHvdcAcEmulation(true);
+        LoadFlowResult result = loadFlowRunner.run(network, p);
```
rosiereflo, line 486: Is there a test with HvdcAngleDroopActivePowerControl but no HvdcOperatorActivePowerRange extension?

vidaldid-rte: @rosiereflo  Yes. All other tests with HVDC in AC emulation. For example testHvdcDirectionChangeAcEmulation

annetill: Do we have to take into account the active power range extension if the hvdc is operated in set point ?

vidaldid-rte: Good question @annetill 
   My approach was to consider that an extension has priority.
   This is consistent with the fact that the AC emulation takes priority on both the mode (side of inverter/rectifier) and active power setPoint.
   But there is nature for conflict because data is duplicated and we need to agree whether giving priority on the extension is what we want or not.
   @rosiereflo

vidaldid-rte: Also the fix did not cover the case where AC Emlulation is off.

vidaldid-rte: But I have not seen consistency check between p and pmax in HvdcLineAdder

annetill: In OLF, I have noticed that we can have a generator with a `targetP` over `Pmax`, it is not forbidden but maybe it should... for VSC converter station, no check either. Any idea @geofjamg ?

jeandemanged: For generators it is very common to have a targetP outside Pmin/Pmax on real data.
Usual expected behavior for these is just to leave them where they are outside Pmin/Pmax and don't change them at all, e.g. exclude them from slack distribution.
I would be against enforcing targetP within Pmin/Pmax for generators in iIDM. Same for reactive power.
For HVDC I don't have any strong opinion though... There aren't that much, and AC emulation is ultra specific to few links.

annetill: For we can say that it is the user's responsibility and not the purpose of this fix, related to hvdc in AC emulation.

annetill (approved):
Still approved

PR #1009, opened 2024-04-02
Improve DC initialisation 

annetill:

Yes, after PR https://github.com/powsybl/powsybl-open-loadflow/pull/1005, we have found several load flow divergences in integration tests. The initialisation with DC values gives a state that is out of bounds at the first AC NR. This leads to an impossible jump and a divergence. If we switch to an initialization with uniform values, we have a convergence. But this is not a good solution because initialisation with DC values is required for operational processes.

Note that in main branch, if we run an AC load flow with an HVDC in AC emulation, the DC load flow used for initialisation takes for the two converter stations zero MW as targets because of 

image

and not the targetP calculated from the active power set point. It is not better not worst. 

We introduce HVDC "AC" emulation equations for DC load flows.

- [ ] Yes
- [ ] No

**Other information**:

Rework of https://github.com/powsybl/powsybl-open-loadflow/pull/691/, with half features in this PR.


src/main/java/com/powsybl/openloadflow/dc/equations/AbstractHvdcDcEmulationFlowEquationTerm.java
```diff
@@ -0,0 +1,89 @@
...
+import com.powsybl.openloadflow.equations.VariableSet;
+import com.powsybl.openloadflow.network.LfBus;
+import com.powsybl.openloadflow.network.LfHvdc;
+
+import java.util.List;
+
+/**
+ * @author Anne Tilloy {@literal <anne.tilloy at rte-france.com>}
+ */
+public abstract class AbstractHvdcDcEmulationFlowEquationTerm extends AbstractElementEquationTerm<LfHvdc, DcVariableType, DcEquationType> {
```
jeandemanged, line 20: `HvdcDcEmulation` is a bit startling. I'd call this `HvdcAcEmulation` also in DC load flow.

src/main/java/com/powsybl/openloadflow/dc/equations/HvdcDcEmulationSide1ActiveFlowEquationTerm.java
```diff
@@ -0,0 +1,73 @@
...
+import com.powsybl.openloadflow.equations.VariableSet;
+import com.powsybl.openloadflow.network.LfBus;
+import com.powsybl.openloadflow.network.LfHvdc;
+
+import java.util.Objects;
+
+/**
+ * @author Anne Tilloy {@literal <anne.tilloy at rte-france.com>}
+ */
+public class HvdcDcEmulationSide1ActiveFlowEquationTerm extends AbstractHvdcDcEmulationFlowEquationTerm {
```
jeandemanged, line 19: Same `HvdcDcEmulation` vs `HvdcAcEmulation`

src/main/java/com/powsybl/openloadflow/dc/equations/HvdcDcEmulationSide2ActiveFlowEquationTerm.java
```diff
@@ -0,0 +1,73 @@
...
+import com.powsybl.openloadflow.equations.VariableSet;
+import com.powsybl.openloadflow.network.LfBus;
+import com.powsybl.openloadflow.network.LfHvdc;
+
+import java.util.Objects;
+
+/**
+ * @author Anne Tilloy {@literal <anne.tilloy at rte-france.com>}
+ */
+public class HvdcDcEmulationSide2ActiveFlowEquationTerm extends AbstractHvdcDcEmulationFlowEquationTerm {
```
jeandemanged, line 19: Same `HvdcDcEmulation` vs `HvdcAcEmulation`

src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
```diff
@@ -55,6 +55,7 @@ protected DcLoadFlowParameters createParameters(LoadFlowParameters lfParameters,
         var dcParameters = OpenLoadFlowParameters.createDcParameters(network, lfParameters,
                 lfParametersExt, matrixFactory, connectivityFactory, false);
         dcParameters.getNetworkParameters()
+                .setHvdcAcEmulation(false) // we only support it for DC load flow (initialization)
```
jeandemanged, line 58: I guess this should work in SA too ? why restricting here ?

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/sa/DcSecurityAnalysis.java
annetill: Done, still not supported for DC sensitivity analysis.

jeandemanged (approved):
OK for me, but PR title to be adjusted. 
I suggest `Support HVDC AC emulation simulation in DC Load Flow`.
(And a byproduct of that is that the DC Init of an AC Load Flow is correctly performed).

vidaldid-rte:
@annetill  @geofjamg 
this PR breaks <hat I have understood as basic assumptions of the DC mode:
    - loss are neglected
   -  equations should be linear
I do not enough enough DC computations tu understand the impact of breaking this rule.
However:


It seems you can restore the properties by making some private functions protected and providing this implementation at DC level
    protected double boundedP(double rawP) {
        // to be linear
        return rawP;
    }

    protected boolean isInOperatingRange(double rawP) {
        return true;
    }

    protected double getVscLossMultiplier() {
        return 1;
    }

In my test, this resolve the problem of providing a good starting point for AC computation so that the equation converges.
Does it work for you other cases ?

annetill:
It makes sense. We can remove this behavior in the equations for DC computations and introducing a small outer loop for DC calculations that handles HVDC limits.

rosiereflo (approved):

vidaldid-rte (changes requested):
The equations are not linear. The derivative are not constant. This it not a DC model.

vidaldid-rte:
As discussed with @annetill  here is an alternative proposal inspired from this PR. Differences are
    - Linear equations
    - Support of non null P0
Still not support for PMax (but is it a good idea to introduce this non linearity that appears at large angles in a DC model ?)

https://github.com/powsybl/powsybl-open-loadflow/pull/1013

annetill:
> The equations are not linear. The derivative are not constant. This it not a DC model.

I invite you to add your simplifications in the PR to have a linear model. Your proposal on branch test2 leads to all passing tests of open RAO. But with simple equations in DC equation system as you mention me earlier today.

jeandemanged:
I would not say that "lossless" and "linear" are _rules_ of DC Load Flow. These are only _simplifications_ made, not rules. These simplifications are made to AC part of the network, so it can be solved in a single iteration. But I don't think anything exists in literature about HVDC AC emulation, on how this should be done in a DC Load Flow.

If it is possible to deal with converter losses and converter limits, while keeping the very nice property of DC Load Flow of being ultra fast and always providing a solution (of course approximate), I won't be against it.

Thinking about it, if converter losses are modelled, it is not possible anymore to distribute slack easily in one shot as a DC Load Flow preprocessing, so indeed this would be the good and valid reason to ignore them.

annetill:
> I would not say that "lossless" and "linear" are _rules_ of DC Load Flow. These are only _simplifications_ made, not rules. These simplifications are made to AC part of the network, so it can be solved in a single iteration. But I don't think anything exists in literature about HVDC AC emulation, on how this should be done in a DC Load Flow.
> 
> If it is possible to deal with converter losses and converter limits, while keeping the very nice property of DC Load Flow of being ultra fast and always providing a solution (of course approximate), I won't be against it.
> 
> Thinking about it, if converter losses are modelled, it is not possible anymore to distribute slack easily in one shot as a DC Load Flow preprocessing, so indeed this would be the good and valid reason to ignore them.

For converter and hvdc losses, you cannot do it in a single iteration because you must know the sign of ph1 - ph2. So I think what @vidaldid-rte proposes is a good approximation.

annetill:
Rework with #1013, thanks everybody for the support!

PR #1011, opened 2024-04-03
Minimal support of limit reductions for security analysis

annetill:

No, it is the support of a new feature from `powsybl-core 6.3.0`. In that PR, we only support a minimal feature, without relying on utility methods available in core. But we can imagine testing this approach and with utility methods in order to compare results and performances (on core side, could be a problem).

- [ ] Yes
- [ ] No

**Other information**:


olperr1 (changes requested):

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBranch.java
```diff
@@ -74,21 +78,26 @@ public List<BranchResult> createBranchResult(double preContingencyBranchP1, doub
...
+                return getLimits1(type, danglingLine.getCurrentLimits().orElse(null), limitReductionManager);
             case VOLTAGE:
             default:
                 throw new UnsupportedOperationException(String.format("Getting %s limits is not supported.", type.name()));
         }
     }
 
+    @Override
+    public List<Double> getLimitReductions(TwoSides side, LimitReductionManager limitReductionManager, LoadingLimits limits) {
+        return Collections.emptyList();
```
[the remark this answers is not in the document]
annetill: Yes :-) My PR was just as minimal as possible because the plan is to use powsybl-core utility methods and not my work. Indeed this PR is just a MVP to validate the interface (which we did together and the utility methods) and to give something quickly to users but not definitive at all.

src/main/java/com/powsybl/openloadflow/sa/LimitReductionManager.java
```diff
@@ -0,0 +1,173 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
```
olperr1, line 6:
```suggestion
/**
 * Copyright (c) 2024, RTE (http://www.rte-france.com)
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 * SPDX-License-Identifier: MPL-2.0
 */
```

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -77,21 +75,26 @@ protected AbstractLfBranch(LfNetwork network, LfBus bus1, LfBus bus2, PiModel pi
...
 
-    protected static List<LfLimit> createSortedLimitsList(LoadingLimits loadingLimits, LfBus bus) {
+    protected static List<LfLimit> createSortedLimitsList(LoadingLimits loadingLimits, LfBus bus, List<Double> limitReductions) {
         LinkedList<LfLimit> sortedLimits = new LinkedList<>();
         if (loadingLimits != null) {
             double toPerUnit = getScaleForLimitType(loadingLimits.getLimitType(), bus);
 
+            int i = 0;
             for (LoadingLimits.TemporaryLimit temporaryLimit : loadingLimits.getTemporaryLimits()) {
+                i++;
```
rosiereflo, line 85: Would be cleaner to increment an index in the for loop

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -77,21 +75,26 @@ protected AbstractLfBranch(LfNetwork network, LfBus bus1, LfBus bus2, PiModel pi
...
                 if (temporaryLimit.getAcceptableDuration() != 0) {
                     // it is not useful to add a limit with acceptable duration equal to zero as the only value plausible
                     // for this limit is infinity.
                     // https://javadoc.io/doc/com.powsybl/powsybl-core/latest/com/powsybl/iidm/network/CurrentLimits.html
-                    double valuePerUnit = temporaryLimit.getValue() * toPerUnit;
-                    sortedLimits.addFirst(LfLimit.createTemporaryLimit(temporaryLimit.getName(), temporaryLimit.getAcceptableDuration(), valuePerUnit));
+                    Double reduction = limitReductions.isEmpty() ? null : limitReductions.get(i);
```
rosiereflo, line 90: so 0 is not a possible index as the first i value is 1?

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -229,37 +228,70 @@ public List<BranchResult> createBranchResult(double preContingencyBranchP1, doub
...
+        Arrays.fill(limitReductions, 1.);
+        double nominalV = branchRef.get().getTerminal(side).getVoltageLevel().getNominalV();
+        for (LimitReductionManager.TerminalLimitReduction terminalLimitReduction : limitReductionManager.getTerminalLimitReductions()) {
+            if (terminalLimitReduction.getNominalV().contains(nominalV)) {
+                if (terminalLimitReduction.isPermanent()) {
+                    limitReductions[0] = terminalLimitReduction.getReduction();
+                }
+                if (terminalLimitReduction.getAcceptableDuration() != null) {
+                    int i = 1;
+                    for (LoadingLimits.TemporaryLimit temporaryLimit : limits.getTemporaryLimits()) {
```
rosiereflo, line 283: same as above regarding the index

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -10,12 +10,11 @@
...
 import com.powsybl.iidm.network.extensions.LineFortescue;
 import com.powsybl.openloadflow.network.*;
+import com.powsybl.openloadflow.sa.LimitReductionManager;
 import com.powsybl.openloadflow.util.PerUnit;
 import com.powsybl.security.results.BranchResult;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
rosiereflo, line 17: I haven't seen what are the newly used objects that make this change relevant?

src/main/java/com/powsybl/openloadflow/sa/LimitReductionManager.java
```diff
@@ -0,0 +1,174 @@
...
+
+    List<TerminalLimitReduction> terminalLimitReductions = new ArrayList<>();
+
+    public boolean isEmpty() {
+        return terminalLimitReductions.isEmpty();
+    }
+
+    public List<TerminalLimitReduction> getTerminalLimitReductions() {
+        // to be sorted by acceptable duration
+        // return terminalLimitReductions.stream().sorted().toList();
```
rosiereflo, line 78: Please remove useless comment

rosiereflo (changes requested):

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -229,37 +228,70 @@ public List<BranchResult> createBranchResult(double preContingencyBranchP1, doub
...
+            if (terminalLimitReduction.getNominalV().contains(nominalV)) {
+                if (terminalLimitReduction.isPermanent()) {
+                    limitReductions[0] = terminalLimitReduction.getReduction();
+                }
+                if (terminalLimitReduction.getAcceptableDuration() != null) {
+                    int i = 1;
+                    for (LoadingLimits.TemporaryLimit temporaryLimit : limits.getTemporaryLimits()) {
+                        if (terminalLimitReduction.getAcceptableDuration().contains(temporaryLimit.getAcceptableDuration())) {
+                            limitReductions[i] = terminalLimitReduction.getReduction();
+                        }
```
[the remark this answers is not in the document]
olperr1: 1. The array is initialized with (_temporary limits number_ +1) values [^1] set to  `1.` (see l.273 and 274). Even if no reduction is defined for temporary limits, there will always be a reduction value for each temporary limit (`1.`);
2. That's it: when a limit matches the criteria of several `LimitReduction`s, the last one applies. I will add a comment and a unit test to clearly identify (and check) this point.

[^1]: index 0 is reserved for the permanent limit

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
olperr1: The index 0 is reserved for the permanent limit. Temporary limits are stored from 1. This explains the loop syntax that was used.

I will add comments about it.

src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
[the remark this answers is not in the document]
vidaldid-rte: Thanks @olperr1 . I get it (although it is a bit hard to read )
What about the situation where several limits match the duration and voltage ?
It seems in this implementation that the last limit overrides the other.
Is it really the desired behaviour ? Won't it be difficult to troubleshoot is in a complex software several limit reductions would apply for a line ?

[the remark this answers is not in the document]
vidaldid-rte: One more thing. This algorithm is quite expensive since you need to iterate on all limit reductions.
And the function might be called quite frequently (for example for at least once per contingency and per branch).

Why not at least caching the result for each LfBranch so that the iteration is  made at most once per branch ?

rosiereflo (approved):

vidaldid-rte (approved):
@annetill  @olperr1 
I'm OK with this PR with a concern (more in the SA API) that there is no clear rule to define which reduction applies in case several reductions can match a limit.

annetill:
> @annetill @olperr1 I'm OK with this PR with a concern (more in the SA API) that there is no clear rule to define which reduction applies in case several reductions can match a limit.

Yes, we are aware of that: the last one is taken.

PR #1017, opened 2024-04-05
Absolute generator action ignores post-contingency slack

vmouradian:

Bug fix (i guess ?)

Powsybl Generator actions are all converted to relative power actions, applied on top of the power after the contingency (and potential slack distribution)

If the action is absolute, the absolute targetP is directly set, regardless of the potential changes due to slack after contingency

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfAction.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfAction.java
@@ -309,50 +309,70 @@ public final class LfAction {
         }
     }
 
+    private void applyTapPositionChange() {
+        LfBranch branch = tapPositionChange.branch();
+        int tapPosition = branch.getPiModel().getTapPosition();
+        int value = tapPositionChange.value();
+        int newTapPosition = tapPositionChange.isRelative() ? tapPosition + value : value;
+        branch.getPiModel().setTapPosition(newTapPosition);
+    }
+
+    private void applyLoadShift() {
+        LfLoad load = loadShift.load();
+        if (!load.isOriginalLoadDisabled(loadShift.loadId())) {
+            PowerShift shift = loadShift.powerShift();
+            load.setTargetP(load.getTargetP() + shift.getActive());
+            load.setTargetQ(load.getTargetQ() + shift.getReactive());
+            load.setAbsVariableTargetP(load.getAbsVariableTargetP()
+                    + Math.signum(shift.getActive()) * Math.abs(shift.getVariableActive()));
+        }
+    }
+
+    private void applyGeneratorChange(LfNetworkParameters networkParameters) {
+        LfGenerator generator = generatorChange.generator();
+        if (!generator.isDisabled()) {
+            double newTargetP = generatorChange.isRelative() ? generator.getTargetP() + generatorChange.activePowerValue() : generatorChange.activePowerValue();
+            generator.setTargetP(newTargetP);
+            if (!AbstractLfGenerator.checkActivePowerControl(generator.getId(), generator.getTargetP(), generator.getMinP(), generator.getMaxP(),
+                    networkParameters.getPlausibleActivePowerLimit(), networkParameters.isUseActiveLimits(), null)) {
+                generator.setParticipating(false);
+            }
+        }
+    }
+
+    private void applyHvdcAction() {
+        hvdc.setAcEmulation(false);
+        hvdc.setDisabled(true); // for equations only, but should be hidden
+        hvdc.getConverterStation1().setTargetP(-hvdc.getP1().eval()); // override
+        hvdc.getConverterStation2().setTargetP(-hvdc.getP2().eval()); // override
+    }
+
+    private void applySectionChange() {
+        LfShunt shunt = sectionChange.shunt();
+        shunt.getControllers().stream().filter(controller -> controller.getId().equals(sectionChange.controllerId())).findAny()
+                .ifPresentOrElse(controller -> controller.updateSectionB(sectionChange.value()),
+                        () -> LOGGER.warn("No section change: shunt {} not present", sectionChange.controllerId));
+    }
+
     public void apply(LfNetworkParameters networkParameters) {
         if (tapPositionChange != null) {
-            LfBranch branch = tapPositionChange.branch();
-            int tapPosition = branch.getPiModel().getTapPosition();
-            int value = tapPositionChange.value();
-            int newTapPosition = tapPositionChange.isRelative() ? tapPosition + value : value;
-            branch.getPiModel().setTapPosition(newTapPosition);
+            applyTapPositionChange();
         }
 
         if (loadShift != null) {
-            LfLoad load = loadShift.load();
-            if (!load.isOriginalLoadDisabled(loadShift.loadId())) {
-                PowerShift shift = loadShift.powerShift();
-                load.setTargetP(load.getTargetP() + shift.getActive());
-                load.setTargetQ(load.getTargetQ() + shift.getReactive());
-                load.setAbsVariableTargetP(load.getAbsVariableTargetP()
-                        + Math.signum(shift.getActive()) * Math.abs(shift.getVariableActive()));
-            }
+            applyLoadShift();
         }
 
         if (generatorChange != null) {
-            LfGenerator generator = generatorChange.generator();
-            if (!generator.isDisabled()) {
-                double newTargetP = generatorChange.isRelative() ? generator.getTargetP() + generatorChange.activePowerValue() : generatorChange.activePowerValue();
-                generator.setTargetP(newTargetP);
-                if (!AbstractLfGenerator.checkActivePowerControl(generator.getId(), generator.getTargetP(), generator.getMinP(), generator.getMaxP(),
-                        networkParameters.getPlausibleActivePowerLimit(), networkParameters.isUseActiveLimits(), null)) {
-                    generator.setParticipating(false);
-                }
-            }
+            applyGeneratorChange(networkParameters);
         }
 
         if (hvdc != null) {
-            hvdc.setAcEmulation(false);
-            hvdc.setDisabled(true); // for equations only, but should be hidden
-            hvdc.getConverterStation1().setTargetP(-hvdc.getP1().eval()); // override
-            hvdc.getConverterStation2().setTargetP(-hvdc.getP2().eval()); // override
+            applyHvdcAction();
         }
 
         if (sectionChange != null) {
-            LfShunt shunt = sectionChange.shunt();
-            shunt.getControllers().stream().filter(controller -> controller.getId().equals(sectionChange.controllerId())).findAny()
-                    .ifPresentOrElse(controller -> controller.updateSectionB(sectionChange.value()),
-                            () -> LOGGER.warn("No section change: shunt {} not present", sectionChange.controllerId));
+            applySectionChange();
         }
     }
 }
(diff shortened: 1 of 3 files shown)


obrix:
@vmouradian This is more an adjustment of behavior than a bugfix. It has been discussed and validated that absolute generator action setting a target P should be respected and enforced.

obrix:
@Hadrien-Godard Can you have a quick look ?

Hadrien-Godard:
> @Hadrien-Godard Can you have a quick look ?

I am looking at it

vmouradian:
>  @vmouradian can you do the same fix for powershift ?
I am facing an issue here. An LfLoad aggregates all the powsybl loads that are connected to the same bus, and their targetPs are summed. 
This causes no issue for applying relative actions since you add the to the total regardless of the precise load the Action was targetting.
But to apply an absolute action, my impression is that we would need to keep track of the values of each load aggregated in the LfLoad, otherwise we cannot know what change to apply.
This will need some refactoring, but i would like to discuss with you if we want to do that and the possible ways of doing it

vidaldid-rte:
@annetill @Hadrien-Godard @rosiereflo 
     I agree this is fixing a bug for the absolute value.
    There is still an ambiguity for the relative case that may cause inconsistencies between powsybl simulators if they do not interprete it all the same way. The problem is the doc of GeneratorAction in com.powsyb.action. For relative active power change, the doc does not specify "Relative to what". 
I suggest to specify that in the doc.
The current implementaton means "relative to the actual active power production in the state to which the action is applied." (personal note: this looks strange to me because this reference is unknown before the flow is computed - it depends on the contingency and the compensation made).
But one would think "relative to the initial target P before any compensation is applied".

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
```diff
@@ -869,75 +869,75 @@ private void testGeneratorAction(boolean dc, LoadFlowParameters.BalanceType bala
...
-        assertEquals(network.getLine("l23").getTerminal1().getP(), getPostContingencyResult(result, lineInContingencyId).getNetworkResult().getBranchResult("l23").getP1(), LoadFlowAssert.DELTA_POWER);
-        assertEquals(network.getLine("l34").getTerminal1().getP(), getPostContingencyResult(result, lineInContingencyId).getNetworkResult().getBranchResult("l34").getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getLine("l12").getTerminal1().getP(), getPostContingencyResult(result, lineInContingencyId).getNetworkResult().getBranchResult("l12").getP1(), deltaPowerAssert);
+        assertEquals(network.getLine("l14").getTerminal1().getP(), getPostContingencyResult(result, lineInContingencyId).getNetworkResult().getBranchResult("l14").getP1(), deltaPowerAssert);
+        assertEquals(network.getLine("l23").getTerminal1().getP(), getPostContingencyResult(result, lineInContingencyId).getNetworkResult().getBranchResult("l23").getP1(), deltaPowerAssert);
+        assertEquals(network.getLine("l34").getTerminal1().getP(), getPostContingencyResult(result, lineInContingencyId).getNetworkResult().getBranchResult("l34").getP1(), deltaPowerAssert);
```
vidaldid-rte, line 891: @vmouradian why do you need a larger tolerance here ? The comparison is between the security analysis engine and the manual disconnection. The results should be close an independant of the change made on the generatorActions.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
```diff
@@ -839,7 +839,7 @@ void testLoadAction() {
     }
 
     private void testGeneratorAction(boolean dc, LoadFlowParameters.BalanceType balanceType, double deltaG1, double deltaG2,
```
annetill, line 841: I don't understand why you have changed the method signature and the tolerances. What "We need a larger delta power assert since we distribute some power in slack distribution steps" means?

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
```diff
@@ -869,75 +869,75 @@ private void testGeneratorAction(boolean dc, LoadFlowParameters.BalanceType bala
         parameters.setDistributedSlack(true).setBalanceType(balanceType);
         parameters.setDc(dc);
         OpenLoadFlowParameters openLoadFlowParameters = new OpenLoadFlowParameters();
-        openLoadFlowParameters.setSlackBusPMaxMismatch(slackBusPMaxMismatch);
+        openLoadFlowParameters.setSlackBusPMaxMismatch(1e-3);
         parameters.addExtension(OpenLoadFlowParameters.class, openLoadFlowParameters);
```
vidaldid-rte, line 873: There is really no reason to tolerate a larger difference between security analyses and manual modification in the network since they are supposed to be the same.
But you compare active power with a small resolution.
To ensure accurate results, use the Per Equation criteria type. This way, you avoid that errors accumulate on a terminal and cause your tests to tail, and know that for each node the error is bounded by MaxActivePowerMismatch
        openLoadFlowParameters.setNewtonRaphsonStoppingCriteriaType(NewtonRaphsonStoppingCriteriaType.PER_EQUATION_TYPE_CRITERIA);
        openLoadFlowParameters.setMaxActivePowerMismatch(1e-3);

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
Hadrien-Godard: Hello @vidaldid-rte,  I did this modification, not @vmouradian.
We need a larger tolerance since the computations are not exactly the same.

Indeed, with the manual disconnection, you run only one loadflow with different NRs separated by the slack distribution steps.
With the generator action, you run two loadflows, one before and one after action application. Each load flow has its own slack distribution steps.

So, in theory you should have exactly same results, but in practice, there is a stopping criteria on the loadflow algorithm, using the slackBusPMaxMismatch parameter. Since some non-negligeable power is distributed in those tests, depending on the value of this parameter, you end up with some differences between manual disconnection application and action application results.

vidaldid-rte: @Hadrien-Godard 
I believe that the issue is the stopping criteria. I see no reason why the compensations would be different.
See my comment below in changing the stopping criteria.

You can as well use that one instead. It remains a global criteria on the sum of errors but wit the size of the test network prevent an error larger than 1e-3 to accumulate on a single equation.
        openLoadFlowParameters.setNewtonRaphsonConvEpsPerEq(1e-5);

Hadrien-Godard: It seems to work well with 
openLoadFlowParameters.setNewtonRaphsonStoppingCriteriaType(NewtonRaphsonStoppingCriteriaType.PER_EQUATION_TYPE_CRITERIA);
openLoadFlowParameters.setMaxActivePowerMismatch(1e-3);

I push the modification.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
Hadrien-Godard: Is it ok now?

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
Hadrien-Godard: I guess that the second question is solved with inputs from @vidaldid-rte.

I changed the method signature because it seems not very satisfying to have the slack active power mismatch that depends on the test. It seems better to use the NewtonRaphsonStoppingCriteriaType.PER_EQUATION_TYPE_CRITERIA criterion as @vidaldid-rte suggested.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
vidaldid-rte: yes ! Thanks for the modifcation.

vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -224,16 +224,10 @@ private static Optional<LfAction> create(GeneratorAction action, LfNetwork lfNet
...
-                    deltaTargetP = activePowerValue.getAsDouble() / PerUnit.SB;
-                } else {
-                    deltaTargetP = activePowerValue.getAsDouble() / PerUnit.SB - generator.getInitialTargetP();
-                }
-                var generatorChange = new GeneratorChange(generator, deltaTargetP);
+                var generatorChange = new GeneratorChange(generator, activePowerValue.getAsDouble() / PerUnit.SB, relativeValue.get());
                 return Optional.of(new LfAction(action.getId(), null, null, null, null, generatorChange, null, null));
             } else {
-                throw new UnsupportedOperationException("Generator action: configuration not supported yet.");
+                throw new UnsupportedOperationException("Generator action on " + action.getGeneratorId() + " : configuration not supported yet. Active power value and boolean for relative value are required.");
```
annetill, line 230: The previous message was more correct: a generator action is really rich, we only support active power change, in an absolute way (only the double is required) or in a relative way (both double and boolean is required).

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
```diff
@@ -869,7 +870,9 @@ private void testGeneratorAction(boolean dc, LoadFlowParameters.BalanceType bala
         parameters.setDistributedSlack(true).setBalanceType(balanceType);
         parameters.setDc(dc);
         OpenLoadFlowParameters openLoadFlowParameters = new OpenLoadFlowParameters();
-        openLoadFlowParameters.setSlackBusPMaxMismatch(slackBusPMaxMismatch);
+        openLoadFlowParameters.setNewtonRaphsonStoppingCriteriaType(NewtonRaphsonStoppingCriteriaType.PER_EQUATION_TYPE_CRITERIA);
+        openLoadFlowParameters.setMaxActivePowerMismatch(1e-3);
+        openLoadFlowParameters.setSlackBusPMaxMismatch(1e-3);
```
annetill, line 875: So it is less precise than before? I think we have to do a dedicated PR because our unit tests are not homogeneous when we compare LFs and ASs : not same methodology, not same criteria, not same tolerance...

src/main/java/com/powsybl/openloadflow/network/LfAction.java
Hadrien-Godard: I agree.
I suggest to delete the misleading addition and to keep the addition on the generator Id.
I have pushed a commit.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
Hadrien-Godard: With this version, it is as precise as LF tests (and as before) since it is using LoadFlowAssert.DeltaPower to check active power flows differences.

The setting of the "slack bus Pmax mismatch" parameter is here to ensure that we can obtain results precise enough to validate the comparison with LoadFlowAssert.DeltaPower assert.

Maybe I miss something, but I do not see issues to launch tests with different parameters since the asserted values are compared with the same thresholds.

annetill: It is just for other developers not to "loose" time with these issues.

annetill (approved):

rosiereflo (approved):

PR #1028, opened 2024-05-10
Avoid slack distribution to fictitious loads

caioluke:

Feature.

Currently, LfLoads do not have the method isFictitious like LfGenerators do. Therefore, when distributing slack on loads, we do not take into account if its fictitious or not.

Avoid slack distribution to fictitious loads.

- [ ] Yes
- [X] No


src/main/java/com/powsybl/openloadflow/network/impl/LfLoadImpl.java
```diff
@@ -73,6 +73,17 @@ public LfBus getBus() {
         return bus;
     }
 
+    @Override
+    public boolean isFictitious() {
```
annetill, line 77: It is not so simple... the changes must be at lower level, when we compute the participation factor using `getAbsVariableTargetP()`. Note that this method is used at several places, for reach you must ask yourself if you need to change something or not. You also have to change the state update.

annetill: And other remark, in iidm model, we have two ways to say that a load is fictitious (and I don't know why), through the type and through the boolean you are using. Please check both with a ||.

annetill:
Applying a contingency:
````
for (var e : lostLoads.entrySet()) {
    LfLoad load = e.getKey();
    LfLostLoad lostLoad = e.getValue();
    PowerShift shift = lostLoad.getPowerShift();
    load.setTargetP(load.getTargetP() - getUpdatedLoadP0(load, balanceType, shift.getActive(), shift.getVariableActive()));
    load.setTargetQ(load.getTargetQ() - shift.getReactive());
    load.setAbsVariableTargetP(load.getAbsVariableTargetP() - Math.abs(shift.getVariableActive()));
    lostLoad.getOriginalIds().forEach(loadId -> load.setOriginalLoadDisabled(loadId, true));
}
````
Here, the idea is to remove the effect of a load on a bus. If the load is fictitious, be careful:
- Is the shift ok for a fictitious load, knowing that the variable part is used for slack distribution.
- The updated load P0 that reflects the slack bus distribution on pre-contingency state must adapted?
- The abs variable targetP must be adapted? 
If you prefer, you can do a unit test on a LfNetwork and applying a contingency and check the values. See `LfNetworkTest`.

Same analysis must be done for actions.

File: src/main/java/com/powsybl/openloadflow/network/LfAction.java [hunk 2]
```diff
@@ -8,10 +8,7 @@
 package com.powsybl.openloadflow.network;
 
 import com.powsybl.action.*;
-import com.powsybl.iidm.network.Bus;
-import com.powsybl.iidm.network.Load;
-import com.powsybl.iidm.network.Network;
-import com.powsybl.iidm.network.Terminal;
+import com.powsybl.iidm.network.*;
 import com.powsybl.openloadflow.graph.GraphConnectivity;
```
vidaldid-rte, line 12: Did you remove the explicit import on purpose ?
Looks like many other files in this project don't use wilcards on import.
In case you use IntelliJ you may see  https://sergiodelamo.com/blog/intellij-idea-disable-wildcard-imports.html for a configuration that does not put wildcards on import event with more than 3 classes imported from a package.

This is cleaner in cases where the same class name can be user in different packages.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -34,7 +31,7 @@ public final class LfAction {
     private record TapPositionChange(LfBranch branch, int value, boolean isRelative) {
     }
 
-    private record LoadShift(String loadId, LfLoad load, PowerShift powerShift) {
+    private record LoadShift(Load load, LfLoad lfLoad, PowerShift powerShift) {
     }
```
vidaldid-rte, line 35: You should not keep iidm.network objects in memory since fast restart is implemented (https://github.com/powsybl/powsybl-open-loadflow/pull/635)
If you need to keep the Load in a record, use a WeakRef<Load> instead.
Otherwise, this will prevent the GC from deleting a Network if the fast restart mode is on.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -319,16 +316,20 @@ private void applyTapPositionChange() {
...
-                    + Math.signum(shift.getActive()) * Math.abs(shift.getVariableActive()));
+            lfLoad.setTargetP(lfLoad.getTargetP() + shift.getActive());
+            lfLoad.setTargetQ(lfLoad.getTargetQ() + shift.getReactive());
+            lfLoad.setAbsVariableTargetP(lfLoad.getAbsVariableTargetP() + getUpdatedLoadAbsVariableTargetP(load, shift));
         }
     }
 
+    private static double getUpdatedLoadAbsVariableTargetP(Load load, PowerShift shift) {
+        return load.isFictitious() || LoadType.FICTITIOUS.equals(load.getLoadType()) ? 0.0 : Math.signum(shift.getActive()) * Math.abs(shift.getVariableActive());
+    }
```
vidaldid-rte, line 331: This implementation might be confusing to some users because it silently does something else than the user asked.
If the user created a LoadShift with a fictive load, why do you want not to honnor this ?

In case you really want to do this, it would bet better to generate and error when the LoadShift is created with a fictive load, or at least Warning in the log.

Also it the facto of ignoring fictive load a feature desired by ALL users of load compensation ? If not this  implementation that is not controlled by a parameter  might be considered as a breaking change.

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
```diff
@@ -149,7 +149,7 @@ public void apply(LoadFlowParameters.BalanceType balanceType) {
             PowerShift shift = lostLoad.getPowerShift();
             load.setTargetP(load.getTargetP() - getUpdatedLoadP0(load, balanceType, shift.getActive(), shift.getVariableActive()));
             load.setTargetQ(load.getTargetQ() - shift.getReactive());
-            load.setAbsVariableTargetP(load.getAbsVariableTargetP() - Math.abs(shift.getVariableActive()));
+            load.setAbsVariableTargetP(load.getAbsVariableTargetP() - getUpdatedLoadAbsVariableTargetP(load, shift.getVariableActive()));
             lostLoad.getOriginalIds().forEach(loadId -> load.setOriginalLoadDisabled(loadId, true));
```
vidaldid-rte, line 153: Why do you need to do somthing special for fictive load in setAbsVariableTargetP for firktive loads since the getAbsVariableTargetP is harcoded for fiktive load , by this new code::
   private double getAbsVariableTargetP(Load load) {
        if (isLoadFictitious(load)) {
            return 0.0;
        }

src/main/java/com/powsybl/openloadflow/network/LfAction.java
caioluke: Changed it back to what it was
Indeed, IntelliJ automatically swapped multiple imports for a wildcard import

src/main/java/com/powsybl/openloadflow/network/LfAction.java
caioluke: The shift is applied for any type of load.
```java
lfLoad.setTargetP(lfLoad.getTargetP() + shift.getActive());
lfLoad.setTargetQ(lfLoad.getTargetQ() + shift.getReactive());
lfLoad.setAbsVariableTargetP(lfLoad.getAbsVariableTargetP() + getUpdatedLoadAbsVariableTargetP(loadIsFictitious, shift));
```

The only difference is that the `absVariableTargetP` is not updated if the load is fictitious.
From what I understood, this `absVariableTargetP` is only used for slack distribution on loads.
Therefore, it is aligned with what I'm proposing.

Your point is still valid though: maybe I should also add a parameter to define if the user wants to distribute slack on fictitious loads or not.
I assumed that would always be the case, but maybe it is not...

What do you think about that?

src/main/java/com/powsybl/openloadflow/network/LfContingency.java
caioluke: Ha, you are right.
I didn't realize that.
Reverted back to what it was.

src/main/java/com/powsybl/openloadflow/network/LfAction.java
vidaldid-rte: I think there are two issues. One for Actions, and one for compensation.
- For Actions, I would consider that if a user created an action on a fictive load, it is his decision and the software should just do it. 
If however there are good arguments to say that this can only be an error (I do not see them) the software should not silently ignore this. Choices are then throwing an Exception or printing a warning/error in the log. But not silently doing something else.

For compensation, they key is to know if some users  have valid use cases to compensate on fictive loads. I don't hace the answer. If nobody has a need, it's OK to make it a default BUT AT THE MINIMUM make the change visible in the relase notes  If valid use cases to compensate on fictive loads exist then you have, unfortunately,  to add a parameter to the 80 or more already existing... ANd to select a good default.

vidaldid-rte: Thanks for the explanation. Talking of several loads on a bus there are other events that may change the list of loads during during security analysis (disconnect / disable events). Have you taken this into account ? I will try to review the code with this in mind and see if there is a way to centralize more this logic).

src/main/java/com/powsybl/openloadflow/network/LfAction.java
vidaldid-rte: Luke,
  I stepped your test and it happens that when the contingency is processed a PowerShift  is created with a Variable active power. This happens in PropagatedContingency.getLoadPowerShift.
 
  I am not very familiar with load compensation, but it looks to me as a bug.The contingency on a fictive load should not add a shift of variable active power for a fictive load, in my understanding.

 So it seems that we already have duplicated code and several places to compute the active power of an IIDM Load.
 It think it would would be better if there was a single function that returned this value in the LF wordls, that would contain the  logic on fictive in this unique place. And that NO other place of OpenLoadFlow computes the variable power of a load.

caioluke: You are right, there is a problem in my implementation.

The test now reflects it.

I think I have to add the isFictitious information in the LfLostLoad object...

And then handle each lost load accordingly.

What do you think about it?

vidaldid-rte: For me the ideal solution would be that the logic that computes the variable active power of a load is in only one place.
That means for example that if in the future, you need to add a parameter to take fictive into account or not, you would have only one function to modify.
The the logic is spread in different places, for actions, for contingnencies etc... there is a maintainaility issue and a risk that if in the future that logic has to be modified it is done only in one place.

caioluke: Hi @vidaldid-rte, I have just fixed the behaviour. We have something that is working now.

Could you have a look at this final version and tell me what you think?

I agree with you that the fictitious load information is all over the place (contingency, action), but I don't see another way of doing it...
I would gladly hear your suggestions about that.

caioluke: It looks much better!!! Thank you :)
I have only one comment: don't you think it's better to name isOriginalLoadPassive instead of isIidmLoadPassive? It is coherent with the other methods' naming, and it doesn't mention a data model

vidaldid-rte: No problem for the renaming. Go ahead !

vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/network/LfLostLoad.java
```diff
@@ -30,6 +30,22 @@ public Set<String> getOriginalIds() {
         return ids;
     }
 
+    /**
+     * Updates the contribution of loads that do not participate to slack distribution
+     */
+    public void updateNotParticipatingLoad(LfLoad load, String originalLoadId, PowerShift powerShift) {
```
annetill, line 36: @vidaldid-rte I suggest to change to "updateNotParticipatingLoadP0"

src/main/java/com/powsybl/openloadflow/network/impl/LfLoadImpl.java
```diff
@@ -30,15 +30,15 @@ public class LfLoadImpl extends AbstractLfInjection implements LfLoad {
...
+    private final Map<String, Ref<Load>> loadsRefs = new HashMap<>();
 
     private final List<Ref<LccConverterStation>> lccCsRefs = new ArrayList<>();
 
     private double targetQ = 0;
 
     private boolean ensurePowerFactorConstantByLoad = false;
 
-    private final List<Double> loadsAbsVariableTargetP = new ArrayList<>();
+    private final HashMap<String, Double> loadsAbsVariableTargetP = new HashMap<>();
```
caioluke, line 41: Do we need to define the object as the implementation class? @vidaldid-rte 

`Map<String, Double>` instead of `HashMap<String, Double>`

vidaldid-rte: You lean define it as a Map ? It avoids some pedantic discussions with Sonar lovers so why not, although it does not really matter in a private field definition.
There is still an issue with that - in theory it slows down iterations on the array (that happen frequently) because whe would break the processor cache here. I don't know how much it matters in real perf.
But we may revert this change and instead add commentarys that it is important to update the array and the map in sync because the code relies on the order.

annetill (approved):

PR #1034, opened 2024-05-21
Migrate to core 6.4.0

olperr1:

No

Dependency update

powsybl-open-loadflow depends on powsybl-core 6.3.0.

powsybl-open-loadflow depends on powsybl-core 6.4.0.

- [ ] Yes
- [X] No

**Other information**:


src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -348,14 +348,9 @@ public static ReportNode createRootLfNetworkReportNode(int networkNumCc, int net
...
-                .withMessageTemplate("lfNetwork", "Network CC${networkNumCc} SC${networkNumSc}")
-                .withUntypedValue(NETWORK_NUM_CC, networkNumCc)
-                .withUntypedValue(NETWORK_NUM_SC, networkNumSc)
-                .add();
-        newReportNode.include(lfNetworkReportNode);
-        return newReportNode;
+    public static ReportNode includeLfNetworkReportNode(ReportNode reportNode, ReportNode lfNetworkReportNode) {
+        reportNode.include(lfNetworkReportNode);
+        return lfNetworkReportNode;
     }
```
vidaldid-rte, line 354: @olperr1 why is this change part of an integration branch ? I didn't find the API change in core that requries to change this behaviour.

src/test/java/com/powsybl/openloadflow/sa/AbstractOpenSecurityAnalysisTest.java
```diff
@@ -29,11 +29,7 @@
 import com.powsybl.openloadflow.network.LfBus;
 import com.powsybl.openloadflow.network.NodeBreakerNetworkFactory;
 import com.powsybl.openloadflow.network.SlackBusSelectionMode;
-import com.powsybl.security.LimitViolationFilter;
-import com.powsybl.security.SecurityAnalysisParameters;
-import com.powsybl.security.SecurityAnalysisReport;
-import com.powsybl.security.SecurityAnalysisResult;
-import com.powsybl.security.detectors.DefaultLimitViolationDetector;
+import com.powsybl.security.*;
 import com.powsybl.security.limitreduction.LimitReduction;
```
vidaldid-rte, line 33: @olperr1 is this on purpose ? (just to be sure - I have no strong opinion about explicit or implicit import - but some projects prefer to always explicit the imports)

vidaldid-rte (commented):
Just one question about a change in report that doesn't seem releated to integration with core. And one about collapsing import to be sure this was on purpose.

src/main/java/com/powsybl/openloadflow/util/Reports.java
olperr1: It's an adaptation for https://github.com/powsybl/powsybl-core/pull/3007.
When a `ReportNode` was included, its root was omitted and only its children were in fact included. The powsybl-core PR fixed that, so I reverted the workaround that was made in OLF.

src/test/java/com/powsybl/openloadflow/sa/AbstractOpenSecurityAnalysisTest.java
olperr1: It wasn't on purpose. I will revert this change.
Thanks.

vidaldid-rte (approved):
Thanks Olivier for your clarifications. All good for me.

src/main/java/com/powsybl/openloadflow/util/Reports.java
jeandemanged: Yes indeed at the time I made this code in #919 I went a bit fast and didn't fully realize it was actually a powsybl-core bug I was workaround-ing here ...

Only when I saw https://github.com/powsybl/powsybl-core/pull/3007 I understood this wasn't right and noted here: https://github.com/powsybl/powsybl-open-loadflow/pull/919#discussion_r1602876818

thanks @olperr1 for the fix

geofjamg (approved):

PR #1032, opened 2024-05-17
Improve transformer voltage control `AFTER_GNERATOR_VOLTAGE_CONTROL`

vidaldid-rte:
**Please check if the PR fulfills these requirements**

- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?**
In order to provide a ratio tap change simulation that better matches our Network we introduce the following evolutions to the AfterVoltageControlOuterLoop:

- Taps are not discretized before all taps are within their operating range. Instead, if a continuous ratio is outside of operating range, it is discretized at its limit and a loadflow is run again to update the continous ratios.
 - A new parameter ("transformerVoltageControlUseInitialTapPosition") minimize the tap changes in the following way:
   - If the voltage of a bus of parallel tranformers is within range, the taps of the parallel transformers are not changed, even if they are not the closest tap to the continuous ratio.
    - If the tap need to be changed but parallel transformers have different ratios initially, the difference is kept at discretization.
    - This simulates the behavior of independent automates for parallel transformers. Additionally it can preserve some ration misalignment that could be done on purpose to consume some reactive power (due to a reactive current loop created by parallel  transformers with misaligned ratios)
    -  If the mode is enabled, transformer that have no group controlling tension above them that is PV at computation time are excluded from tap computation.
- The parameter "generatorVoltageControlMinNominalVoltage" (introduced in https://github.com/powsybl/powsybl-open-loadflow/pull/1038 )  is  used to define the nominal voltage below which generators are set to PQ while continuous ratio are computed. 
- Additionnally, generators that are just behind a step-up transformer that is above the limit continue to control voltage. This enable to support network models where generators are separated from the transformer connecting them to the network.

The default value of the new parameter (transformerVoltageControlUseInitialTapPosition) keep a behavior close to the previous behavior of TransformerVoltageControlOuterLoop.

**What is the current behavior?**

**What is the new behavior (if this is a feature change)?**
With current default of new parameters, almost same as before.
If transformerVoltageControlStable is set to true, simulation much closer to actual behaviour of our network.

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [X] No

**What changes might users need to make in their application due to this PR? (migration steps)**

PR #1038, opened 2024-05-28
Fix transformer voltage control outer loop AFTER GENERATOR

annetill:

No, but it occurs on real RTE network where now we have some transformers with a voltage control on 225 kV (transformers used for off shore wind farm for example). So the highest nominal voltage controlled by transformers is quite high and too high to determine which generators must have its voltage control disabled during outer loop.  

Just a proposal, I have no good idea of the good fix. I would like an automatic way, that could analysed generator voltage controls or transformer voltage controls, or both. But we need also a parameter.

We must include in this PR all the good idea of #1032 : parameter, VSC management, etc.

image

- [ ] Yes
- [ ] No

**Other information**:


src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -118,6 +118,8 @@ public enum SlackDistributionFailureBehavior {
 
     public static final boolean WRITE_REFERENCE_TERMINALS_DEFAULT_VALUE = true;
 
+    protected static final double GENERATOR_VOLTAGE_CONTROL_MIN_NOMINAL_VOLTAGE_DEFAULT_VALUE = 0;
+
```
vidaldid-rte, line 122: @annetill why is the default 0 ? (and not -1 for automatic ? )

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -1153,6 +1160,20 @@ public OpenLoadFlowParameters setVoltageTargetPriorities(List<String> voltageTar
         return this;
     }
 
+    /**
+     * Only if transformer voltage control is active and with mode `AFTER_GENERATOR_VOLTAGE_CONTROL`. Set the nominal
+     * voltage under which the generator voltage control are disabled during outer loop. This parameter overrides the
+     * automatic nominal voltage computation if > 0.
+     */
```
vidaldid-rte, line 1167: @annetill  the ">0" contradicts the parameter doc where auto is for <0.  Setting the parameter to 0, which means keeping all generators in tension control is interesting and should be allowed.

src/main/java/com/powsybl/openloadflow/ac/outerloop/TransformerVoltageControlOuterLoop.java
```diff
@@ -53,15 +87,9 @@ public void initialize(AcOuterLoopContext context) {
...
-        double[] maxControlledNominalVoltage = new double[1];
-        maxControlledNominalVoltage[0] = Double.MIN_VALUE;
-        for (LfBus bus : context.getNetwork().getBuses()) {
-            if (!bus.isDisabled() && bus.isTransformerVoltageControlled()) {
-                maxControlledNominalVoltage[0] = Math.max(maxControlledNominalVoltage[0], bus.getNominalV());
-            }
-        }
-        ((ContextData) context.getData()).setMaxControlledNominalVoltage(maxControlledNominalVoltage[0]);
+        // all transformer voltage control are disabled for the first equation system resolution.
+        double maxControlledNominalVoltage = maxControlledNominalVoltageOverride > 0 ? maxControlledNominalVoltageOverride : calculateMaxControlledNominalVoltage(context);
```
vidaldid-rte, line 91: @annetill again here there auto should be -1. You should authorize 0 to keep all generators in voltage control.

src/main/java/com/powsybl/openloadflow/AbstractAcOuterLoopConfig.java
```diff
@@ -56,11 +56,11 @@ protected static Optional<AcOuterLoop> createMonitoringVoltageOuterLoop(OpenLoad
     }
 
     protected static Optional<AcOuterLoop> createTransformerVoltageControlOuterLoop(LoadFlowParameters parameters, OpenLoadFlowParameters.TransformerVoltageControlMode controlMode,
-                                                                                    int incrementalTransformerVoltageControlOuterLoopMaxTapShift) {
+                                                                                    int incrementalTransformerVoltageControlOuterLoopMaxTapShift, Double generatorVoltageControlMinNominalVoltage) {
```
vidaldid-rte, line 59: @annetill why a Double and not a double ? Is see no interrest to support the case where the value is null (and by the wy it would crash in the current code).

vidaldid-rte (changes requested):
Ambiguity in the value for automatic computation of limit (0 or -1).
I think that 0 should mean NO LIMIT, and should not be used for AUTO.

Some other minor comments.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
geofjamg: why -1 would be better that 0 ?

geofjamg:
> Ambiguity in the value for automatic computation of limit (0 or -1). I think that 0 should mean NO LIMIT, and should not be used for AUTO.
> 
> Some other minor comments.

What could be the purpose of no limit ?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
vidaldid-rte: as discussed in other places, to allow no limit (why not allowing this ?  :-) )

vidaldid-rte:
> > Ambiguity in the value for automatic computation of limit (0 or -1). I think that 0 should mean NO LIMIT, and should not be used for AUTO.
> > Some other minor comments.
> 
> What could be the purpose of no limit ?

Not much sense according to the algo logic, but from my experiments it works well and without convergence issues.

vidaldid-rte (approved):

geofjamg (approved):

PR #1039, opened 2024-05-31
Fix cache invalidation when change on another component

geofjamg:

No

Bug fix

When an element is changed and not on a Lf network of the cache the cache is invalidated. This is a mistake because sometimes we can just have a change on a valid equipment but present on another component.  

- [ ] Yes
- [ ] No

**Other information**:


src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -316,47 +317,44 @@ public void onExtensionUpdate(Extension<?> extension, String attribute, Object o
...
             }
 
-            if (!done[0]) {
-                reset();
-            }
+            processUpdateResult((Identifiable<?>) extension.getExtendable(), attribute, result);
         }
 
-        private void onSecondaryVoltageControlExtensionUpdate(SecondaryVoltageControl svc, String attribute, Object newValue, boolean[] done) {
+        private CacheUpdateResult onSecondaryVoltageControlExtensionUpdate(SecondaryVoltageControl svc, String attribute, Object newValue) {
```
annetill, line 329: You don't have any element not found for extensions.

geofjamg: Fixed

annetill (approved):

PR #1042, opened 2024-06-04
DC security analysis: RHS is not invalidated

annetill:

Yes, https://github.com/powsybl/pypowsybl/issues/758

Nothing for the moment, just a unit test that shows the bug. In DC security analysis only, equations can have rhs. If the rhs of these equations changed, because in contingency for example, but it could be a tap position change of a P0 change for HVDC, we have to invalidate the DC target vector. It is not the case on main branch. 

We have the issue just today because it was hidden before through the sensitivity analysis API.   

- [ ] Yes
- [ ] No

**Other information**:


vidaldid-rte (approved):
Looks good to me.
Maybe you could add a similar test for AC (although it should logically not be subject to this bug)

geofjamg (approved):

PR #1043, opened 2024-06-04
Implement minTargetP and maxTargetP from ActivePowerControl extension

vidaldid-rte:
It implements https://github.com/powsybl/powsybl-core/pull/3054 for OpenLoadFlow

In some situations the actual pMin and pMax that can be used for Active Power Operation is different from the group capabilities.
https://github.com/powsybl/powsybl-core/pull/3054 enables to capture these temporary active power limitations in the AtivePowerOperation extension.

This PR implements this override in openloadflow.

The generators genuine pmin and pmax are still used in operations that are not related to active power operations (such as deetermining if a generator is really up and can control voltage).
But in all situations related to power operations such as slakc distribution, generator change and related details such as slack bus selection, the pMin and PMAx can be overridden by the Active Power Operation extension. 
At implementation, these overridable values are called activePowerOperationMinP and activePowerOperationMaxP

Impossible to define a temporary limit for Pmin and PMax to be used in active power operations.

No change by default, but if pMin and pMax are overridden in the Active Power Operations, these values are taken into account for compensation operations and generator actions (if p limits are to be checked for these actions).

- [ ] Yes
- [X] No

Application developers should be aware the pMin and pMax can be overriden in the Active Power Control Operation Extension
for example to enable users to change temorary pMin or pMax,  to display those limits, or to import generator data fro exetrnal datasources..


jeandemanged (dismissed):
please align namings, (_both_ in PR description and in code), this would make things consistent / less confusing:
- the iIDM extension is ActivePowerControl, not ActivePowerOperation
- activePowerOperationMinP contains redundant information "activePower" and "P", same for activePowerOperationMaxP. Please find a better name (why not just name maxPOverride just like in the iIDM extension?)
- in the iIDM extension we have minP**Override**, why change in OLF the name to **operation**MinP ?
- in PR description is mentionned also "temporary" min/max P, now we have 3 terms override, temporary, operational
- I have the list of where the override is applied if existing, I lack the list of the places where it is not used - and the rationale for not using it.

vidaldid-rte:
>     * I have the list of where the override is applied if existing, I lack the list of the places where it is not used - and the rationale for not using it.
Before working on naming, vocabulary standardization and synonym eradication, I'd like to first address this question. 

The purpose of this PR is that we have for generators some pmin and pmax that we consider the specification of the generator, independant of any condition, and some other limits (called in the extension pMinOverride and pMaOverride) that specify how much the generator can move in the current context.
We made the decision to:
     - keep the generator specifications in the data as pmin and pmax
     - if more restrictive limites exist, add them in the extension that controls compensation
     - be agnostic about the cause of the restriction. It is just the degree of freedom for compensation  for the timeframe and horizon of a simulation., in general expected to be around target P,

So we end up with the two concepts:
   The absolute maxP/minP that are a characteristic of the Generators
   The effective maxP/minP (currently called  activePowerOperationMaxP/minP) that are the limits of what can be used in the scope of the simulation.

In this PR, the usage of the limits are as follow. The choices can be challenged and discussed - except for compensation where it is important to use the "effective" maxP/minP 

slackBusSelection:    absolute maxP
isGeneratorInvalid :   absolute maxP
limits for volage control:   absolute maxP/minP
check stared for voltage control: absolute minP

compensation: effective minP/maxP
generator actions: if limits are checked, it is against effective maxP/minP

jeandemanged:
> > ```
> > * I have the list of where the override is applied if existing, I lack the list of the places where it is not used - and the rationale for not using it.
> > ```
> 
> Before working on naming, vocabulary standardization and synonym eradication, I'd like to first address this question.
> 
> The purpose of this PR is that we have for generators some pmin and pmax that we consider the specification of the generator, independant of any condition, and some other limits (called in the extension pMinOverride and pMaOverride) that specify how much the generator can move in the current context. We made the decision to: - keep the generator specifications in the data as pmin and pmax - if more restrictive limits exist, add them in the extension that controls compensation - be agnostic about the cause of the restriction. It is just the degree of freedom for compensation for the timeframe and horizon of a simulation., in general, targetP is expected to be between those limits.
> 
> So we end up with the two concepts: The absolute maxP/minP that are a characteristic of the Generators The effective maxP/minP (currently called activePowerOperationMaxP/minP) that are the limits of what can be used to modify targetP in the scope of the simulation.
> 
> In this PR, the usage of the limits are as follow. The choices can be challenged and discussed - except for compensation where it is important to use the "effective" maxP/minP
> 
> slackBusSelection: absolute maxP isGeneratorInvalid : absolute maxP limits for volage control: absolute maxP/minP check stared for voltage control: absolute minP
> 
> compensation: effective minP/maxP generator actions: if limits are checked, it is against effective maxP/minP

----

Thanks, clarifies what you want to achieve.

You mention:

> if more restrictive limits exist

Looking at https://github.com/powsybl/powsybl-core/pull/3054 there is not concept here of _more restrictive limits_, only an override, which by its naming alone suggests a complete replacement: i.e. it can restrict the range, but can also extend the range, and can also replace with a completely disjoint range.

I am OK with the choices you made on when to use generator Pmin/Pmax vs. when to use generator Pmin/Pmax override. I am OK also for both override possible interpretations:
a. it is an override = it replaces the generator value
b. it is a restriction = it replaces the generator value only if more restrictive

Currently a. is implemented in this PR, please confirm this is what you ant to do. If you want b. instead some of the code needs to be revisited (also maybe on the powsybl-core extension side).

vidaldid-rte:
> You mention:
> 
> > if more restrictive limits exist
> 
> Looking at [powsybl/powsybl-core#3054](https://github.com/powsybl/powsybl-core/pull/3054) there is not concept here of _more restrictive limits_, only an override, which by its naming alone suggests a complete replacement: i.e. it can restrict the range, but can also extend the range, and can also replace with a completely disjoint range.

Yes. That is a very good point.
My immediate use case is a restriction. But I decided to name it and use it as an override since we can imagine other use cases where by exception you authorize to go beyond normal generators limits.

If you agree with that, then I can propose naming as a next steps.

vidaldid-rte:
> Naming proposal for what I called effective pmin/pmax earlier:
targetPMax / targetPMin.

vidaldid-rte:
Renaming done. Won't compile until new release of powsyble-core is done that includes the renaming in the extension.
Criteria to check that targetP is within bounds for checkActivePowerControl now uses min/maxTargetP.
Checks that min/maxTargetP is in [minP,maxP] is delegated to the ActivePowerControl extension that does this in a robust way.

src/main/java/com/powsybl/openloadflow/network/LfGenerator.java
```diff
@@ -72,6 +72,22 @@ static double qToK(LfGenerator generator, double q) {
 
     void setTargetP(double targetP);
 
+    /**
+     * The minimum target P for active power operations (can be different from minP if minTargetP is set in the ActivePowerControl extension)
```
annetill, line 76: operations? What do you mean?
compensation -> use slack bus distribution.

src/main/java/com/powsybl/openloadflow/network/LfGenerator.java
```diff
@@ -72,6 +72,22 @@ static double qToK(LfGenerator generator, double q) {
...
+     * This limit is taken into account in the compensation.
+     */
+    default double getMinTargetP() {
+        return getMinP();
+    }
+
+    /**
```
annetill, line 83: Same comment.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -419,7 +419,7 @@ public double getLoadTargetQ() {
 
     @Override
     public double getMaxP() {
```
annetill, line 421: It seems to be used only here:
image
So we prefer to use the generator min and max P restrictions here?

vidaldid-rte: @annetill I also think it is onnly used here ans this is the only justification to add minTargetP and maxTargetP in LFGenerator instead of only using minP/maxP that would be overriden.

To really know what to do here, if you are unsure of my interpretation, maybe be should ask the features author or someone who uses that feature. I assumed that the intention is to not start voltage control if the generator is outside of the generator limits minP/macP and not outside of the authorized limites to move targetP (minTargetP/maxTargetP).

src/main/java/com/powsybl/openloadflow/network/impl/LfBatteryImpl.java
```diff
@@ -93,6 +93,20 @@ public double getMaxP() {
         return getBattery().getMaxP() / PerUnit.SB;
     }
 
+    @Override
+    public double getMinTargetP() {
+        Battery battery = getBattery();
```
annetill, line 98: Why do you prefer to do everything here rather than do it at creation through 2 attributes? Same remark for generator impl.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
annetill: @geofjamg what do we have to do here when we force to keep one PV bus in the network?

annetill (approved):
Thanks a lot @jeandemanged, as usual.

PR #1048, opened 2024-06-14
Ac emulation outerloop

Hadrien-Godard:

No

This PR introduces an outerloop to model the ability of AcEmulationControl to be bounded by Pmax and to change feeding side.

Currently, those abilities are directly model in equation terms, leading to discontinuity in the computed derivatives.

An outerloop is introduced to control that an HVDC line in AcEmulation mode can be bounded by its Pmax values and that a loss coefficient is applied at the fed side.

- [ ] Yes
- [x ] No


src/main/java/com/powsybl/openloadflow/DefaultAcOuterLoopConfig.java
```diff
@@ -36,6 +36,8 @@ public List<AcOuterLoop> configure(LoadFlowParameters parameters, OpenLoadFlowPa
         createTransformerReactivePowerControlOuterLoop(parametersExt).ifPresent(outerLoops::add);
         // shunt compensator voltage control
         createShuntVoltageControlOuterLoop(parameters, parametersExt).ifPresent(outerLoops::add);
+        // AC emulation
+        createAcEmulationOuterLoop(parameters).ifPresent(outerLoops::add);
         // automation system
```
vidaldid-rte, line 41: I'm wondering where this loop should be. It seems to me that it is better placed right after the dsitributed slack as the flow of active current is what would mostly drive the angles. And also that any voltage control computation made with a wrong direction of power flow in an HVDC is of little.  @geofjamg  any thought on this topic ?

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcEmulationOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+        LfHvdc.AcEmulationControl acEmulationControl = hvdc.getAcEmulationControl();
+
+        if (acEmulationControl.getFeedingSide() == TwoSides.ONE) {
+            if (hvdc.getP1().eval() < 0) {
+                // Switch feeding side
+                LOGGER.trace("Switching feeding side from One to Two for Hvdc: {}", hvdcId);
+                contextData.incrementFeedingSideSwitchCount(hvdcId);
+                hvdc.updateFeedingSide(TwoSides.TWO);
+                if (contextData.getFeedingSideSwitchCount(hvdcId) == MAX_FEEDING_SIDE_SWITCH) {
+                    LOGGER.debug("Two many feeding side switches (flow blocked to 0 MW) for Hvdc: {}", hvdcId);
```
vidaldid-rte, line 50: I have a problem with this NULL status.
In the end it would produce a solution with status FULLY CONVERGED and where the HVDC transports no power.
But it may not represent the reality and be misleading. And cause users of the simulator to make wrong decisions.

For example the flip fop may result from changes of Phase Tap changers, or other automates, that changes the active power routes. It is not necessarily that the active power in the HVDC has small oscillations around 0 and can be neglected.

Have you found examples where small endless oscillations happen here ? Wouldn't a dead angle zone for which P is 0 be better ?

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcEmulationOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Override
+    public OuterLoopResult check(AcOuterLoopContext context, ReportNode reportNode) {
+        OuterLoopStatus status = OuterLoopStatus.STABLE;
+        ContextData contextData = (ContextData) context.getData();
+
+        for (LfHvdc hvdc : context.getNetwork().getHvdcs()) {
+            if (!hvdc.isAcEmulation() || hvdc.getBus1().isDisabled() || hvdc.getBus2().isDisabled() || hvdc.isDisabled()) {
+                continue;
```
vidaldid-rte, line 131: What you really want to know here is whether the hvdc link is active. That is if the hvdc in AC emulation mode and the  P1 and P2 equations terms are active.
This condition here looks like a code duplication and may create problems in the future if we need to change the conditions and notifications that activate/deactivate P1 and P2 for LfHDVC.
Why not making getP1 and getP2 returning an EquationTerm instead of an evaluable and checking their active status ?
That would ensure consistency.

src/main/java/com/powsybl/openloadflow/dc/DcAcEmulationOuterLoop.java
```diff
@@ -0,0 +1,97 @@
...
+        if (acEmulationControl.getAcEmulationStatus() == LfHvdc.AcEmulationControl.AcEmulationStatus.FREE) {
+            // Check Pmax
+            if (hvdc.getP1().eval() > acEmulationControl.getPMaxFromCS1toCS2()) {
+                hvdc.updateFeedingSide(TwoSides.ONE);
+                // Switch mode
+                LOGGER.trace("Bound Hvdc flow to Pmax from CS1 to CS2 for Hvdc: {}", hvdcId);
+                contextData.incrementModeSwitchCount(hvdcId);
+                hvdc.updateAcEmulationStatus(LfHvdc.AcEmulationControl.AcEmulationStatus.BOUNDED);
+                if (contextData.getModeSwitchCount(hvdcId) == MAX_MODE_SWITCH) {
+                    LOGGER.debug("Two many mode switches (flow blocked to Pmax from CS1 to CS2) for Hvdc: {}", hvdcId);
```
vidaldid-rte, line 48: Here again I am concerned that we give wrong result.
Is there a way to distringuish small oscillations near the maxP (where locking at bound status is acceptable) from large oscillations that mean no convergence ?

src/main/java/com/powsybl/openloadflow/DefaultAcOuterLoopConfig.java
Hadrien-Godard: Yes, I agree.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcEmulationOuterLoop.java
Hadrien-Godard: The idea here is that, if the HVDC is changing its feeding side multiple times during the LF algorithm, then its flow should be very close to 0. In realistic case, the feeding side should never change at all during iterations.

We do not have any cases where endless oscillations around zero occur, but in theory, without this security, it may happen. It is mainly because of the discontinuity of the derivative from k to -k around 0.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcEmulationOuterLoop.java
```diff
@@ -0,0 +1,149 @@
...
+import com.powsybl.openloadflow.lf.outerloop.OuterLoopStatus;
+import com.powsybl.openloadflow.network.LfHvdc;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Hadrien Godard {@literal <hadrien.godard at artelys.com>}
+ */
+public class AcEmulationOuterLoop
+        extends AbstractAcEmulationOuterLoop<AcVariableType, AcEquationType, AcLoadFlowParameters, AcLoadFlowContext, AcOuterLoopContext>
```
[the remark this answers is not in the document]
Hadrien-Godard: Yes, I agree.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcEmulationOuterLoop.java
Hadrien-Godard: Ok I will try that

src/main/java/com/powsybl/openloadflow/dc/DcAcEmulationOuterLoop.java
Hadrien-Godard: Not really. But I think large system oscillations won't be solved with just fixing the flow on one HVDC line. Also, there is the same security with PV/PQ buses where switches are locked after some oscillations.

vidaldid-rte (changes requested):
Hi @Hadrien-Godard . I made a first macro review on the outerloop with focus on convergence criteria  I have some concerns on potential incorrect convergence. Can you see if there is a better way to distinguish small oscillations around discontinuities from real convergence issues ?

Hadrien-Godard:
> Hi @Hadrien-Godard . I made a first macro review on the outerloop with focus on convergence criteria I have some concerns on potential incorrect convergence. Can you see if there is a better way to distinguish small oscillations around discontinuities from real convergence issues ?

Hi @vidaldid-rte . Many thanks for the review.

I do not really see how easily distinguish small oscillations around discontinuities from real convergence issues. All I have to say is that real convergence issues have very small chances to be solved by fixing the flow on one line. Not real stronger arguments.

Also, the pattern chosen here is similar to what is done for PV/PQ switching.

vidaldid-rte:
> I do not really see how easily distinguish small oscillations around discontinuities from real convergence issues. All I have to say is that real convergence issues have very small chances to be solved by fixing the flow on one line. Not real stronger arguments.
> 
> Also, the pattern chosen here is similar to what is done for PV/PQ switching.

Technically it is possible to store the value at previous loop iteration and compare the distance.

Hadrien-Godard:
> > I do not really see how easily distinguish small oscillations around discontinuities from real convergence issues. All I have to say is that real convergence issues have very small chances to be solved by fixing the flow on one line. Not real stronger arguments.
> > Also, the pattern chosen here is similar to what is done for PV/PQ switching.
> 
> Technically it is possible to store the value at previous loop iteration and compare the distance.
> 
> [EDIT] @Hadrien-Godard I also think that the comparison with PV/PQ switching is incorrect. When Generators are blocked in PQ, they may not honnor a goal to control voltage. But the solution is physically correct. The values of angles, V, P, Q follow the physics.l
> 
> WIth the current implementation of this outerloop you could have a result where an HVDC link is in state null, but the angles of the final solution are so that the HVDC link would flow power in reality. OpenLoadFlow should NOT return a convergence status if this situation happened. And the current implementations provides no such warranty.

Hello @vidaldid-rte , thank you for your reply.
Yes you are right, there is a question about results interpretation. Notably, one could end up with a flow at 0 on a HVDC link and with a non-null angle difference at its ends.

But with ReactiveLimitsOuterloop, one could also end up with a generator controlling voltage and producing Q=Qmax with a voltage above its target (because blocked at PQ status). It is non-compliant with the model of generator voltage control, since reactive power output should value Qmax only if the generator cannot reach the voltage target.

Furthermore, both solutions are physically correct in terms of Ohm's and Kirchhoff's laws. They are incorrect in terms of control device model.

From my point of view, we can keep the design of this PR if we are ok with this method to fix impossible convergence issues, as we do for reactive limits.
Or we can drop the side shifting count, and consider that there is no maximal side shifting. In this case, impossible convergence will lead to "max outerloop" exit criteria. It is still better than current main branch "NR divergence" exit criteria.

The question also holds for an Hvdc link blocked to Pmax.

Maybe @geofjamg has an opinion on this.

src/main/java/com/powsybl/openloadflow/DefaultAcOuterLoopConfig.java
Hadrien-Godard: Done

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcEmulationOuterLoop.java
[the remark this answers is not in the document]
Hadrien-Godard: Done

PR #1051, opened 2024-06-20
Fix state monitor duplicated results with tie lines

geofjamg:

No

Bug fix

When configuring a dangling line ID in the state monitor, we get an exception because of duplicated network results. 

We do not have duplicated results anymore.

- [ ] Yes
- [ ] No

**Other information**:


annetill (approved):
Thanks!

PR #1055, opened 2024-07-01
Area Interchange Control as OuterLoop

### LINKED ISSUE #978: Area Interchange Control as OuterLoop

No support for Area Interchange Control

Ability to include Area Interchange Control (AIC) in OLF simulations.
The classical methods of doing AIC in a load flow are:
- Though an OuterLoop
- Directly in the Newton-Raphson

We propose as a first step to implement the OuterLoop approach only.
The feature can be enabled/disabled through OpenLoadFlowParameters.

### Describe the motivation

Area Interchange Control is a classic feature of load flows - just like simulating ratio/phase tap changers. A quick google search reveals e.g. AIC is supported by:
- [DIgSILENT PowerFactory](https://www.digsilent.de/en/load-flow-analysis.html)
- [PSS/E](https://www.linkedin.com/pulse/power-flow-analysis-solution-techniques-psse-tahir-abbas/)
- [Neplan](https://www.neplan.ch/description/load-flow/)
- [powerworld](https://www.powerworld.com/WebHelp/Content/MainDocumentation_HTML/Area_Control.htm)

Entso-e documents consider also AIC as a load flow option:
- [EMF requirements](https://eepublicdownloads.entsoe.eu/clean-documents/Network%20codes%20documents/Implementation/cacm/cgmm/European_Merging_Function_Requirements_Specification.pdf) section 4.8
- [ENTSO-E SKOS codelist](https://www.entsoe.eu/Documents/EDI/Library/ENTSO-E%20SKOS%20codelist.zip) in [EDI Library](https://www.entsoe.eu/publications/electronic-data-interchange-edi-library/) also [here](https://energy.referencedata.eu/PowerFlowSettings/e4068581-713b-4d37-88e7-bca1fd78a8f7.jsonld): `cim:PowerFlowSettings.interchangeControlEnabled": "true"`

In the PowSyBl ecosystem this can be achieved today using powsybl-entsoe/balance-adjustment module, which is a quite complicated and slow implementation. Integrating the feature directly in OLF as an OuterLoop allows for faster performance, and enables also simulating AIC in security analysis - while powsybl-entsoe/balance-adjustment approach is not suitable when simulating thousands of contingencies.

### Extra Information

Requires https://github.com/powsybl/powsybl-core/issues/2883 to be implemented first in powsybl-core/iIDM

m-guibert:

Fixes  #978 

- [x] No


src/main/java/com/powsybl/openloadflow/network/impl/LfControlAreaImpl.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import java.util.function.Supplier;
+
+public class LfControlAreaImpl extends AbstractPropertyBag implements LfControlArea {
+
+    private final LfNetwork network;
+    private final Ref<Area> areaRef;
+
+    private final double targetAcInterchange;
```
m-guibert, line 16: Rename it to interchangeTarget

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -503,6 +513,40 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
...
+        LfControlArea lfControlArea = LfControlAreaImpl.create(area, network, parameters);
+        network.addControlArea(lfControlArea);
+        area.getAreaBoundaryStream().forEach(areaBoundary -> {
+            if (areaBoundary.getTerminal().isPresent()) {
+                Terminal terminal = areaBoundary.getTerminal().get();
+                loadingContext.controlAreaBoundaries.put(terminal, lfControlArea);
+            }
+            if (areaBoundary.getBoundary().isPresent()) {
+                DanglingLine danglingLine = areaBoundary.getBoundary().get().getDanglingLine();
+                loadingContext.controlAreaBoundaries.put(danglingLine.getTerminal(), lfControlArea);
```
m-guibert, line 537: Can we really "just" consider the Boundary terminal? @vmouradian : please check if this is the right way to do it

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -503,6 +513,40 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
...
+            }
+        });
+        return lfControlArea;
+    }
+
+    private static void updateControlAreaBoundaryP(Terminal terminal, LoadingContext loadingContext, Supplier<Evaluable> getP) {
+        LfControlArea controlArea = loadingContext.controlAreaBoundaries.get(terminal);
```
m-guibert, line 544: Check if the map contains the key first

src/test/java/com/powsybl/openloadflow/ac/AICTests.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.powsybl.iidm.network.Network;
+import com.powsybl.loadflow.LoadFlow;
+import com.powsybl.loadflow.LoadFlowParameters;
+import com.powsybl.math.matrix.DenseMatrixFactory;
+import com.powsybl.openloadflow.OpenLoadFlowProvider;
+import com.powsybl.openloadflow.network.BoundaryFactory;
+import org.junit.jupiter.api.Test;
+
+class AICTests {
```
m-guibert, line 11: Rename to AreaInterchangeControlTest

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -1639,6 +1677,7 @@ static LfNetworkParameters getNetworkParameters(LoadFlowParameters parameters, O
                 .setUseLoadModel(parametersExt.isUseLoadModel())
                 .setSimulateAutomationSystems(parametersExt.isSimulateAutomationSystems())
                 .setReferenceBusSelector(ReferenceBusSelector.fromMode(parametersExt.getReferenceBusSelectionMode()))
+                .setAreaInterchangeControlAreaType(parametersExt.getAreaInterchangeControlAreaType())
```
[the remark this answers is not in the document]
jeandemanged: @vmouradian 
cc @m-guibert 

current practice is to put new parameters at the bottom

src/main/java/com/powsybl/openloadflow/network/LfControlArea.java
jeandemanged, line 1: rename to `LfArea`

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -92,6 +92,9 @@ public enum SlackDistributionFailureBehavior {
 
     public static final double MAX_SUSCEPTANCE_MISMATCH_DEFAULT_VALUE = 1e-4;
 
+    public static final boolean AREA_INTERCHANGE_CONTROL_DEFAULT_VALUE = false;
+
+    public static final String AREA_INTERCHANGE_CONTROL_AREA_TYPE_DEFAULT_VALUE = "ControlArea";
```
jeandemanged, line 97: this default should be defined in `LfNetworkParameters`

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -428,6 +433,8 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
             LfBus lfBus2 = getLfBus(branch.getTerminal2(), lfNetwork, parameters.isBreakers());
             LfBranchImpl lfBranch = LfBranchImpl.create(branch, lfNetwork, lfBus1, lfBus2, topoConfig, parameters);
             addBranch(lfNetwork, lfBranch, report);
+            updateControlAreaBoundaryP(branch.getTerminal1(), loadingContext, lfBranch::getP1);
+            updateControlAreaBoundaryP(branch.getTerminal2(), loadingContext, lfBranch::getP2);
```
jeandemanged, line 437: `update` is not the right term, `add` instead ?

src/main/java/com/powsybl/openloadflow/network/LfControlArea.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.Set;
+import java.util.function.Supplier;
+
+public interface LfControlArea extends PropertyBag {
+    String getId();
+
+    double getTargetAcInterchange();
+
+    double getAcInterchange();
```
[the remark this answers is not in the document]
vmouradian: changed in 63eb39786f9f0e1cfa606e8f9471591b54a34da5

src/main/java/com/powsybl/openloadflow/network/impl/LfControlAreaImpl.java
vmouradian: changed in 63eb39786f9f0e1cfa606e8f9471591b54a34da5

src/main/java/com/powsybl/openloadflow/network/LfControlArea.java
vmouradian: renamed in 743827bc994708c06f31f3270927c85d456b74f2

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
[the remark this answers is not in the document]
vmouradian: changed in 743827bc994708c06f31f3270927c85d456b74f2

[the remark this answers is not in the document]
m-guibert: Good to know, thanks!

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -1209,6 +1218,29 @@ public OpenLoadFlowParameters setFictitiousGeneratorVoltageControlCheckMode(Fict
...
+    public String getAreaInterchangeControlAreaType() {
+        return areaInterchangeControlAreaType;
+    }
+
+    public OpenLoadFlowParameters setAreaInterchangeControlAreaType(String areaInterchangeControlAreaType) {
+        this.areaInterchangeControlAreaType = Objects.requireNonNull(areaInterchangeControlAreaType);
+        return this;
+    }
+
+    public OpenLoadFlowParameters setVoltagePerReactivePowerControl(boolean voltagePerReactivePowerControl) {
```
m-guibert, line 1239: @vmouradian it looks like you moved this function by error

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -55,6 +56,8 @@ private static class LoadingContext {
         private final Set<ShuntCompensator> shuntSet = new LinkedHashSet<>();
 
         private final Set<HvdcLine> hvdcLineSet = new LinkedHashSet<>();
+
+        private final Map<Terminal, LfArea> areaBoundaries = new HashMap<>();
```
m-guibert, line 60: Naming is still confusing beacause not all those terminals are boundaries of the area (because of dangling lines). 
Name it areaTerminalMap instead?

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -428,6 +433,7 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
             LfBus lfBus2 = getLfBus(branch.getTerminal2(), lfNetwork, parameters.isBreakers());
             LfBranchImpl lfBranch = LfBranchImpl.create(branch, lfNetwork, lfBus1, lfBus2, topoConfig, parameters);
             addBranch(lfNetwork, lfBranch, report);
+            addLineAreaBoundaries(branch, lfBranch, loadingContext);
```
m-guibert, line 436: Rename method to addBranchAreaBoundaries

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -503,6 +511,65 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
         }
     }
 
+    private static void updateControlArea(Bus bus, LfBus lfBus, LfNetwork network, LfNetworkParameters parameters, LoadingContext loadingContext, LfNetworkLoadingReport report) {
+        Optional<Area> areaOpt = bus.getVoltageLevel().getArea(parameters.getAreaInterchangeControlAreaType());
```
m-guibert, line 515: You should explain here that only areas necessary to the AIC are created in the LfNetwork

src/test/java/com/powsybl/openloadflow/ac/AICTests.java
m-guibert: Almost: you wrote TestS instead of Test

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
m-guibert: Discussed together: the map name was confusing but you are not adding boundaries in this map, just terminals which enable you to detect and add the boundaries later

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
vmouradian: :sweat_smile: oups. Changed in 5a29ee99ce070c764b03988fb43c2636eb909798

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -299,6 +301,11 @@ public void addBus(LfBus bus) {
         bus.getLoads().forEach(load -> load.getOriginalIds().forEach(id -> loadsById.put(id, load)));
     }
 
+    public void addControlArea(LfArea controlArea) {
```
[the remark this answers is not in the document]
vmouradian: Forgot to push it previously, my bad. Done in 5a29ee99ce070c764b03988fb43c2636eb909798

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -366,6 +373,11 @@ public LfLoad getLoadById(String id) {
         return loadsById.get(id);
     }
 
+    public LfArea getControlAreaById(String id) {
```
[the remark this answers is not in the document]
vmouradian: same, updated in 5a29ee99ce070c764b03988fb43c2636eb909798

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: Yes I agree, I renamed it in 5a29ee99ce070c764b03988fb43c2636eb909798

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: changed in 5a29ee99ce070c764b03988fb43c2636eb909798

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: changed in 5a29ee99ce070c764b03988fb43c2636eb909798

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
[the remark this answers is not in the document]
m-guibert: > changed in [743827b](https://github.com/powsybl/powsybl-open-loadflow/commit/743827bc994708c06f31f3270927c85d456b74f2)

I see you moved the line to follow Damien's instructions but you didn't add the line I asked for.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
m-guibert: I think there is still a problem: you didn't move it back to its initial place. See the diff view to check if you rolled-back correctly all those modifications
image

src/main/java/com/powsybl/openloadflow/lf/outerloop/DistributedActivePowerContextData.java
jeandemanged, line 1: I would keep this one unchanged, and make another one extending it specifically for AIC, containing also the details of distributed power per area

src/main/java/com/powsybl/openloadflow/network/LfArea.java
```diff
@@ -0,0 +1,46 @@
...
+
+    double getInterchangeTarget();
+
+    double getInterchange();
+
+    Set<LfBus> getBuses();
+
+    void addBus(LfBus bus);
+
+    void addBoundaryP(Supplier<Evaluable> p);
```
jeandemanged, line 30: addBus and addBoundaryP should not be part of LfArea interface, this is only internal to LfNetworkLoaderImpl

src/main/java/com/powsybl/openloadflow/network/LfArea.java
```diff
@@ -0,0 +1,46 @@
...
+
+    void addExternalBusSlackParticipationFactor(LfBus bus, double shareFactor);
+
+    /**
+     * Some buses are not in the Area, but their slack should be considered for the slack of the Area
+     * (This happens for example when a bus is connected to multiple Areas but the flow through the bus is not considered for those areas' interchange power flow).
+     * The slack of this buses can need to be shared with other Areas with a factor for each area.
+     * @return A map of the external buses considered for the slack of the Area with the share factor of the Area
+     */
+    Map<LfBus, Double> getExternalBusesSlackParticipationFactors();
```
jeandemanged, line 42: should not be part of LfArea interface, this is internal to AIC outerloop

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,160 @@
...
+import java.text.MessageFormat;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public class AreaInterchangeControlOuterloop implements AcOuterLoop {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DistributedSlackOuterLoop.class);
```
jeandemanged, line 33:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(AreaInterchangeControlOuterloop.class);
```

src/main/java/com/powsybl/openloadflow/network/impl/LfAreaImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private final double interchangeTarget;
+
+    private final Set<LfBus> buses;
+
+    private Set<Supplier<Evaluable>> boundariesP;
+
+    private Map<LfBus, Double> externalBusesSlackParticipationFactors = new HashMap<>();
+
+    public LfAreaImpl(Area area, LfNetwork network, LfNetworkParameters parameters) {
```
jeandemanged, line 38: protected

src/main/java/com/powsybl/openloadflow/network/impl/LfAreaImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public void addBus(LfBus bus) {
+        buses.add(bus);
+    }
+
+    @Override
+    public void addBoundaryP(Supplier<Evaluable> getP) {
+        boundariesP.add(getP);
+    }
```
jeandemanged, line 82: buses and boundaries should be part of constructor/`create`

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
jeandemanged, line 1: to be revised so that LfArea-s are created once with all buses and boundaries without being mutated further, there should not be any addBus or addBoundary method in LfArea or LfAreaImpl

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -899,6 +974,52 @@ private LfNetwork create(int numCC, int numSC, Network network, List<Bus> buses,
...
+     * However, it is possible that the flow through this bus is not considered for the interchange flow of some of the areas it is connected to.
+     * In this case, the slack injection of this bus will be shared between those areas to ensure a correct balancing
+     */
+    // TODO add reports
+    private static void checkBusesWithNoArea(LfNetwork lfNetwork, LfNetworkParameters parameters, LoadingContext loadingContext) {
+        if (parameters.isAreaInterchangeControl()) {
+            List<LfBus> handledBuses = lfNetwork.getAreas().stream().flatMap(area ->
+                    area.getBuses().stream())
+                    .toList();
+            List<LfBus> unhandledBuses = lfNetwork.getBuses().stream().filter(b -> !handledBuses.contains(b)).toList();
```
jeandemanged, line 990: performance issue here, especially the `.filter(b -> !handledBuses.contains(b))` because handledBuses is a List.

changing to a Set could be an option, but we can see here that we need in LfBus a method like

`Optional<LfArea> getLfArea()`

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -899,6 +974,52 @@ private LfNetwork create(int numCC, int numSC, Network network, List<Bus> buses,
...
+                );
+
+                // If the flow through this bus is considered for the interchange of some of the connected areas,
+                // then it has already been added to the Area's externalBusesSlackParticipationFactors map with a factor 0.0
+                // Here we want to share the slack injection of this bus between the areas that are not considering the flow through this bus for their interchange power flow
+                Set<LfArea> areasSharingSlack = connectedAreas.stream().filter(area -> !area.getExternalBusesSlackParticipationFactors().containsKey(unhandledBus)).collect(Collectors.toSet());
+                if (!areasSharingSlack.isEmpty()) {
+                    areasSharingSlack.forEach(area -> area.addExternalBusSlackParticipationFactor(unhandledBus, 1.0 / areasSharingSlack.size()));
+                    LOGGER.info("Bus {} is not in any Area and is connected to Areas: {} but the flow through this bus is not considered for their interchange flow. For a correct balancing, its slack injection will be shared between those areas", unhandledBus.getId(), connectedAreas.stream().map(LfArea::getId).toList());
+                }
```
jeandemanged, line 1014: this, and probably the entire method, should be moved outside LfNetworkLoader to be AIC outerloop code (e.g. in `init` part)

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -1049,6 +1170,19 @@ private void createAutomationSystems(Network network, LfNetwork lfNetwork) {
...
+                .flatMap(lfNetwork -> lfNetwork.getAreas().stream())
+                .collect(Collectors.groupingBy(LfArea::getId, Collectors.counting()))
+                .entrySet().stream()
+                .filter(e -> e.getValue() > 1)
+                .map(Map.Entry::getKey)
+                .toList();
+        if (!duplicatedAreas.isEmpty()) {
+            throw new PowsyblException("Areas with ids " + duplicatedAreas + " are present in more than one LfNetwork. Load flow computation with area interchange control is not supported in this case.");
+        }
+    }
```
jeandemanged, line 1184: with this, in practice, you are going to throw for all real case networks which always have small portions here and there.

The correct check is verifying that each area has all its boundaries are in the same LfNetwork.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,160 @@
...
+        boolean movedBuses = false;
+        Map<LfArea, Double> remainingMismatchMap = new HashMap<>();
+        for (LfArea area : areasToBalance) {
+            double areaActivePowerMismatch = getInterchangeMismatchWithSlack(area, slackBusActivePowerMismatch);
+            ActivePowerDistribution.Result result = activePowerDistribution.run(area.getBuses(), areaActivePowerMismatch);
+            double remainingMismatch = result.remainingMismatch();
+            double distributedActivePower = areaActivePowerMismatch - remainingMismatch;
+            totalDistributedActivePower += distributedActivePower;
+            movedBuses |= result.movedBuses();
+            totalIterations += result.iteration();
```
jeandemanged, line 99: I don' think this sum has anything meaningful, only the value per area is meaningful.

src/main/java/com/powsybl/openloadflow/network/LfArea.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Set;
+import java.util.function.Supplier;
+
+/**
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public interface LfArea extends PropertyBag {
+    String getId();
+
+    double getInterchangeTarget();
```
jeandemanged, line 22: please prepare also a setter

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -618,6 +630,24 @@ public boolean isVoltagePerReactivePowerControl() {
         return voltagePerReactivePowerControl;
     }
 
+    public boolean isAreaInterchangeControl() {
```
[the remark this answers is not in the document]
vmouradian: Mutliple commits changed this, last one is 42bfe28f4674712745e95aa4539c53d9bd335aee and this is fixed now

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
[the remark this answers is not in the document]
vmouradian: adressed in 1b271346, and cleaned in a7c6a581 (Damien <3)

src/test/java/com/powsybl/openloadflow/ac/AICTests.java
vmouradian: 😅 fixed in e83f8af3

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: changed this, mostly in 097d801814619091d7d66cafc06b7cca1e399e97 and f3930a88979fe7d13f1eba90fcec07554932f868

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
vmouradian: Do not remember the last commit on this but the diff seems ok now

src/main/java/com/powsybl/openloadflow/lf/outerloop/DistributedActivePowerContextData.java
vmouradian: @jeandemanged I agree, and this makes me think that we should change AcLoadFlowResult to have distributed active power per area. Is that also what you were thinking about ?

src/main/java/com/powsybl/openloadflow/AbstractAcOuterLoopConfig.java
```diff
@@ -19,13 +21,27 @@
  */
 abstract class AbstractAcOuterLoopConfig implements AcOuterLoopConfig {
```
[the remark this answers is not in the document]
vmouradian: fixed in 7294008c23a0cc0dee0452a9408f75c6f29bcf01  (unit test updated in 9a9478c80af6e5e216bc90e464be48e7287c92c6)

src/main/java/com/powsybl/openloadflow/network/LfArea.java
vmouradian: refactored in 9a9478c80af6e5e216bc90e464be48e7287c92c6

src/main/java/com/powsybl/openloadflow/network/LfArea.java
vmouradian: refactored in 9a9478c80af6e5e216bc90e464be48e7287c92c6

src/main/java/com/powsybl/openloadflow/network/impl/LfAreaImpl.java
vmouradian: changed to protected in 9a9478c80af6e5e216bc90e464be48e7287c92c6

src/main/java/com/powsybl/openloadflow/network/impl/LfAreaImpl.java
vmouradian: refactored in 9a9478c80af6e5e216bc90e464be48e7287c92c6

src/main/java/com/powsybl/openloadflow/network/impl/LfAreaImpl.java
```diff
@@ -0,0 +1,109 @@
...
+    public void addExternalBusSlackParticipationFactor(LfBus bus, double shareFactor) {
+        if (bus != null) {
+            externalBusesSlackParticipationFactors.put(bus, shareFactor);
+        }
+    }
+
+    @Override
+    public Map<LfBus, Double> getExternalBusesSlackParticipationFactors() {
+        return externalBusesSlackParticipationFactors;
+    }
```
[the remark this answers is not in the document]
vmouradian: refactored in 9a9478c80af6e5e216bc90e464be48e7287c92c6

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: adressed in 9a9478c80af6e5e216bc90e464be48e7287c92c6

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: refactored in 9a9478c80af6e5e216bc90e464be48e7287c92c6

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
vmouradian: Refactored so that in the logging and reports we see the value per area. 
Now the totalDistributedActivePower remains used only for the  "Ac loadflow complete" log which has "distributedActivePower = ..."
I started to think about how to update this log. Do you have any ideas ?

src/main/java/com/powsybl/openloadflow/network/LfArea.java
vmouradian: added in c2f88b5fdd845f63af4e002a2722d160cbcf8a3b

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: main refactor was made in 9a9478c80af6e5e216bc90e464be48e7287c92c6
Post processors added in 365e12a7aae1b5ded8f0c5de08389f95ed490103

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: I updated it in 164e8993a75d730a8c72dbfe3c35d2227e1bb7e6 to check the boundaries.
For the moment it still throws.

Are you sure that we want just a warn ? 
We convert only the AIC areas, which means that with boundaries of a same area beeing in diffrent islands, it s pretty guaranteed that we will have weird results.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -503,6 +514,121 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
...
+            if (boundary.getTerminal().isPresent()) {
+                Terminal terminal = boundary.getTerminal().get();
+                bus = terminal.getBusBreakerView().getConnectableBus();
+            } else if (boundary.getBoundary().isPresent()) {
+                DanglingLine danglingLine = boundary.getBoundary().get().getDanglingLine();
+                bus = danglingLine.getTerminal().getBusBreakerView().getConnectableBus();
+            } else {
+                return;
+            }
+            if (bus.getConnectedComponent().getNum() != numCC || bus.getSynchronousComponent().getNum() != numSC) {
```
jeandemanged, line 591: caution, `bus.getConnectedComponent()` / `bus.getSynchronousComponent()` can be null, would thrown NPE here

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
jeandemanged: I need to think about it ... could be either automatically disbabling AIC, or returning a status failed for the involved areas.

jeandemanged: Best way to do I think is to create a new enum constant in `com.powsybl.openloadflow.network.LfNetwork.Validity` e.g. 

`INVALID_FRAGMENTED_AREA("Network contains area(s) spanning multiple components")`

jeandemanged: Actually the best would be to not perform interchange control, but only for the fragmented areas, i.e. area that are not fragmented still perform AIC

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,219 @@
...
+                        .filter(Objects::nonNull)
+                        .map(LfBus::getArea)
+                        .filter(Optional::isPresent)
+                        .map(Optional::get)
+                        .collect(Collectors.toSet());
+
+                Set<LfArea> areasSharingSlack = connectedAreas.stream()
+                                .filter(area -> area.getBoundaries().stream().noneMatch(boundary -> connectedBranches.contains(boundary.getBranch())))
+                                .collect(Collectors.toSet());
+                if (!areasSharingSlack.isEmpty()) {
```
m-guibert, line 232: Add comments to explain what it tested/done here :
// if the slack bus is on a boundary point considered for net position, 
// it will resolve naturally because deviations caused by slack are already present on tie line flows
// no need to include in any area net position calculation

src/main/java/com/powsybl/openloadflow/network/AbstractLfNetworkLoaderPostProcessor.java
```diff
@@ -31,4 +31,9 @@ public void onBranchAdded(Object element, LfBranch lfBranch) {
     public void onInjectionAdded(Object element, LfBus lfBus) {
         // to implement
     }
+
+    @Override
+    public void onAreaAdded(Object element, LfArea lfArea) {
+        // to implement
```
m-guibert, line 37: Forgotten TODO

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -503,6 +535,129 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
...
+
+    /**
+     * If we perform Area Interchange Control, we need all buses to have an Area or to be boundary buses
+     * A boundary bus is connected to one or multiple buses that are all in areas different from each other
+     * If a bus is not in any area, and is not a boundary bus, the area interchange control could not be performed and an exception will be thrown
+     * However, it is possible that the flow through this bus is not considered for the interchange flow of some of the areas it is connected to.
+     * In this case, the slack injection of this bus will be shared between those areas to ensure a correct balancing
+     */
+
+    private static void checkBusesWithoutArea(LfNetwork network) {
```
m-guibert, line 634: Delete this function and the one below as it is now dead code

src/main/java/com/powsybl/openloadflow/network/impl/LfTieLineBus.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public class LfTieLineBus extends AbstractLfBus {
+
+    private final Ref<TieLine> tieLineRef;
+
+    private final double nominalV;
+
+    public LfTieLineBus(LfNetwork network, TieLine tieLine, LfNetworkParameters parameters, LfNetworkLoadingReport report) {
```
m-guibert, line 26: Last parameter is not used

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -61,6 +66,7 @@ public void initialize(AcOuterLoopContext context) {
         var contextData = new DistributedSlackContextData();
         context.setData(contextData);
         LfNetwork network = context.getNetwork();
+        listBusesNoArea(network);
```
m-guibert, line 69: Rename to listBuses**W**oArea ?

src/main/java/com/powsybl/openloadflow/AbstractAcOuterLoopConfig.java
```diff
@@ -19,13 +21,27 @@
...
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractAcOuterLoopConfig.class);
+
     protected AbstractAcOuterLoopConfig() {
     }
 
     protected static Optional<AcOuterLoop> createDistributedSlackOuterLoop(LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
         if (parameters.isDistributedSlack()) {
+            if (parametersExt.isAreaInterchangeControl()) {
+                LOGGER.warn("DistributedSlack and AreaInterchangeControl outerloops are both enabled. DistributedSlack outerloop will be disabled. Slack will handled by the AreaInterchangeControl outerloop.");
```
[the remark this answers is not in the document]
vmouradian: fixed in dae81198cb9c57278031756521f81e895fe80804

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -1439,11 +1471,15 @@ public OpenLoadFlowParameters update(Map<String, String> properties) {
...
+        Optional.ofNullable(properties.get(AREA_INTERCHANGE_CONTROL_PARAM_NAME))
+                .ifPresent(prop -> this.setAreaInterchangeControl(Boolean.parseBoolean(prop)));
+        Optional.ofNullable(properties.get(AREA_INTERCHANGE_CONTROL_AREA_TYPE_PARAM_NAME))
+                .ifPresent(this::setAreaInterchangeControlAreaType);
         return this;
     }
 
     public Map<String, Object> toMap() {
-        Map<String, Object> map = new LinkedHashMap<>(66);
+        Map<String, Object> map = new LinkedHashMap<>(70);
```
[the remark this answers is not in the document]
vmouradian: Yes, the two last commits that added parameters did not update the initial capacity

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,219 @@
...
+                .map(entry -> String.format(Locale.US, "%s: %.2f MW", entry.getKey().getId(), entry.getValue() * PerUnit.SB))
+                .collect(Collectors.joining(", "));
+    }
+
+    private double getInterchangeMismatch(LfArea area) {
+        return area.getInterchangeTarget() - area.getInterchange();
+    }
+
+    private double getInterchangeMismatchWithSlack(LfArea area, double slackBusActivePowerMismatch) {
+        return area.getInterchange() - area.getInterchangeTarget() + getSlackInjection(area, slackBusActivePowerMismatch);
```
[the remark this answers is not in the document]
vmouradian: It is confusing I agree, in dae81198cb9c57278031756521f81e895fe80804 I used the same convention and added a minus sign to the sum of mismatch without slack when it is used.
(the sign is opposite because getInterchangeMismatchWithSlack is used to distribute the mismatch of areas inside them and getInterchangeMismatch to distribute it outside)

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,219 @@
...
+    private void allocateSlackDistributionParticipationFactors(LfNetwork lfNetwork) {
+        List<LfBus> slackBuses = lfNetwork.getSlackBuses();
+        int totalSlackBusCount = slackBuses.size();
+        for (LfBus slackBus : slackBuses) {
+            Optional<LfArea> areaOpt = slackBus.getArea();
+            if (areaOpt.isPresent()) {
+                areaSlackDistributionParticipationFactor.put(areaOpt.get(), areaSlackDistributionParticipationFactor.getOrDefault(areaOpt.get(), 0.0) + 1.0 / totalSlackBusCount);
+            } else {
+//                When a bus is connected to one or multiple Areas but the flow through the bus is not considered for those areas' interchange power flow,
+//                its slack injection should be considered for the slack of some Areas that it is connected to.
```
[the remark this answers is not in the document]
vmouradian: fixed in dae8119

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
vmouradian: added in dae8119

src/main/java/com/powsybl/openloadflow/network/AbstractLfNetworkLoaderPostProcessor.java
vmouradian: This comment is present in all the other methods of the class that are not implemented, I just copied without asking myself if this is rigth 😅

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -428,20 +436,43 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
...
-                    LfBranch lfBranch = LfTieLineBranch.create(tieLine, lfNetwork, lfBus1, lfBus2, parameters);
-                    addBranch(lfNetwork, lfBranch, report);
-                    postProcessors.forEach(pp -> pp.onBranchAdded(tieLine, lfBranch));
-                    visitedDanglingLinesIds.add(tieLine.getDanglingLine1().getId());
-                    visitedDanglingLinesIds.add(tieLine.getDanglingLine2().getId());
+                    DanglingLine danglingLine1 = tieLine.getDanglingLine1();
+                    DanglingLine danglingLine2 = tieLine.getDanglingLine2();
+                    LfBus lfBus1 = getLfBus(danglingLine1.getTerminal(), lfNetwork, parameters.isBreakers());
+                    LfBus lfBus2 = getLfBus(danglingLine2.getTerminal(), lfNetwork, parameters.isBreakers());
+                    if (parameters.isAreaInterchangeControl()) {
```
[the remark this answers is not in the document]
vmouradian: added in dae8119

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
vmouradian: fixed in dae8119

src/main/java/com/powsybl/openloadflow/network/impl/LfTieLineBus.java
vmouradian: fixed in dae8119

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
vmouradian: I prefer this too, renamed in dae8119

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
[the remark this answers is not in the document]
jeandemanged: indeed this one is often forgotten when new params are added, 70 is correct and is in line with OpenLoadFlowProviderTest / specificParametersTest

jeandemanged (dismissed):
my review is incomplete sorry ... will continue later on

src/main/java/com/powsybl/openloadflow/network/AbstractLfNetworkLoaderPostProcessor.java
m-guibert: My bad, I just misunderstood the comment. Thanks for the clarification!

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,248 @@
...
+        List<LfArea> areas = context.getNetwork().getAreas();
+        double slackBusActivePowerMismatch = context.getLastSolverResult().getSlackBusActivePowerMismatch();
+
+        Map<LfArea, Double> areaInterchangeWithSlackMismatches = areas.stream()
+                .collect(Collectors.toMap(area -> area, area -> getInterchangeMismatchWithSlack(area, slackBusActivePowerMismatch)));
+        List<LfArea> areasToBalance = areaInterchangeWithSlackMismatches.entrySet().stream()
+                .filter(entry -> {
+                    double areaActivePowerMismatch = entry.getValue();
+                    double absMismatch = Math.abs(areaActivePowerMismatch);
+                    return absMismatch > this.areaInterchangePMaxMismatch / areas.size() / PerUnit.SB && absMismatch > ActivePowerDistribution.P_RESIDUE_EPS;
```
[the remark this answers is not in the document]
vmouradian: In large networks where every bus is in an area, if all areas have their mismatches within `areaInterchangePMaxMismatch`, mismatches can add up and generate slack higher than `areaInterchangePMaxMismatch` (= mismatch higher than this threshold for the area with slack bus at the end of the day).

It was a temporary fix that lasted but I will try to find a way to do this better

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowReportTest.java
```diff
@@ -219,4 +219,22 @@ void shuntVoltageControlDiscarded() throws IOException {
         LoadFlowAssert.assertReportEquals("/shuntVoltageControlDiscarded.txt", reportNode);
     }
 
+    @Test
+    void areaVoltageControlOuterloop() throws IOException {
```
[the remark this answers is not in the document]
vmouradian: Wow 😅😅😅 
fixed in 78188368bd470c602002e0284023d932d6625f9c

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingencyCreationParameters.java
```diff
@@ -24,6 +25,8 @@ public class PropagatedContingencyCreationParameters {
 
     private boolean hvdcAcEmulation = LoadFlowParameters.DEFAULT_HVDC_AC_EMULATION_ON;
 
+    private boolean areaInterchangeControlOn = OpenLoadFlowParameters.AREA_INTERCHANGE_CONTROL_DEFAULT_VALUE;
```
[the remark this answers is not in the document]
vmouradian: fixed in 78188368bd470c602002e0284023d932d6625f9c

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -503,6 +537,94 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
...
+            if (boundary.getTerminal().isPresent()) {
+                Terminal terminal = boundary.getTerminal().get();
+                bus = terminal.getBusBreakerView().getConnectableBus();
+            } else if (boundary.getBoundary().isPresent()) {
+                DanglingLine danglingLine = boundary.getBoundary().get().getDanglingLine();
+                bus = danglingLine.getTerminal().getBusBreakerView().getConnectableBus();
+            } else {
+                continue;
+            }
+            if (bus.getConnectedComponent() != null && bus.getSynchronousComponent() != null && (bus.getConnectedComponent().getNum() != numCC || bus.getSynchronousComponent().getNum() != numSC)) {
```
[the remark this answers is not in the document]
vmouradian: fixed in 78188368bd470c602002e0284023d932d6625f9c

src/main/java/com/powsybl/openloadflow/network/impl/LfAreaImpl.java
```diff
@@ -0,0 +1,111 @@
...
+    private double interchangeTarget;
+
+    private final Set<LfBus> buses;
+
+    private final Set<Boundary> boundaries;
+
+    protected LfAreaImpl(Area area, Set<LfBus> buses, Set<Boundary> boundaries, LfNetwork network, LfNetworkParameters parameters) {
+        this.network = network;
+        this.areaRef = Ref.create(area, parameters.isCacheEnabled());
+        this.interchangeTarget = area.getInterchangeTarget().orElseThrow(() -> new PowsyblException("Area " + area.getId() + " does not have a net position target.")) / PerUnit.SB;
```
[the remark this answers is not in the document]
vmouradian: changed in 78188368bd470c602002e0284023d932d6625f9c

Now if the interchange target is Empty it is defaulted to 0.0 in LfArea. But in the network loader the area is simply not created, same as for fragmented areas

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,248 @@
...
+                })
+                .map(Map.Entry::getKey)
+                .toList();
+        boolean shouldBalance = !areasToBalance.isEmpty();
```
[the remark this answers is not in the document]
vmouradian: fixed in 78188368bd470c602002e0284023d932d6625f9c

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    private OuterLoopResult distributeActivePowerMismatches(Map<String, Pair<Set<LfBus>, Double>> areas, AcOuterLoopContext context, ReportNode reportNode) {
+        double totalDistributedActivePower = 0.0;
+        int totalIterations = 0;
+        boolean movedBuses = false;
+        Map<String, Double> remainingMismatchMap = new HashMap<>();
+
+        for (Map.Entry<String, Pair<Set<LfBus>, Double>> e : areas.entrySet()) {
+            Double areaActivePowerMismatch = e.getValue().getRight();
```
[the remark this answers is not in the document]
vmouradian: fixed in 78188368bd470c602002e0284023d932d6625f9c

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
```diff
@@ -0,0 +1,248 @@
...
+
+    private final ActivePowerDistribution activePowerDistribution;
+
+    /**
+     * The part of the total slack active power mismatch that should be added to the Area's net interchange mismatch, ie the part of the slack that should be distributed in the Area.
+     */
+
+    private Map<String, Double> areaSlackDistributionParticipationFactor;
+
+    private Set<LfBus> listBusesWoArea = new HashSet<>();
```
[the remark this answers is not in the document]
vmouradian: fixed in 78188368bd470c602002e0284023d932d6625f9c
I created an extension of DistributedSlackContextData as you recommended

vmouradian:
> https://github.com/powsybl/powsybl-open-loadflow/pull/1055#discussion_r1752188627 
you renamed the attribute but not the method, but see earlier comment about context data.  
I'm not a big fan of wo, not everyone understands it, and also aggravated because the usual abbreviation is w/o but the slash makes it impossible to use in code. The method and attributes are short so I suggest just Without

fixed in 78188368bd470c602002e0284023d932d6625f9c

vmouradian:
> https://github.com/powsybl/powsybl-open-loadflow/pull/1055#discussion_r1752207287
actually the comment was about totalIterations, not totalDistributedActivePower for which the sum is meaningful

I splitted the report of iterations among areas as for the power in 78188368bd470c602002e0284023d932d6625f9c

src/main/java/com/powsybl/openloadflow/ac/outerloop/AreaInterchangeControlOuterloop.java
vmouradian: fixed in 78188368bd470c602002e0284023d932d6625f9c

docs/loadflow/loadflow.md
```diff
@@ -149,3 +149,66 @@ Note that the vector $b$ of right-hand sides is linearly computed from the given
...
+$$
+
+If this mismatch for all areas and the slack injection of the buses without area are below `slackBusPMaxMismatch` the outerloop is stable and finishes, meaning that the interchanges are correct and the slack bus active power is distributed.
+
+If not, the remaining mismatch is first distributed over the buses that have no area.
+
+If some mismatch still remains, it is distributed equally over all the areas.
+
+### Areas validation
+There are some cases where areas are considered invalid and will not be considered for the area interchange control :
```
m-guibert, line 192: No space before ":" in English. Check below tto, it is done several times

docs/loadflow/loadflow.md
```diff
@@ -149,3 +149,66 @@ Note that the vector $b$ of right-hand sides is linearly computed from the given
...
+
+## Area Interchange Control
+
+The area interchange control can be activated in the parameters and is performed by an outer loop.
+
+The outer loop performs an active power distribution in each one of the selected areas (for areas selection see `areaInterchangeControlAreaType`),
+in order to have all areas' active power interchanges matching their target interchanges.
+It can handle networks where part (or even all) of the buses are not in an area. For networks that have no areas at all, the behaviour will be the same as with the distributed slack outer loop.
+
+It is preformed in an iterative way. After each iteration, a load flow computation is run and the next iteration will apply its changes based on this new calculated network state.
```
vidaldid-rte, line 161: performed ? (instead of preformed)

src/main/java/com/powsybl/openloadflow/AbstractAcOuterLoopConfig.java
```diff
@@ -19,13 +21,27 @@
...
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractAcOuterLoopConfig.class);
+
     protected AbstractAcOuterLoopConfig() {
     }
 
     protected static Optional<AcOuterLoop> createDistributedSlackOuterLoop(LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
         if (parameters.isDistributedSlack()) {
+            if (parametersExt.isAreaInterchangeControl()) {
+                LOGGER.info("DistributedSlack and AreaInterchangeControl outer loops are both enabled. DistributedSlack outerloop will not be created, slack will be distributed by the AreaInterchangeControl outerloop.");
```
vidaldid-rte, line 32: This will break if an explicit outerloop list is specified and if it contains DistributedSclackOuterLoop but not AreaInterchangeControlOuterloop.
This control should not be in this factory level.
Maybe be:
  this logic in DefaultAcOuterLoop (that creates outerloop based on parameters)
  a check in Explicit that you do not have both DistributedSclackOuterLoop and AreaInterchangeControlOuterloop at the same time because they are incompatible.

src/test/java/com/powsybl/openloadflow/ac/AreaInterchangeControlTest.java
```diff
@@ -0,0 +1,226 @@
...
+        parametersExt = OpenLoadFlowParameters.create(parameters)
+                .setAreaInterchangeControl(true)
+                .setSlackBusPMaxMismatch(1e-3);
+    }
+
+    @Test
+    void twoAreasWithXnodeTest() {
+        Network network = MultiAreaNetworkFactory.createTwoAreasWithXNode();
+        runLfTwoAreas(network, -40, 40, -30, 2);
+    }
```
vidaldid-rte, line 50: Usability comment:
   If you create an unbalanced situation (for example -40, 41 as  interchange targets), it is difficult to troubleshoot.
  The solver fails with max iteration reached. I have seen no other warning.

   I wonder if it would not help users to add some checks, at least when all boundaries belong each to at least two areas.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -503,6 +514,96 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
...
+
+    private static void createAreas(int numCC, int numSC, LoadingContext loadingContext, List<LfNetworkLoaderPostProcessor> postProcessors, LfNetworkParameters parameters, LfNetwork network) {
+        if (parameters.isAreaInterchangeControl()) {
+            loadingContext.areaBusMap
+                    .entrySet()
+                    .stream()
+                    .filter(e -> {
+                        if (e.getKey().getAreaBoundaryStream().findAny().isEmpty()) {
+                            LOGGER.warn("Area {} does not have any area boundary. The area will not be considered for area interchange control", e.getKey().getId());
+                            return false;
```
vidaldid-rte, line 566: If an area has this issue and has 1000 buses the warning will be displayed 1000 times.
Might be better to filter the eligible areas earlier.

docs/loadflow/loadflow.md
vmouradian: thanks ! fixed in e5d9d8d1e3e29980fbe2d62de80be0cf53de1340

src/main/java/com/powsybl/openloadflow/AbstractAcOuterLoopConfig.java
vmouradian: i suggest another way to fix it in 6cc333a90c271a4448bce95b72d440602624d227

vidaldid-rte: I think your approach works as well.
One minor remark: in the new test you could explicitely set the param distribtuedSlack to true. This way you won't rely on the default value to test that the loop is actually filtered in the default configuration.

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -365,6 +372,15 @@ public LfLoad getLoadById(String id) {
         return loadsById.get(id);
     }
 
+    public List<LfArea> getAreas() {
+        return new ArrayList<>(areasById.values());
```
vidaldid-rte, line 376: The copy here is expensive, not necessary,  and will require unnecessary GC work.
Can't you instead create a getAreasStream  and hasArea() to replace getAreas().isEmpty() ?

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBranch.java
```diff
@@ -7,10 +7,7 @@
  */
 package com.powsybl.openloadflow.network.impl;
 
-import com.powsybl.iidm.network.DanglingLine;
-import com.powsybl.iidm.network.LimitType;
-import com.powsybl.iidm.network.LoadingLimits;
-import com.powsybl.iidm.network.TwoSides;
+import com.powsybl.iidm.network.*;
```
vidaldid-rte, line 10: Did you remove the explicit imports on purpose ?
Or is it in anitelliJ missconfiguration ? In this case this post may help https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports

docs/loadflow/parameters.md
```diff
@@ -82,6 +82,15 @@ The `slackBusCountryFilter` defines a list of countries where slack bus should b
...
 
+**areaInterchangeControl**  
+The `areaInterchangeControl` property is an optional property that defines if the area interchange control outer loop is enabled. 
+If set to `true`, the area interchange control outer loop will be used instead of the slack distribution outer loop.
+The default value is `false`.
+
+**areaInterchangeControlAreaType**  
+Defines the `areaType` of the areas on which the area interchange control is applied.
+Only the areas of the input network that have this type will be converted and considered.
+
```
vidaldid-rte, line 93: Can you document here the default value ?

src/test/java/com/powsybl/openloadflow/ac/AreaInterchangeControlTest.java
vidaldid-rte: Closing this one since a new parameter exists now for area mismatch

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBranch.java
vidaldid-rte: Closing since this has been fixed.

docs/loadflow/parameters.md
vidaldid-rte: Resolving since this has been added.

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
vidaldid-rte: Resolved

jeandemanged (commented):
All fine for me, thank you, @vidaldid-rte could you make a final review please ?

docs/loadflow/loadflow.md
```diff
@@ -201,3 +201,81 @@ Note that the vector $b$ of right-hand sides is linearly computed from the given
...
+
+In iIDM each area defines the boundary points to be considered in the interchange. iIDM supports two ways of modeling area boundaries:
+- either via an equipment terminal,
+- or via a DanglingLine boundary.
+
+In the DanglingLine case, the flow at the boundary side is considered as it should be, for both unpaired DanglingLines and DanglingLines paired in a TieLine.
+
+### Slack bus mismatch attribution
+Depending on the location of the slack bus(es), the role of distributing the active power mismatch will be attributed based on the following logic:
+- If the slack bus part of an area: the slack power is attributed to the area (see "total mismatch" calculation in [Algorithm description](#area-interchange-control---algorithm-description)).
```
vidaldid-rte, line 275: Typo (if the slack bus **is** part...)

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -191,7 +192,7 @@ public AcLoadFlowResult run() {
         for (var outerLoopAndContext : Lists.reverse(outerLoopsAndContexts)) {
             var outerLoop = outerLoopAndContext.getLeft();
             var outerLoopContext = outerLoopAndContext.getRight();
-            if (outerLoop instanceof DistributedSlackOuterLoop) {
+            if (outerLoop instanceof DistributedSlackOuterLoop || outerLoop instanceof AreaInterchangeControlOuterloop) {
```
vidaldid-rte, line 195: We now have two distribution outerloops. 2 is a little small number to generalize but sill I wonder if we should not create an interface (for example IActivePowerDistributionOuterLoop) that would provide:
    A way to identify those loops (since apparently there can be only one) - and simplify this condition
    A method to get the distributedActivePower
    A consistent way to handle the slackDistributionFailureBehavior parameter

vidaldid-rte (commented):
Looks good. A doc typo and one suggestion about creating a common interface between the two active power distribution outerloops that have important services in common.

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
jeandemanged: > We now have two distribution outerloops. 2 is a little small number to generalize but sill I wonder if we should not create an interface (for example IActivePowerDistributionOuterLoop) that would provide: A way to identify those loops (since apparently there can be only one) - and simplify this condition A method to get the distributedActivePower A consistent way to handle the slackDistributionFailureBehavior parameter

this definitely needs to be done, considering that in the near future we would also create an area interchange control outerloop for DC load flow (in DC the problem is trivial when all areas are "closed" meaning all boundaries are accounted, but when some interconnections are not to be considered in the exchange, then an iterative method is needed...
So 2 will become 3, and I would propose to do these refactoring at the time of introduction of the DC AIC outerloop if this is ok for you.

vidaldid-rte: That works for me.

vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/network/impl/LfAreaImpl.java
```diff
@@ -0,0 +1,122 @@
...
+                    double nominalV1 = danglingLine1.getTerminal().getVoltageLevel().getNominalV();
+                    return new SV(lfTieLineBranch.getP1().eval() * PerUnit.SB, lfTieLineBranch.getQ1().eval() * PerUnit.SB, lfTieLineBranch.getV1() * nominalV1, Math.toDegrees(lfTieLineBranch.getAngle1()), side)
+                            .otherSideP(danglingLine1, true) / PerUnit.SB;
+                } else if (side == TwoSides.TWO) {
+                    DanglingLine danglingLine = lfTieLineBranch.getHalf2();
+                    double nominalV2 = danglingLine.getTerminal().getVoltageLevel().getNominalV();
+                    return new SV(lfTieLineBranch.getP2().eval() * PerUnit.SB, lfTieLineBranch.getQ2().eval() * PerUnit.SB, lfTieLineBranch.getV2() * nominalV2, Math.toDegrees(lfTieLineBranch.getAngle2()), side)
+                            .otherSideP(danglingLine, true) / PerUnit.SB;
+                }
+            }
```
jeandemanged, line 115: splitShuntAdmittance to be set to false, see also https://github.com/powsybl/powsybl-core/pull/3169 and https://github.com/powsybl/powsybl-open-loadflow/pull/1098

jeandemanged (changes requested):
one last thing to solve before merging

PR #1062, opened 2024-07-10
Computing Hvdc line loss in AC emulation

SylvestreSakti:

No

The computation of power loss due to resistance in HVDC line in AC emulation has been added.

The resistance R of an HVDC line is not taken into account, the result of this approximation is that active power transit in the HVDC line can be overestimated.

The power loss formula has been introduced in the computation of power in AC emulation and is given in the figure below. 
image

The Joule loss is minor compared to the power transit so its derivative has been neglected in the computation of the Jacobian matrix.

- [ ] Yes
- [x] No

**Other information**:


src/main/java/com/powsybl/openloadflow/network/impl/LfHvdcImpl.java
```diff
@@ -52,6 +54,7 @@ public LfHvdcImpl(String id, LfBus bus1, LfBus bus2, LfNetwork network, HvdcLine
         this.id = Objects.requireNonNull(id);
         this.bus1 = bus1;
         this.bus2 = bus2;
+        this.r = hvdcLine.getR() / PerUnit.zb(hvdcLine.getNominalV()); // Storing R directly in per unit
```
annetill, line 57: Can you remain consistent with other treatment. P0 is not "stored" in per-unit.

src/main/java/com/powsybl/openloadflow/ac/equations/AbstractHvdcAcEmulationFlowEquationTerm.java
```diff
@@ -55,6 +58,20 @@ protected double rawP(double p0, double k, double ph1, double ph2) {
         return p0 + k * (ph1 - ph2);
     }
 
+    protected double rawPWithLoss(double rawP, double loss1, double loss2, double v, double r) {
```
annetill, line 61: We already have in `powsybl-core` the method in `HvdcUtils.getHvdcLineLosses(...)`. Can we imagine reusing it here because it is what we use when the hvdc is operated in active power set point mode.

src/main/java/com/powsybl/openloadflow/network/impl/LfHvdcImpl.java
SylvestreSakti: Yes, we had this question of consistency with two possible options :
- Being consistent with the other treatments so introducing an additional property nominalV in LfHvdc
- Changing the other treatment to "store" directly p0 and droop in per unit (but it may need some other modifications and maybe another PR)

I don't know which one is the best, but the easiest seems to be the first option

src/main/java/com/powsybl/openloadflow/ac/equations/AbstractHvdcAcEmulationFlowEquationTerm.java
```diff
@@ -78,6 +81,17 @@ protected double getVscLossMultiplier() {
         return (1 - lossFactor1) * (1 - lossFactor2);
     }
 
+    // FIXME: This method is a copy of HvdcUtils.getHvdcLineLosses() from powsybl-core but it cannot be used yet because this static method is private
+    protected static double getHvdcLineLosses(double rectifierPDc, double nominalV, double r) {
```
annetill, line 85: Thanks, should be fixed soon with https://github.com/powsybl/powsybl-core/pull/3098, that should be included in powsbl-core 6.4.1.

annetill: You can upgrade the version, powsybl-core 6.4.1 available.

SylvestreSakti: Thanks, the code now uses the method getHvdcLineLosses().

src/main/java/com/powsybl/openloadflow/ac/equations/HvdcAcEmulationSide1ActiveFlowEquationTerm.java
```diff
@@ -24,9 +24,13 @@ public HvdcAcEmulationSide1ActiveFlowEquationTerm(LfHvdc hvdc, LfBus bus1, LfBus
...
 
     private double p1(double ph1, double ph2) {
-        double rawP = rawP(p0, k, ph1, ph2);
-        double boundedP = boundedP(rawP);
-        return (isController(rawP) ? 1 : getVscLossMultiplier()) * boundedP;
+        double rawP = boundedP(rawP(p0, k, ph1, ph2));
+        if (isController(rawP)) {
+            return rawP;
+        } else {
+            double rectifierPDc = (1 - lossFactor1) * rawP;
```
annetill, line 31: Can you try to put more commun code in the abstract?

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowVscTest.java
```diff
@@ -450,6 +481,7 @@ void testAcEmuWithOperationalLimits() {
         Network network = HvdcNetworkFactory.createHvdcLinkedByTwoLinesAndSwitch(HvdcConverterStation.HvdcType.VSC);
         // without limit p=195
         network.getHvdcLine("hvdc23")
+                .setR(0d) //Removing resistance to ignore cable loss
```
annetill, line 484: As a general comment, if unit tests changed because we improve/fix something, we prefer to fix them than to retry to code the old behavior.

src/main/java/com/powsybl/openloadflow/ac/equations/HvdcAcEmulationSide1ActiveFlowEquationTerm.java
SylvestreSakti: Some modifications have been done, especially by creating a method getAbsActivePowerWithLosses() that computes the absolute value of the active power taking into account loss of converters (in the correct order depending on the direction of the transit) and the line loss.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowVscTest.java
SylvestreSakti: Ok, then the values of the tests have been updated with the new behavior (that have been verified). A new test had been added for the case without R to have an independant test ignoring line loss.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
```diff
@@ -1284,6 +1284,7 @@ void testVSCLossAcEmulation() {
         // but other converter station keeps its voltage control capability.
         // remedial action re-enables the ac emulation of the hvdc line.
         Network network = HvdcNetworkFactory.createHvdcLinkedByTwoLinesAndSwitch(HvdcConverterStation.HvdcType.VSC);
+        network.getHvdcLine("hvdc23");
```
annetill, line 1287: Line to be removed?

src/main/java/com/powsybl/openloadflow/ac/equations/AbstractHvdcAcEmulationFlowEquationTerm.java
```diff
@@ -78,6 +83,15 @@ protected double getVscLossMultiplier() {
         return (1 - lossFactor1) * (1 - lossFactor2);
     }
 
+    protected double getAbsActivePowerWithLosses(double boundedP, double lossRectifier, double lossInverter) {
```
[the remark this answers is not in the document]
SylvestreSakti: Thank you for this detail, I understand now. The orders of the losses are important in this formula so the name are important. On one side there is the "controller", and which name on the other side ? What about "controlled" station ?

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisWithActionsTest.java
SylvestreSakti: Yes, it had to be removed.

src/main/java/com/powsybl/openloadflow/ac/equations/AbstractHvdcAcEmulationFlowEquationTerm.java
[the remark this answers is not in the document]
annetill: I think that in a previous version, I used controller vs non controller :-)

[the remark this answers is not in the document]
SylvestreSakti: Alright :)

src/main/java/com/powsybl/openloadflow/ac/equations/HvdcAcEmulationSide1ActiveFlowEquationTerm.java
```diff
@@ -24,9 +24,9 @@ public HvdcAcEmulationSide1ActiveFlowEquationTerm(LfHvdc hvdc, LfBus bus1, LfBus
     }
 
     private double p1(double ph1, double ph2) {
-        double rawP = rawP(p0, k, ph1, ph2);
-        double boundedP = boundedP(rawP);
-        return (isController(rawP) ? 1 : getVscLossMultiplier()) * boundedP;
+        double boundedP = boundedP(rawP(ph1, ph2));
+        // if converterStation1 is controller, then p1 is positive, otherwise it is negative
+        return isController(boundedP) ? boundedP : -getAbsActivePowerWithLosses(boundedP, lossFactor1, lossFactor2);
```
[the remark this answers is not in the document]
SylvestreSakti: Because isController() only checks the sign of the power transit. I thought there was no need to store rawP instead of directly checking the sign of boundedP.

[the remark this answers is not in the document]
annetill: I think it is equivalent but as Didier is not here (these lines about boundedP are from him) maybe keep former lines.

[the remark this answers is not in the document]
SylvestreSakti: Yes it is equivalent (unless bounds of P are negative which they shouldn't). But I modified in order to keep the former lines.

PR #1063, opened 2024-07-10
[WIP] Slack distribution on generators (optionally) only considering generators on voltage control

jeandemanged:

No

Feature

List of generators participating in slack distribution does not consider whether the generator is on voltage control.

We can restrict the list of generators participating in slack distribution to generators on voltage control.

- [x] No

**Other information**:

Motivation: [European Merging Function Requirements Specification v3](https://www.entsoe.eu/Documents/CIM_documents/Grid_Model_CIM/EMF%20requirements%20specification%20v3_Approved.pdf), page 40, option `generation distribution active power and voltage nodes only`.

 `generation distribution active power ...` suggests this option should apply only for Balance Type `PROPORTIONAL_TO_GENERATION_P`, but in this PR we allow also to restrict also for the other possible Balance Types involving generators (today `PROPORTIONAL_TO_GENERATION_P_MAX`, `PROPORTIONAL_TO_GENERATION_REMAINING_MARGIN`, `PROPORTIONAL_TO_GENERATION_PARTICIPATION_FACTOR`)


docs/loadflow/parameters.md
```diff
@@ -112,6 +112,21 @@ When slack distribution is enabled (`distributedSlack` set to `true` in LoadFlow
...
+**slackDistributionGeneratorsVoltageControlOnly**  
+When set to `true`, only generators on voltage control will participate in slack distribution.
+
+Notes:
+* Effective only when in LoadFlowParameters:
+  1. `distributedSlack` is set to `true`, and,
+  2. `balanceType` is set to a mode distributing on generators
+* If a generator voltage control is enabled in input data but internally disabled due to a consistency check,
+(see e.g. `reactiveRangeCheckMode` and `disableVoltageControlOfGeneratorsOutsideActivePowerLimits`), 
+then no slack distribution will occur for that generator.
```
jeandemanged, line 124: not sure this is right, changing PR to WIP

src/main/java/com/powsybl/openloadflow/network/util/GenerationActivePowerDistributionStep.java
```diff
@@ -134,8 +137,11 @@ private double getParticipationFactor(LfGenerator generator) {
     }
 
     private boolean isParticipating(LfGenerator generator) {
-        // check first if generator is set to be participating
-        if (!generator.isParticipating()) {
+        // check first if generator:
+        // - is set to be participating
+        // - and if option is set to distribute only on generators on voltage control
+        if (!generator.isParticipating() ||
```
annetill, line 143: At each distribution outer loop, the controlled buses are changing (PV buses) so are the associated controller buses too. If you want a dynamic behavior, you must prefer using the method `isGeneratorVoltageControlEnabled()` from the controller bus (even if in case of remote voltage control? It is not the PV bus...). If you don't want a dynamic behavior, I think you can use the type (but without a dedicated method called `hasVoltageControl' because we have too much and we have tried to make the difference between voltage control, controllers and controlled buses).

jeandemanged: I don't want dynamic behavior. Also I fear that doing this would result in "distributed slack" and "reactive limits" fighting each other.

But checking again my requirements, the `hasVoltageControl` as currently implemented is not what I need functionally speaking, i need usage of the "raw" voltageRegulatorOn generator status.

docs/loadflow/parameters.md
jeandemanged: this is not right, or say, not what I need. I need to consider only the input setting of the generator.

src/main/java/com/powsybl/openloadflow/network/util/GenerationActivePowerDistributionStep.java
annetill: So, maybe you have to add a new type because "OFF" can mean off in input data and off because checks are not passing.

jeandemanged: yes exactly, this is why I wrote https://github.com/powsybl/powsybl-open-loadflow/pull/1063#discussion_r1673546364 ...

but I am wondering if this is a feature we want in OLF or if we want to leave this feature outside ... it is quite easy to disable slack distribution on a generator via ActivePowerControl extension ...

annetill: Sorry I miss that, yes if you only need IIDM "voltageRegulatorOn" status, better try first to use active power control extension.

jeandemanged:
this requirement is best handled externally to OLF by using active power control extension - closing

PR #1064, opened 2024-07-10
Extract Woodbury engine from DC sensitivity analysis

p-arvy:

No.

The calculation associated with Woodbury equality in DC sensitivity analysis is extracted into a new class : WoodburyEngine.

The calculations are made in DC sensitivity analysis.

- [ ] Yes
- [ ] No

**Other information**:


Hadrien-Godard (approved):

annetill (approved):
Nice work. All tests passing, including our internal cucumber tests. Maybe need a check from @jeandemanged because we create a new engine.

jeandemanged:
> Nice work. All tests passing, including our internal cucumber tests. Maybe need a check from @jeandemanged because we create a new engine.

will have a look by the end of the day

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
```diff
@@ -0,0 +1,109 @@
...
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ * @author Gaël Macherel {@literal <gael.macherel@artelys.com>}
+ */
+public class WoodburyEngine {
+
+    final DcLoadFlowContext loadFlowContext;
+
+    final List<ComputedContingencyElement> contingencyElements;
+
+    final DenseMatrix contingenciesStates;
```
jeandemanged, line 32: `private` ?

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ * @author Gaël Macherel {@literal <gael.macherel@artelys.com>}
+ */
+public class WoodburyEngine {
+
+    final DcLoadFlowContext loadFlowContext;
```
jeandemanged, line 28: only usage of `DcLoadFlowContext` is in `calculatePower`, and actually only using the `DcEquationSystemCreationParameters` 

isn't actually the `DcEquationSystemCreationParameters` that you need in the WoodburyEngine ?

or are there more future developments to come where you will need really the `DcLoadFlowContext` ?

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
```diff
@@ -0,0 +1,109 @@
...
+    public WoodburyEngine(DcLoadFlowContext loadFlowContext, List<ComputedContingencyElement> contingencyElements,
+                          DenseMatrix contingenciesStates) {
+        this.loadFlowContext = Objects.requireNonNull(loadFlowContext);
+        this.contingencyElements = Objects.requireNonNull(contingencyElements);
+        this.contingenciesStates = Objects.requireNonNull(contingenciesStates);
+    }
+
+    private double calculatePower(LfBranch lfBranch) {
+        PiModel piModel = lfBranch.getPiModel();
+        DcEquationSystemCreationParameters creationParameters = loadFlowContext.getParameters().getEquationSystemCreationParameters();
```
jeandemanged, line 43: see other comment, probably should store the DcEquationSystemCreationParameters in the woodbury engine directly

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -413,16 +332,14 @@ private void calculateSensitivityValuesForAContingency(DcLoadFlowContext loadFlo
                 newFactorStates = calculateFactorStates(loadFlowContext, factorGroups, newParticipatingElements);
             }
 
-            DenseMatrix newFlowStates = calculateActivePowerFlows(loadFlowContext, factors, newParticipatingElements, disabledNetwork, reportNode);
+            DenseMatrix newFlowStates = calculateFlowStates(loadFlowContext, newParticipatingElements, disabledNetwork, reportNode);
 
-            calculateSensitivityValues(loadFlowContext, factors, newFactorStates, contingenciesStates, newFlowStates, contingencyElements,
-                    contingency, resultWriter, disabledNetwork);
+            WoodburyEngine engine = new WoodburyEngine(loadFlowContext, contingencyElements, contingenciesStates);
```
jeandemanged, line 337: If I am not mistaken this is exactly the same engine as line 284. Maybe create this engine only at one place, before the if...else because it is just the same engine.

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
Hadrien-Godard: Yes I have modified it, thanks.

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
Hadrien-Godard: Yes you are right. I have switched the whole DcLoadFlowContext to only the DcEquationSystemCreationParameters.

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
Hadrien-Godard: Yes done, thanks.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
Hadrien-Godard: Done.

jeandemanged (approved):
commit signoff to be fixed

annetill (approved):
Tests are passing: powsybl-open-rao ok and internal tests too. Note that refactoring must continue in another PR because the remaining "weird" design is that the run method of the engine takes as argument a "new flow states", slightly modified from the N states. I think this must be hidden in the engine.

PR #1075, opened 2024-08-07
AC loadflow: no network update on NO_CALCULATION

flo-dup:

No

Bug fix

In AC loadflow, network is updated on the connected component even if no calculation has been made. This leads for instance to some incorrect values on islands with no active generators (p0 and q0 set to p and q on loads).

In AC loadflow, network is only updated is the result is successful, similarly to DC loadflow.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
@@ -104,7 +104,7 @@ public class OpenLoadFlowProvider implements LoadFlowProvider {
         }
         try {
             // update network state
-            if (atLeastOneComponentHasToBeUpdated || parametersExt.isAlwaysUpdateNetwork()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -104,7 +104,8 @@ private void updateAcState(Network network, LoadFlowParameters parameters, OpenL
         }
         try {
             // update network state
-            if (atLeastOneComponentHasToBeUpdated || parametersExt.isAlwaysUpdateNetwork()) {
+            if (atLeastOneComponentHasToBeUpdated && result.isSuccess()
+                    || parametersExt.isAlwaysUpdateNetwork()) {
```
flo-dup, line 108: Just to mention that `alwaysUpdateNetwork` parameter makes the network to be updated even on `NO_CALCULATION` components, but that should be the expected behaviour.

vidaldid-rte (approved):

PR #1079, opened 2024-08-23
Multi-thread security analysis

geofjamg:

No

Feature

Security analysis runs on a single thread.

We can specify the number of thread, used by security analysis.
Performance benchmark on my laptop (13th Gen Intel(R) Core(TM) i7-13700H) for a 6000 buses network simulating 500 contingencies:
     - 1 thread: 105 s
     - 2 threads: 59 s
     - 4 threads: 33 s
     - 8 threads: 23 s

- [ ] Yes
- [x] No

**Other information**:


jeandemanged:
excellent :)

please add new `threadCount` SA parameter doc to docs/security/parameters.md

geofjamg:
> excellent :)
> 
> please add new `threadCount` SA parameter doc to docs/security/parameters.md

Done, thanks.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -133,32 +146,102 @@ SecurityAnalysisReport runSync(SecurityAnalysisParameters securityAnalysisParame
...
+            try (LfNetworkList lfNetworks = Networks.load(network, parameters.getNetworkParameters(), topoConfig, saReportNode)) {
+                // run simulation on largest network
+                result = lfNetworks.getLargest().filter(n -> n.getValidity() == LfNetwork.Validity.VALID)
+                        .map(largestNetwork -> runSimulations(largestNetwork, propagatedContingencies, parameters, securityAnalysisParameters, operatorStrategies, actions, limitReductions))
+                        .orElse(createNoResult());
+            }
+        } else {
+            if (!network.getVariantManager().isVariantMultiThreadAccessAllowed()) {
+                throw new PowsyblException("To run a security analysis on multiple threads (and so on variants), you need to set allowVariantMultiThreadAccess to true");
+            }
```
jeandemanged, line 172: Java API users will make it, but wondering how to make this more user friendly, in particular for PyPowSyBl users.

Any way to make this transparent / working out of the box ?
Or should we add an `iidm` platform module config so that variant manager initializes with multithread access enabled ?

geofjamg: We could automatically set this mode when running the multi threaded SA but that might be dangerous and not consistent with what we have done until today to change a network model when running a security. I think users expect that the SA only read data and not change anything in the data model.

So yes maybe we could just allow to configure the IIDM multithread access default value using the platform config.

geofjamg:
@jonenst @jeandemanged would it be useful to have a mode (for instance threadCount = -1) so that we automatically use all the threads of the machine ?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -96,7 +102,7 @@ public CompletableFuture<SecurityAnalysisReport> run(String workingVariantId, Se
         Objects.requireNonNull(contingenciesProvider);
         return CompletableFutureTask.runAsync(() -> {
             network.getVariantManager().setWorkingVariant(workingVariantId);
```
vidaldid-rte, line 104: Is this still necessary  since the runAsync is supposed to do it ?

vidaldid-rte: You probably need this to be done before loadContingencies(network).
But it would seem more consistent to me to do that in runAsync since you pass the workingVariantId as param.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -133,32 +146,102 @@ SecurityAnalysisReport runSync(SecurityAnalysisParameters securityAnalysisParame
...
+            // create networks including all necessary switches
+            try (LfNetworkList lfNetworks = Networks.load(network, parameters.getNetworkParameters(), topoConfig, saReportNode)) {
+                // run simulation on largest network
+                result = lfNetworks.getLargest().filter(n -> n.getValidity() == LfNetwork.Validity.VALID)
+                        .map(largestNetwork -> runSimulations(largestNetwork, propagatedContingencies, parameters, securityAnalysisParameters, operatorStrategies, actions, limitReductions))
+                        .orElse(createNoResult());
+            }
+        } else {
+            if (!network.getVariantManager().isVariantMultiThreadAccessAllowed()) {
+                throw new PowsyblException("To run a security analysis on multiple threads (and so on variants), you need to set allowVariantMultiThreadAccess to true");
```
vidaldid-rte, line 171: Is there a reason for not switching automatically to a thread local variant manager ? That would make the feature easier to use. And then document that using multithreading automatically switches the variant manager to the thread local version .

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
geofjamg: A security analysis mode, we do not update the network (IIDM) with N state.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -133,32 +146,102 @@ SecurityAnalysisReport runSync(SecurityAnalysisParameters securityAnalysisParame
...
+                    //      variant (for instance to get reactive capability curve), so allowVariantMultiThreadAccess mode
+                    //      is absolutely required
+                    //  so in order to be thread safe, we need to:
+                    //    - lock LF network creation (which create a working variant, see {@code LfNetworkList}
+                    //    - delay {@code LfNetworkList} closing (which remove a working variant) out of worker thread
+                    LfNetworkList lfNetworks;
+                    List<PropagatedContingency> propagatedContingencies;
+                    P parameters;
+                    networkLock.lock();
+                    try {
```
vidaldid-rte, line 202: I did not get why you need a lock since all threads will set the same variantId. And moreover the variant ID is local to  a thread.
Which function in this section is modifying the Network ?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -133,32 +146,102 @@ SecurityAnalysisReport runSync(SecurityAnalysisParameters securityAnalysisParame
...
+                        lfNetworksList.add(0, lfNetworks); // FIXME to workaround variant removal bug, to fix in core
+                    } finally {
+                        networkLock.unlock();
+                    }
 
-            return new SecurityAnalysisReport(result);
+                    // run simulation on largest network
+                    partitionResults.set(partitionNum, lfNetworks.getLargest().filter(n -> n.getValidity() == LfNetwork.Validity.VALID)
+                            .map(largestNetwork -> runSimulations(largestNetwork, propagatedContingencies, parameters, securityAnalysisParameters, operatorStrategies, actions, limitReductions))
+                            .orElse(createNoResult()));
```
vidaldid-rte, line 219: Isn't it easier here to use a Callable that using the synchronized list to store the resutls ?

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -3276,4 +3276,27 @@ void testPhaseTapChangerContingency() {
...
+        OpenSecurityAnalysisParameters securityAnalysisParametersExt = new OpenSecurityAnalysisParameters();
+        securityAnalysisParameters.addExtension(OpenSecurityAnalysisParameters.class, securityAnalysisParametersExt);
+
+        List<Contingency> contingencies = Stream.of("L1", "L2")
+                .map(id -> new Contingency(id, new BranchContingency(id)))
+                .toList();
+
+        SecurityAnalysisResult resultOneThread = runSecurityAnalysis(network, contingencies, Collections.emptyList(), securityAnalysisParameters);
+        securityAnalysisParametersExt.setThreadCount(2);
+        SecurityAnalysisResult resultTwoThreads = runSecurityAnalysis(network, contingencies, Collections.emptyList(), securityAnalysisParameters);
```
vidaldid-rte, line 3298: Note that this test won't detect any race condition issue in case some may happen today or in the future.
The number of threads is too small.

jonenst (approved):

antoinebhs (approved):

PR #1085, opened 2024-09-06
Fast restart for active power target change

geofjamg:

No

Feature

We are able to change target p of generators and batteries with fast restart

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/NetworkCache.java
+++ b/src/main/java/com/powsybl/openloadflow/NetworkCache.java
@@ -178,7 +178,7 @@ public enum NetworkCache {
                 } else if (attribute.equals("targetP")) {
                     double valueShift = (double) newValue - (double) oldValue;
                     LfGenerator lfGenerator = lfBus.getNetwork().getGeneratorById(generator.getId());
-                    double newTargetP = lfGenerator.getTargetP() + valueShift / PerUnit.SB;
+                    double newTargetP = lfGenerator.getInitialTargetP() + valueShift / PerUnit.SB;
                     lfGenerator.setTargetP(newTargetP);
                     lfGenerator.setInitialTargetP(newTargetP);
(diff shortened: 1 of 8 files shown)


src/main/java/com/powsybl/openloadflow/network/LfAction.java
```diff
@@ -335,6 +335,7 @@ private void applyGeneratorChange(LfNetworkParameters networkParameters) {
             double newTargetP = generatorChange.isRelative() ? generator.getTargetP() + generatorChange.activePowerValue() : generatorChange.activePowerValue();
             generator.setTargetP(newTargetP);
             generator.setInitialTargetP(newTargetP);
+            // FIXME should be use generator.reApplyActivePowerControlChecks() instead ?
```
geofjamg, line 338: @annetill if we do that we break some unit test ? 
Do this behaviour on purpose to not being able to get back the generator to active power control after contingency if target is now correct ?

annetill: It is an error of the code on main branch, your proposal is the right one and unit tests have to be fixed.

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -156,6 +157,16 @@ private CacheUpdateResult onInjectionUpdate(Injection<?> injection, BiFunction<A
             return CacheUpdateResult.elementNotFound();
         }
 
+        private static CacheUpdateResult updateLfGeneratorTargetP(String id, double oldValue, double newValue, AcLoadFlowContext context, LfBus lfBus) {
+            double valueShift = newValue - oldValue;
```
annetill, line 161: Just a question: why do you need this temporary "valueShift" ?

annetill (approved):

PR #1087, opened 2024-09-19
Add an option that allow checking unrealistic voltage only on final state of the loadflow.

murgeyseb:

No

Feature

Current implementation of unrealistic voltage check does stop computation with status Unrealistic Voltage each time any Newton Raphson iteration ends with a found voltage on any node outside the allowed range.

Adds an option that allow checking unrealistic voltage only on final state of the loadflow. It allows outerl oops to fix some voltage issues before failing.

- [ ] Yes
- [x] No


src/main/java/com/powsybl/openloadflow/ac/solver/NewtonRaphsonParameters.java
```diff
@@ -14,6 +14,13 @@
  */
 public class NewtonRaphsonParameters extends AbstractNewtonParameters<NewtonRaphsonParameters> {
 
+    public enum UnrealisticVoltageCheckBehavior {
```
annetill, line 17: @murgeyseb can you add the documentation of the new parameter you have introduced. Or I can do it if you prefer. Just tell me !

annetill (approved):
Just documentation missing.

vidaldid-rte (commented):
As I understand it, this PR is a workaround to the fact that remote voltage control can create very low voltages, either in intermediate state of the solution of in final states. It does not solve the problem and does not resolve all divergence issues due to low voltage with this algorithm. 
Unless there are other use cases for this PR, I would prefer that we do not commit this workaround before we have a full solution to stabilize the remote voltage algorithm.

vidaldid-rte:
Closing this PR. Is replaced by https://github.com/powsybl/powsybl-open-loadflow/pull/1135

Issue #948, opened 2024-01-09
Ac security analysis on several synchronous components


annetill:
### Describe the current behavior

Currently, the security analysis is performed on the main synchronous component, which is not sufficient for common grid model calculation.  

### Describe the expected behavior

The first solution is to run a security analysis on each synchronous component and to aggregate results. A better solution is not reachable easily because it means that we are able to run a load flow on a connected component without any loop.

PR #1089, opened 2024-09-24
Security analysis: run simulation on all components

obrix:

Yes : https://github.com/powsybl/powsybl-open-loadflow/issues/948

Security analysis simulation is now runned on all components of the network. Results are merged.

Security analysis was only runned on largest component.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
@@ -347,7 +347,7 @@ public abstract class AbstractSecurityAnalysis<V extends Enum<V> & Quantity, E e
     }
 
     private static <T> ArrayList<T> ensureMutable(List<T> orig) {
-        return orig instanceof ArrayList<T> arrayList? arrayList : new ArrayList<>(orig);
+        return orig instanceof ArrayList<T> arrayList ? arrayList : new ArrayList<>(orig);
     }
 
     static NetworkResult mergeNetworkResult(NetworkResult source, NetworkResult target) {
--- a/src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
@@ -262,7 +262,7 @@ public abstract class AbstractSecurityAnalysis<V extends Enum<V> & Quantity, E e
         NetworkResult mergedPreContingencyNetworkResult = result.getPreContingencyResult().getNetworkResult();
         List<LimitViolation> preContingenctViolations = result.getPreContingencyResult().getLimitViolationsResult().getLimitViolations();
 
-        if (lfParameters.getConnectedComponentMode().equals(LoadFlowParameters.ConnectedComponentMode.ALL)) {
+        if (lfParameters.getConnectedComponentMode() == LoadFlowParameters.ConnectedComponentMode.ALL) {
             Map<String, PostContingencyResult> postContingencyResultMap = new LinkedHashMap<>();
             Map<String, OperatorStrategyResult> operatorStrategyResultMap = new LinkedHashMap<>();
             postContingencyResults.forEach(r -> postContingencyResultMap.put(r.getContingency().getId(), r));
@@ -282,7 +282,7 @@ public abstract class AbstractSecurityAnalysis<V extends Enum<V> & Quantity, E e
                 mergedPreContingencyNetworkResult = mergeNetworkResult(mergedPreContingencyNetworkResult, resultOtherComponent.getPreContingencyResult().getNetworkResult());
 
                 // PostContingency and OperatorStrategies results
-                mergeSecurityAnalysisResult(resultOtherComponent, postContingencyResultMap, operatorStrategyResultMap);
+                mergeSecurityAnalysisResult(resultOtherComponent, postContingencyResultMap, operatorStrategyResultMap, n.getNumCC());
             }
             postContingencyResults = postContingencyResultMap.values().stream().toList();
             operatorStrategyResults = operatorStrategyResultMap.values().stream().toList();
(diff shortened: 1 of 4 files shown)


src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -244,14 +238,119 @@ SecurityAnalysisReport runSync(SecurityAnalysisParameters securityAnalysisParame
...
+                .findFirst()
+                .map(n -> runSimulations(n, propagatedContingencies, parameters, securityAnalysisParameters, operatorStrategies, actions, limitReductions))
+                .orElse(createNoResult());
+
+        List<PostContingencyResult> postContingencyResults = new ArrayList<>(result.getPostContingencyResults());
+        List<OperatorStrategyResult> operatorStrategyResults = new ArrayList<>(result.getOperatorStrategyResults());
+        NetworkResult mergedPreContingencyNetworkResult = result.getPreContingencyResult().getNetworkResult();
+        List<LimitViolation> preContingenctViolations = new ArrayList<>(result.getPreContingencyResult().getLimitViolationsResult().getLimitViolations());
+
+        if (lfParameters.getConnectedComponentMode().equals(LoadFlowParameters.ConnectedComponentMode.ALL)) {
```
vidaldid-rte, line 265: minor remark: you can use == here instead of equals since you compare enum values.

vidaldid-rte:
It seems that this implementation faces limitations due to the lack of result API in powsybl-core.
- convergence status  of secondary components are not reported in precontingency status
- convergence status of contingencies that affect multiple components are not reported (only the first first run component's status)

Users of this feature may miss severe contingencies that make the network unstable and result in a divergence.

If this has to be merged before the powsybl-core API supports multiple component security results, at the minimum the limtiations should be documented. 

Maybe in docs/security/security.md.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -244,14 +238,119 @@ SecurityAnalysisReport runSync(SecurityAnalysisParameters securityAnalysisParame
...
+                    .findAny();
+
+            if (optOriginalResult.isPresent()) {
+                PostContingencyResult originalResult = optOriginalResult.get();
+                NetworkResult mergedNetworkResult = mergeNetworkResult(postContingencyResult.getNetworkResult(), originalResult.getNetworkResult());
+                List<LimitViolation> violations = new ArrayList<>(postContingencyResult.getLimitViolationsResult().getLimitViolations());
+                violations.addAll(originalResult.getLimitViolationsResult().getLimitViolations());
+
+                PostContingencyResult mergedPostContingencyResult =
+                        new PostContingencyResult(originalResult.getContingency(), originalResult.getStatus(),
```
vidaldid-rte, line 302: if original status is different from the current contigency result (eg one converged the other failed), since you cannot report this with the current API, a warning should be logged (and maybe something reported to be able show this to an application user ?)
The best solution would however to have the right result API.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -3421,4 +3422,113 @@ private void testWithFictitiousLoad2(LoadFlowParameters.BalanceType balanceType)
...
+        assertNull(postContingencyResultsl13.getBranchResult("lc12Bis"));
+
+        // Result for post contingency on dummyLoad is available for part of the network on which the contingency has an impact
+        assertEquals(0.498, postContingencyResultslc12Bis.getBranchResult("lc12").getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(0.498, postContingencyResultslc12Bis.getBranchResult("lc12Bis").getP1(), LoadFlowAssert.DELTA_POWER);
+    }
+
+    @Test
+    void multiComponentSaTestContingencyBothComponentsAndOperatorStrategy() {
+
```
vidaldid-rte, line 3478: What happens if a remedial action reconnects two networks that are initially isolated ?

Is this supported ? Should this be a listed as a limitation of this feature ?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -244,14 +238,119 @@ SecurityAnalysisReport runSync(SecurityAnalysisParameters securityAnalysisParame
...
+
+                postContingencyResults.remove(originalResult);
+                postContingencyResults.add(mergedPostContingencyResult);
+            } else {
+                postContingencyResults.add(postContingencyResult);
+            }
+        });
+
+        resultToMerge.getOperatorStrategyResults().forEach(operatorStrategyResult -> {
+            Optional<OperatorStrategyResult> optOriginalResult = operatorStrategyResults.stream()
```
vidaldid-rte, line 313: This is very slow.
If you have N contingencies, you iterate on average N times (assuming duplicates are rare)
And the loop is performed for each contingency. Total complexity is quadratic vs number of contingencies [ N*(N-1)/2]

It would be better to use a Map.

obrix: I think the best way to fix that is to make the SecurityAnalysisResult API evolve so it allows us to either efficiently merge results, or store them per component and have an efficient API to query them.
I'll see if I can do better here but I wouldn't over optimize it too much on this side as this is a temporary fix and this logic should not be the permanent fix.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
obrix: I think behavior might depend on the network being isolated only after contingency and or also in the base case. This might be a good idea to test this case specifically yes.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
obrix: Yes, LoadFlowResult can be an inspiration on that.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
vidaldid-rte: Oups. I only saw your statement on not overoptimizing and didn(t see that you said you would try something.

I committed something to reduce iterations and list copies. Let me think if it works for you. Sorry if it conficlted with work you already started.
https://github.com/powsybl/powsybl-open-loadflow/pull/1089/commits/207787f7c2c64a9550f2eba00948115d75569dba

obrix: No problem, looks good to me.

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -15,10 +15,7 @@
 import com.powsybl.computation.ComputationManager;
 import com.powsybl.contingency.ContingenciesProvider;
 import com.powsybl.contingency.Contingency;
-import com.powsybl.iidm.network.Branch;
-import com.powsybl.iidm.network.Network;
-import com.powsybl.iidm.network.Switch;
-import com.powsybl.iidm.network.TieLine;
+import com.powsybl.iidm.network.*;
 import com.powsybl.loadflow.LoadFlowParameters;
```
vidaldid-rte, line 19: Is this on purpose ? (remove explicit imports ?)

vidaldid-rte (approved):
@obrix  this looks very good now. Approved with just one minor comment about explicit imports removed, probably by a misconfigured IDE...

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
obrix: Intellij is doing that when more than five imports are done from the same module. First time I really notice it but I think it is project wide.

PR #1093, opened 2024-09-27
Retrieve specific parameters overrides from PlatformConfig

antoinebhs:
**Requires next powsybl upgrade**

https://github.com/powsybl/powsybl-core/issues/3066


antoinebhs:
@SylvestreSakti, this requires the next powsybl upgrade

SylvestreSakti:
> @SylvestreSakti, this requires the next powsybl upgrade

Yes, ok, now I understand that the checks will work with the next release of core.

jeandemanged (dismissed):
Looking at https://github.com/powsybl/powsybl-core/issues/3066 we should probably apply the change as well to:
- `open-security-analysis-default-parameters`
- `open-sensitivityanalysis-default-parameters`

right ?

olperr1:
> Looking at [powsybl/powsybl-core#3066](https://github.com/powsybl/powsybl-core/issues/3066) we should probably apply the change as well to:
> 
>     * `open-security-analysis-default-parameters`
> 
>     * `open-sensitivityanalysis-default-parameters`
> 
> 
> right ?

@jeandemanged: You are right... but these providers were forgotten in powsybl/powsybl-core#3133.

olperr1:
> I will open an issue in `powsybl-core` on this subject, but for now these providers shouldn't be updated.

The issue is here: powsybl/powsybl-core#3175

geofjamg (approved):

PR #1095, opened 2024-10-02
[WIP] Add PST Actions in Fast DC Security Analysis

p-arvy:

- [ ] Yes
- [ ] No

**Other information**:


src/main/java/com/powsybl/openloadflow/dc/equations/AbstractClosedBranchDcFlowEquationTerm.java
```diff
@@ -54,6 +64,18 @@ protected AbstractClosedBranchDcFlowEquationTerm(LfBranch branch, LfBus bus1, Lf
         }
     }
 
+    /**
```
p-arvy, line 67: This part (aimed at updating the power according to the branch's current tap) should surely be implemented in a listener. As I'm not comfortable with OLF's listener management, the idea is to first validate the operation before implementing it more cleanly (if necessary). 

I note that this development should also help the “slow” DC SA, which doesn't seem to properly take into account the change in impedance, induced by modifying the tap via a PST action. So this could be dealt with in a separate PR.

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
```diff
@@ -0,0 +1,454 @@
...
+        parameters.setDc(true);
+        securityAnalysisParameters.setLoadFlowParameters(parameters);
+        OpenSecurityAnalysisParameters openSecurityAnalysisParameters = new OpenSecurityAnalysisParameters();
+        openSecurityAnalysisParameters.setDcFastMode(true);
+        securityAnalysisParameters.addExtension(OpenSecurityAnalysisParameters.class, openSecurityAnalysisParameters);
+    }
+
+    @Test
+    void testSaDcPhaseTapChangerTapPositionAction() {
+        Network network = MetrixTutorialSixBusesFactory.create();
```
vidaldid-rte, line 61: This looks very much like testSaDcPhaseTapChangerTapPositionAction
Can you, instead of duplicating the test, turn testSaDcPhaseTapChangerTapPositionAction into a parametrized test that would run in fast mode and in default mode ?

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
```diff
@@ -0,0 +1,454 @@
...
+        network.getTwoWindingsTransformer("NE_NO_1").getPhaseTapChanger().setTapPosition(originalTapPosition - 1);
+        loadFlowRunner.run(network, parameters);
+        // Compare results
+        assertEquals(network.getLine("S_SO_2").getTerminal1().getP(), brRel.getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getBranch("S_SO_2").getTerminal2().getP(), brRel.getP2(), LoadFlowAssert.DELTA_POWER);
+    }
+
+    @Test
+    void testFastSaDcPhaseTapChangerTapPositionChangeAdmittanceOnlyAlphaNull() {
+        Network network = PhaseControlFactory.createWithOneT2wtTwoLines();
```
vidaldid-rte, line 100: This test is new but it would be good to run it in fast and normal mode also as a pametrized test.

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
```diff
@@ -0,0 +1,454 @@
...
+        OperatorStrategyResult resultRel = getOperatorStrategyResult(result, "strategyTapRelChange");
+        BranchResult brRelL2 = resultRel.getNetworkResult().getBranchResult("L2");
+        BranchResult brRelPS1 = resultRel.getNetworkResult().getBranchResult("PS1");
+
+        // Apply contingency by hand
+        network.getLine("L1").getTerminal1().disconnect();
+        network.getLine("L1").getTerminal2().disconnect();
+        // Apply remedial action
+        network.getTwoWindingsTransformer("PS1").getPhaseTapChanger().setTapPosition(0);
+        OpenLoadFlowParameters.create(parameters).setLowImpedanceBranchMode(OpenLoadFlowParameters.LowImpedanceBranchMode.REPLACE_BY_MIN_IMPEDANCE_LINE);
```
vidaldid-rte, line 133: Why do you need to set this parameter ? (the test works also if you comment the line)

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
```diff
@@ -0,0 +1,454 @@
...
+        assertEquals(network.getTwoWindingsTransformer("PS1").getTerminal1().getP(), brAbsPS1.getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getTwoWindingsTransformer("PS1").getTerminal2().getP(), brAbsPS1.getP2(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getTwoWindingsTransformer("PS1").getTerminal1().getP(), brRelPS1.getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getTwoWindingsTransformer("PS1").getTerminal2().getP(), brRelPS1.getP2(), LoadFlowAssert.DELTA_POWER);
+    }
+
+    @Test
+    void testSaDcPhaseTapChangerTapPositionChangeWithConnectivityBreak() {
+        Network network = PhaseControlFactory.createNetworkWith3Buses();
+
```
vidaldid-rte, line 253: This is a strange network to test phase shifter action because there is no circuit parallel to the phase shifter.
Actions on the phase shifter are supposed to NOOP no ?

vidaldid-rte: I got it. The slack distribution fails.
Can you please fix the test to have correct power values ?
That may help have the test succeed also for normal mode.

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -222,10 +216,18 @@ public static <T> List<DcLoadFlowResult> run(T network, LfNetworkLoader<T> netwo
...
 
     /**
-     * A simplified version of DcLoadFlowEngine that supports on the fly bus and branch disabling and that do not
-     * update the state vector and the network at the end (because we don't need it to just evaluate a few equations).
+     * A simplified version of DcLoadFlowEngine that supports on the fly bus and branch disabling.
+     * Note that it does not update the state vector and the network at the end (because we don't need it to just evaluate a few equations).
      */
     public static double[] run(DcLoadFlowContext loadFlowContext, DisabledNetwork disabledNetwork, ReportNode reportNode) {
+        return run(loadFlowContext, disabledNetwork, reportNode, new ArrayList<>());
+    }
```
vidaldid-rte, line 224: Better to use Collections.emptyList() here

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -222,10 +216,18 @@ public static <T> List<DcLoadFlowResult> run(T network, LfNetworkLoader<T> netwo
...
+
+    /**
+     * A simplified version of DcLoadFlowEngine that supports on the fly bus and branch disabling, and pst actions.
+     * Note that it does not update the state vector and the network at the end (because we don't need it to just evaluate a few equations).
+     */
+    public static double[] run(DcLoadFlowContext loadFlowContext, DisabledNetwork disabledNetwork, ReportNode reportNode, List<LfAction> lfActions) {
         Collection<LfBus> remainingBuses;
```
vidaldid-rte, line 231: This method:
    Is Badly Named. It only supports PST actions. That should be reflected
   Is specific to the WoodBuryEngine (because it has limitations on actions) and should not be in this class.

vidaldid-rte: Suggestion for refactoring
   Rename run as runWithModifiedTargerVector
   In DCLoadFlowEngine:
        create a an method cloneTargetVector with context, disabledNetwork as argument
        the runWithModifiedTargetVector calls this, then run
   In WoodburyEngine
     create a runWithModifiedTargetVector that takes also the list of supported actions
     call the pritected clonetTargetVector from super
     perform the additional modifications from the actions
     then run

src/main/java/com/powsybl/openloadflow/dc/equations/AbstractClosedBranchDcFlowEquationTerm.java
vidaldid-rte: If there is not managed by the listerner mechanism, how does the lastTapPosition work with the DcIncrementalPhaseControlOuterLoop ?

vidaldid-rte: It seems to me that the right way to do this would be to catch the tapChange events and update the power in the equation term.

vidaldid-rte: And for the "optimized run" that doesn't use the notifications but clones the target vector (is it really faster by the way ?) then call explicitely the eqaution terms attached to the PST modified by the actions.

src/main/java/com/powsybl/openloadflow/dc/equations/ClosedBranchSide1DcFlowEquationTerm.java
```diff
@@ -38,18 +38,18 @@ public static ClosedBranchSide1DcFlowEquationTerm create(LfBranch branch, LfBus
     @Override
     protected double calculateSensi(double ph1, double ph2, double a1) {
         double deltaPhase = ph2 - ph1 + A2 - a1;
-        return -power * deltaPhase;
+        return -calculatePower(element.getPiModel()) * deltaPhase;
     }
```
vidaldid-rte, line 42: If you had a proper notifiation mechanism you wouldn't have to recalculate the value each time and could continue to cache it.

src/main/java/com/powsybl/openloadflow/dc/equations/AbstractClosedBranchDcFlowEquationTerm.java
vidaldid-rte: After more thoughts I still don't know what is best between a notification based implem that would catch the tap change events, that more complex, and an "always recompute approach" that is simple and good enough if the equation is not called to frequently.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
```diff
@@ -149,6 +149,10 @@ public PiModel getPiModel() {
         return piModel;
     }
 
+    public PiModel getPiModel(int tapPosition) {
```
vidaldid-rte, line 152: Why did you add this method ? It is never used.

src/main/java/com/powsybl/openloadflow/network/SimplePiModel.java
```diff
@@ -213,4 +213,9 @@ public PiModel setTapPosition(int tapPosition) {
     public Range<Integer> getTapPositionRange() {
         throw new IllegalStateException(NO_TAP_POSITION_ERROR);
     }
+
```
vidaldid-rte, line 216: Same remark as for the interface. Can we get rid og this method ?

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
p-arvy: Indeed. Thank you.

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
p-arvy: Indeed, thank you! I've added a line parallel to the PST, to make the TU more consistent. 

About slack distribution, it seems correct in fast DC SA. Here, the generator had a Pmax of 150, so it was not able to go up to 200MW of production after distribution. By changing the Pmax of the generator, the distribution is done correctly.

Slack distribution in normal mode should be fixed in another PR :)

vidaldid-rte: @p-arvy  your test was interesting because it helped me find two issues that I plan to address in separate PRs:
      - SlackDistribtionFailuerBheaviour FAIL mode is not honnored by the DC engine
      - The normal DC mode behaves wrongly when a load is lost because of a connectivyt change.

But I wonder why you made a test with unbalanced power and in a mode where the remaining power is left on a slack bus.
If you don't force the slack bus in the test, the value of power in the branches is not really determined because it depends on the slack bus determination algorithm.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
```diff
@@ -79,20 +79,40 @@ protected DcLoadFlowParameters createParameters(LoadFlowParameters lfParameters,
      */
     private double[] calculatePostContingencyStatesForAContingency(DcLoadFlowContext loadFlowContext, DenseMatrix contingenciesStates, double[] flowStates,
                                                                    PropagatedContingency contingency, Map<String, ComputedContingencyElement> contingencyElementByBranch,
-                                                                   Set<LfBus> disabledBuses, Set<String> elementsToReconnect, ReportNode reportNode, Set<LfBranch> partialDisabledBranches) {
+                                                                   Set<LfBus> disabledBuses, Set<String> elementsToReconnect, Set<LfBranch> partialDisabledBranches, ReportNode reportNode) {
+        return calculatePostContingencyStatesForAContingency(loadFlowContext, contingenciesStates, flowStates, contingency, contingencyElementByBranch, disabledBuses, elementsToReconnect,
+                partialDisabledBranches, List.of(), new HashMap<>(), new DenseMatrix(0, 0), reportNode);
+    }
```
vidaldid-rte, line 85: Can you define somewhere en EmptyDenseMatrice that would created only once ?
As seen in https://github.com/powsybl/powsybl-open-loadflow/pull/1109
DenseMatrix should not be created frequently and should be reused as much as possible.
Because the native memory they allocate as a direct buffer is not efficiently manager by the Java Garbage Collector.
Even if it is a small one here, it would be better to use a singleton.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
```diff
@@ -225,10 +316,15 @@ protected SecurityAnalysisResult runSimulations(LfNetwork lfNetwork, List<Propag
             // compute states with +1 -1 to model the contingencies and run connectivity analysis
             ConnectivityBreakAnalysis.ConnectivityBreakAnalysisResults connectivityBreakAnalysisResults = ConnectivityBreakAnalysis.run(context, propagatedContingencies);
 
-            // save base state for later restoration after each contingency
+            // compute states with +1 -1 to model the actions in Woodbury engine
+            Map<String, ComputedActionElement> computedActionElements = createActionsElementsIndexByBranchId(lfActionById, context.getEquationSystem());
+            DenseMatrix actionsStates = ComputedElement.calculateElementsStates(context, computedActionElements.values());
+
```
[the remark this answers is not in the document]
vidaldid-rte: Not sure of my remarks. If it is run once per LFNetwork it might be OK. THe problem is when one matri is allocated per contingency run.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
p-arvy: Yes you are right. Thank you.
I created the singleton in `EmptyDenseMatrix` static class (in `WoodburyEngine`). 
To be clean, I guess this empty matrix should be moved in `DenseMatrix` class. What do you think @vidaldid-rte ?

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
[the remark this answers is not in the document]
p-arvy: Indeed, here the `DenseMatrix` will be created only once per `LfNetwork`. So seems okay.
Also, thanks to PR #1111, `postContingencyAndActionsStates `management will be more efficient.

src/main/java/com/powsybl/openloadflow/network/PiModel.java
```diff
@@ -77,4 +77,6 @@ public interface PiModel {
     PiModel setTapPosition(int tapPosition);
 
     Range<Integer> getTapPositionRange();
+
+    PiModel getModel(int tapPositionIndex);
```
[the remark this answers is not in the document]
p-arvy: Yes you're right, thanks for the comment. 

I modified `TapPositionChange` to check on creation that the branch has a `PiModelArray`. Access to the newPiModel is also interfaced, as I need it to calculate the power in the `WoodburyEngine` (in addition to the new phase shift).

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBranch.java
p-arvy: Solved in another comment.

src/main/java/com/powsybl/openloadflow/network/SimplePiModel.java
p-arvy: Solved in another comment.

src/main/java/com/powsybl/openloadflow/dc/equations/AbstractClosedBranchDcFlowEquationTerm.java
p-arvy: I note that in the current state of the PR, the power is only updated when there is a call with a tap position different from the current tap of the equation term. If notification based implem is chosen, it would be better to catch the tapChange events to apply the modification, as you mentioned above.

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
```diff
@@ -105,22 +193,26 @@ public DenseMatrix run(DenseMatrix preContingencyStates) {
...
-     * Calculate post-contingency states values using pre-contingency states values and some flow transfer factors (alphas).
-     * @return an array of post-contingency voltage angle states (which is the same modified pre-contingency java object).
+     * Calculate post-contingency and post-actions states values, using pre-contingency states values and some flow transfer factors (alphas).
+     * @return an array of post-contingency and post-actions voltage angle states (which is the same modified pre-contingency java object).
      */
     public double[] run(double[] preContingencyStates) {
         Objects.requireNonNull(preContingencyStates);
-        double[] postContingencyStates = preContingencyStates;
+        double[] postContingencyAndOperatorStrategyStates = preContingencyStates;
         setAlphas(new DenseMatrix(preContingencyStates.length, 1, preContingencyStates), 0);
```
vidaldid-rte, line 202: Please make sure that this frequent allocation of a new DenseMatrix will not create memory issues on large networks with a large number contingencies (that would create many allocations), as we have seen for sensitivity analysis.

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
```diff
@@ -105,22 +193,26 @@ public DenseMatrix run(DenseMatrix preContingencyStates) {
     }
 
     /**
-     * Calculate post-contingency states values using pre-contingency states values and some flow transfer factors (alphas).
-     * @return an array of post-contingency voltage angle states (which is the same modified pre-contingency java object).
+     * Calculate post-contingency and post-actions states values, using pre-contingency states values and some flow transfer factors (alphas).
+     * @return an array of post-contingency and post-actions voltage angle states (which is the same modified pre-contingency java object).
      */
     public double[] run(double[] preContingencyStates) {
```
vidaldid-rte, line 199: Is there a reason why you need two methods run ?
  - If there was only one method the problem of DenseMatrix allocation would be solved once. But here we see new cases where a densMatrix is created.
  -  It also creates a problem that two methods with the same name do different things. One will use tapPositionChangeElements if defined in the class, the other won't.
  - It also causes some low level code duplication (ex formulas of postContingencyValue using the Alpha.

vidaldid-rte: (in any case, to avoid conflicts - we should wait for a rebase on https://github.com/powsybl/powsybl-open-loadflow/pull/1111 - once merged - before working an a refacto)

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
vidaldid-rte: @p-arvy I reopen that comment.
I think that every test on DC security should run on fast DC and default DC unless there is a good reason.
We want to know if there is a difference - and it might be a bug.
As of today, 5 tests filas in normal mode.
It is OK for two of them:
   testFastDcSaWithActionNotOnPst because the limitation is only for fast DC
   testSaDcPhaseTapChangerTapPositionChange because there is an issue in normal DC when connectivity is broken, and this is independant from this PR.  (I have added a todo for me to look at this)

I do not think it should be OK for 
   testFastSaDcPhaseTapChangerTapPositionChangeAdmittanceOnlyAlphaNonNull
   testFastSaDcN2ContingencyOneTapPositionChange
   testSaDcPhaseTapChangerTapPositionChange
I have not looked at the details yet but I suspect it is linked to the fact that tap changes are taken into account in AbstractClosedBranchDcFlowEquationTerm but in a way does doesn work well in all cases. If is is true, that should be fixed before the PR is merged.

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
vidaldid-rte: I reopen. This point was not adressed: If the slack is not fully distributed, and if you don't force the slack bus in the test, the value of power in the branches is not really determined because it depends on the slack bus determination algorithm.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
```diff
@@ -282,6 +410,36 @@ protected SecurityAnalysisResult runSimulations(LfNetwork lfNetwork, List<Propag
 
                                 postContingencyResults.add(postContingencyResult);
                                 networkState.restore();
+
```
vidaldid-rte, line 413: There is a lot of code in common between the breaking connectivity case and the non breaking connectivity case.
This code duplication will cause maintainance issues because modifications have to be reported to both branches of code.
What is the real difference between the breaking and non breaking connectivity case ? Is it only the call to calculatePostContingencyStatesForAContingencyBreakingConnectivity instead of calculatePostContingencyAndOperatorStrategyStatesForAContingency ?
In that case can case can you factorise the processing of a contingency (breaking connectivyity or not) in a method that would take an argument breaksConnectivity ?
That would make the code easier to understand and to maintain moving forward.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
```diff
@@ -125,11 +148,21 @@ private double[] calculatePostContingencyStatesForAContingency(DcLoadFlowContext
...
+                                                                                       Map<String, ComputedContingencyElement> contingencyElementByBranch, double[] flowStates, DenseMatrix contingenciesStates,
+                                                                                       ReportNode reportNode) {
+        return calculatePostContingencyAndOperatorStrategyStatesForAContingencyBreakingConnectivity(connectivityAnalysisResult, loadFlowContext, contingencyElementByBranch, flowStates,
+                contingenciesStates, Collections.emptyList(), Collections.emptyMap(), WoodburyEngine.EmptyDenseMatrix.getInstance(), reportNode);
+    }
+
+    /**
+     * Calculate post contingency and post operator strategy states, for a contingency breaking connectivity.
+     */
+    private double[] calculatePostContingencyAndOperatorStrategyStatesForAContingencyBreakingConnectivity(ConnectivityBreakAnalysis.ConnectivityAnalysisResult connectivityAnalysisResult, DcLoadFlowContext loadFlowContext,
```
vidaldid-rte, line 163: The code calling this function has those comments
                                // no need to distribute active mismatch due to connectivity modifications
                                // this is handled when the slack is distributed in pre contingency states override

I see nothing in the method that handles distributed slack among connex components.
How dos this behave if a contingency breaks connectivity and you end up with two parts where in **each** part
sum of loads != sum of production ? Do you have a test for that N

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
```diff
@@ -0,0 +1,427 @@
...
+        // Compare results on the t2wt PS1
+        assertEquals(network.getTwoWindingsTransformer("PS1").getTerminal1().getP(), brAbsPS1.getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getTwoWindingsTransformer("PS1").getTerminal2().getP(), brAbsPS1.getP2(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getTwoWindingsTransformer("PS1").getTerminal1().getP(), brRelPS1.getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getTwoWindingsTransformer("PS1").getTerminal2().getP(), brRelPS1.getP2(), LoadFlowAssert.DELTA_POWER);
+    }
+
+    @Test
+    void testSaDcPhaseTapChangerTapPositionChangeWithConnectivityBreak() {
+        Network network = PhaseControlFactory.createNetworkWith3Buses();
```
vidaldid-rte, line 218: This test doesn't really test a connectivity break. The remaining part after the L23 contingency is just an isolated bus without any branches.

src/main/java/com/powsybl/openloadflow/dc/equations/AbstractClosedBranchDcFlowEquationTerm.java
vidaldid-rte: Closed by https://github.com/powsybl/powsybl-open-loadflow/pull/1095/commits/c03d5665c12b1526ebcd99b35cfde5acacb2a123

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
vidaldid-rte: Current state fied by https://github.com/powsybl/powsybl-open-loadflow/pull/1095/commits/c03d5665c12b1526ebcd99b35cfde5acacb2a123
I keep this thread open as a reminder that all new tests should run in fast dc ad default mode unless there is a justified exception.

src/main/java/com/powsybl/openloadflow/dc/equations/ClosedBranchSide1DcFlowEquationTerm.java
p-arvy: Fixed in commit [c03d566](https://github.com/powsybl/powsybl-open-loadflow/commit/c03d5665c12b1526ebcd99b35cfde5acacb2a123).

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
p-arvy: In this case I propose to rename both methods `runDcLoadFlowWithModifiedTargetVector`, and move them in `WoodburyEngine` (to be specific to fast DC). Do you agree with that @vidaldid-rte ?

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
p-arvy: Work in progress [here](https://github.com/powsybl/powsybl-core/pull/3216).

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
p-arvy: This method will call `calculatePostContingencyAndOperatorStrategyStatesForAContingency`, which will call `DcLoadFlowEngine.run` (because the list of disabled buses is not empty, as there is a connectivity break). In this method, slack is distributed **on buses that are not disabled**. 

So if a contingency breaks connectivity and there is two parts where in each part sum of loads != sum of production, only the active slack of main connex component (including slack bus) will be distributed among buses of that component. 

The already existing TU `testConnectivityResultOnSplitDc` illustrates this behavior. Both components are unbalanced after the contingency, but slack is distributed correctly **in fast DC SA**.

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
p-arvy: I'm not sure to get the point. Why is that a problem if power in branches depends on slack distribution ?

vidaldid-rte: It depends on slack bus location. So the test may break if the default algorithm to define the slack bus changes. If you decide to make a test on an unbalanced network with assertions that depend on slack bus location it is better to force the slack bus location in the test.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
vidaldid-rte: Thanks. I think it is time to refactor because this PR shows that both branch have to evolve in parallel.

src/test/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysisWithActionsTest.java
vidaldid-rte: Thanks for the clarification Pierre. I understand that this test is more about what happens to the main component when many injections are lost due to a split of component in two parts, than what happens to the separate component.
And it is a very useful test since it shows this does not work well in default DC mode.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
vidaldid-rte: I gave a closer look and this is what happens:
    If a contingency breaks a network in two parts,  all bus of the second part are disabled.
   So the broken part is not computed, even in fast DC. So there's no issue if there is no slack bus on this part.

   This being said, I wonder what happens if the slack bus is on the disconnected part of the network because relocateSlackBus is passed to false int toLfContingency.

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
vidaldid-rte: I'm closing that topic after the method renaming, although only one of the three points is adressed.
  - The duplicate code is small
  - the DenseMatrix allocation is only used in the optional fastDC SA and won't cause rregressions.

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
p-arvy: After some thought, this is more of a refactoring task than development related to the PST feature. Therefore, it should be handled in a separate PR. I’ve opened #1126 here accordingly (can be handled before or after merging current PR).

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
vidaldid-rte: I see that you have added a note in the doc related to this:
<pre>
--- doc extract
- Slack relocation following the application of a contingency is not supported. 
As a result, security analysis is carried out only in slack component, and not necessarily in the larger one.
</pre>

Do you have a test that shows the behaviour when the slack but is not in the main component after a contingency breaking connectivity ? That documented behaviour didn't seem clear to me from the code - I didn't find anything related to that in toLfContngency (the main component doesn't seem to be changed if relocateSlackBus is false ans the slack bus is isolated)

vidaldid-rte: After stepping into the code, the documented behaviour might be true but:
   - 1) There is no test the verifies this. Can you add one ? 
   - 2) This seems to be due by a side effect of network.getConnectvity called at the beginning of PropagaetdContingency.findBusesAndBranchesImpactedBecauseOfConnectivityLoss (it sets the main component to the one containing the slack bus). Can you add a comment before the call to connectivity = network.getConnectivity(); to explain that this redefines the main compoment to the one containinge the first slack bus ? That would make the code easier to understand.

p-arvy: I added the comment, and the TU `testDcFastModeInFirstSlackComponentOnly` to clarify this behaviour. Thanks for the careful review @vidaldid-rte.

src/main/java/com/powsybl/openloadflow/sensi/ComputedElement.java
```diff
@@ -0,0 +1,131 @@
+/**
```
vidaldid-rte, line 1: You are adding new classes that are not specific to sensi at all (since you developper them for actions on SA)
Instead of adding more classes to the sensi package, why not moving everything that is specific to Woodbury to a new package com.powsybl.openloadflow.dc.fastdc ?

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
vidaldid-rte: Closing this one. Discussion to be reopen if we see memory issues in large models.

src/main/java/com/powsybl/openloadflow/sensi/ComputedElement.java
p-arvy: I agree 100%. In order to simplify the review of the modifications of this PR (as many elements specific to Woodbury are modified), I propose to move the code in another PR. I open it ASAP.

p-arvy: Done in #1145.

p-arvy: Merged. I resolve the conversation.

Hadrien-Godard (approved):

PR #1096, opened 2024-10-04
Fix Security Analysis in case of no more voltage regulated bus

jeandemanged:

No

Bug fix

- Exception occurring when island is a single bus with a generator on voltage control, and the contingency removes the generator (`PowsyblException: Expected to have same number of equations (2) and variables (1)`)
- Also I think something should be clarified about what happens if a contingency leaves the network without any bus under voltage control, I think it is strange to have the NR converging, why not mathematically speaking, but from a physical point of view this is surprising. Today we rely on unrealistic state check, but I think this is not correct.

`AcLoadFlowEngine` now examines the equation system and returns `SOLVER_FAILED` status in case of no `BUS_TARGET_V` equation found. Add also log and report for this situation.

- [x] No

**Other information**:

Case found on real networks since the introduction of #1089 when running SA on all components.

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
@@ -15,7 +15,6 @@ import com.powsybl.openloadflow.ac.outerloop.AcOuterLoop;
 import com.powsybl.openloadflow.ac.outerloop.AreaInterchangeControlOuterloop;
 import com.powsybl.openloadflow.ac.outerloop.DistributedSlackOuterLoop;
 import com.powsybl.openloadflow.ac.solver.*;
-import com.powsybl.openloadflow.equations.Equation;
 import com.powsybl.openloadflow.lf.LoadFlowEngine;
 import com.powsybl.openloadflow.lf.outerloop.DistributedSlackContextData;
 import com.powsybl.openloadflow.lf.outerloop.OuterLoopResult;
@@ -91,9 +90,9 @@ public class AcloadFlowEngine implements LoadFlowEngine<AcVariableType, AcEquati
             if (outerLoopResult.status() == OuterLoopStatus.UNSTABLE) {
                 LOGGER.debug("Start outer loop '{}' iteration {}", outerLoop.getName(), runningContext.outerLoopTotalIterations);
 
-                ReportNode nrReportNode = context.getNetwork().getReportNode();
+                ReportNode reportNode = context.getNetwork().getReportNode();
                 if (context.getParameters().isDetailedReport()) {
-                    nrReportNode = Reports.createDetailedSolverReporterOuterLoop(nrReportNode,
+                    reportNode = Reports.createDetailedSolverReporterOuterLoop(reportNode,
                             solver.getName(),
                             context.getNetwork().getNumCC(),
                             context.getNetwork().getNumSC(),
@@ -102,7 +101,7 @@ public class AcloadFlowEngine implements LoadFlowEngine<AcVariableType, AcEquati
                 }
 
                 // if not yet stable, restart solver
-                runningContext.lastSolverResult = solver.run(new PreviousValueVoltageInitializer(), nrReportNode);
+                runningContext.lastSolverResult = solver.run(new PreviousValueVoltageInitializer(), reportNode);
 
                 runningContext.nrTotalIterations.add(runningContext.lastSolverResult.getIterations());
                 runningContext.outerLoopTotalIterations++;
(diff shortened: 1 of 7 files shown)


src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -3779,8 +3777,39 @@ void testSlackBusSelectionExcludeBusWithHighestVoltage() {
...
+                .setTargetV(400.)
+                .setVoltageRegulatorOn(true)
+                .add();
+
+        List<Contingency> contingencies = List.of(new Contingency("G1", new GeneratorContingency("G1")));
+
+        LoadFlowParameters parameters = new LoadFlowParameters();
+        OpenLoadFlowParameters.create(parameters);
+        SecurityAnalysisResult result = runSecurityAnalysis(network, contingencies, parameters);
+
```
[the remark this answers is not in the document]
vidaldid-rte: @jeandemanged  I have added the test case.  But the AC fails with the G1 contingency while the N succeds if G1 is disconnected.

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -128,6 +128,26 @@ public AcLoadFlowResult run() {
...
+        ReportNode reportNode = context.getNetwork().getReportNode();
+
+        // Verify whether a regulated bus voltage exists.
+        // If not, then fail immediately with SOLVER_FAILED status.
+        // Note that this approach is not perfect and could be improved in the future, in particular in
+        // security analysis and remedial actions context.
+        // For example: a contingency may cause the last PV node to disappear. In this case here we would
+        // just report SOLVER_FAILED. However, there could be other generators blocked at MinQ or MaxQ that
+        // _could potentially_ recover the situation, but this will not be tried at all...
+        boolean hasVoltageRegulatedBus = context.getEquationSystem().getEquations()
```
vidaldid-rte, line 141: I need to understand the cost of this iteration and if it is acceptable since we go through this at each run of each contingency and each conex component.

vidaldid-rte: The duration of this test is negligible from my measures (<1ms) even if sometimes you need to iterate several thousands of time to find the first active PV bus.

vidaldid-rte (dismissed):
Slowdown of additional test looks acceptable.

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
vidaldid-rte: But this seems much faster
boolean hasVoltageRegulatedBus = context.getNetwork().getBuses().stream().anyMatch(b -> b.isGeneratorVoltageControlEnabled() && !b.isDisabled() && !b.getGeneratorVoltageControl().get().isDisabled())

jeandemanged: Thank you, I changed with your better way

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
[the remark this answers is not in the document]
jeandemanged: added a note about this limitation, this behavior was pre-existing to this PR, so I propose to work on it later

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -128,6 +128,25 @@ public AcLoadFlowResult run() {
...
+
+        // Verify whether a regulated bus voltage exists.
+        // If not, then fail immediately with SOLVER_FAILED status.
+        // Note that this approach is not perfect and could be improved in the future, in particular in
+        // security analysis and remedial actions context.
+        // For example: a contingency may cause the last PV node to disappear. In this case here we would
+        // just report SOLVER_FAILED. However, there could be other generators blocked at MinQ or MaxQ that
+        // _could potentially_ recover the situation, but this will not be tried at all...
+        boolean hasVoltageRegulatedBus = context.getNetwork().getBuses().stream()
+                .anyMatch(b -> b.isGeneratorVoltageControlEnabled() && !b.isDisabled() && !b.getGeneratorVoltageControl().orElseThrow().isDisabled());
```
olperr1, line 142: Isn't it dangerous to have a `orElseThrow()` here?

The `AbstractLfBus.setGeneratorVoltageControlEnabled(boolean)` method doesn't ensure that the generator voltage control is present when it is called with `true`.

Maybe something like this would be more robust:
```java
boolean hasVoltageRegulatedBus = context.getNetwork().getBuses().stream()
        .anyMatch(b -> b.isGeneratorVoltageControlEnabled() && !b.isDisabled()
                && !b.getGeneratorVoltageControl().map(GeneratorVoltageControl::isDisabled).orElseGet(() -> {
                    LOGGER.error("Bus {} is said to have a generator voltage control enabled, but has no generator voltage control", b.getId());
                    ... // Report the error in reportNode 
                    return Boolean.TRUE;
                }));
```

olperr1: OK, you are right. I haven't seen that `hasGeneratorVoltageControllerCapability()` (called in the setter) checks the presence of a generator voltage control.

geofjamg (approved):

PR #1097, opened 2024-10-04
Fix SA exception "No slack bus for network"

obrix:

updateSlackBusesAndReferenceBus implements a slack bus selection algorithm using an exclude list of buses in excludedSlackBuses.

Some time a bus is selected because it is the one and only with the highest voltage. If this bus is also in the exclude list then we have no solution available.

I changed a bit the algorithm by filtering the bus list with the exclude list before applying the selection algorithm so it is not able to base its selection on buses that will not be available anyway.

To illustrate the bug surfaced on a small network similar to this : 

B1(vl 150) ---- B2(vl 150) ---- B3 (vl380)

A contingency cut the line between B2 and B3. B3 is  then put in the exclude listof selectable slack bus because it is no longer in the main connected component after contingency.
But B3 is also the one with the highest voltage level. So it was selected by the algorithm but ultimately discarded because of the exclude list.

Filtering the list of buses with the exclude list before the selection process fix the issue.

Bug fix

IllegalStateException : "No slack bus for network " thrown

Correct slack bus is selected and no exception thrown

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
@@ -214,14 +214,14 @@ public class LfNetwork extends AbstractPropertyBag implements PropertyBag {
 
     public void updateSlackBusesAndReferenceBus() {
         if (slackBuses == null && referenceBus == null) {
-            SelectedSlackBus selectedSlackBus = slackBusSelector.select(busesByIndex, maxSlackBusCount);
-            slackBuses = selectedSlackBus.getBuses().stream()
-                    .filter(bus -> !excludedSlackBuses.contains(bus))
-                    .toList();
+            List<LfBus> selectableBus = busesByIndex.stream()
+                .filter(bus -> !excludedSlackBuses.contains(bus))
+                .toList();
+            SelectedSlackBus selectedSlackBus = slackBusSelector.select(selectableBus, maxSlackBusCount);
+            slackBuses = selectedSlackBus.getBuses();
             if (slackBuses.isEmpty()) { // ultimate fallback
-                selectedSlackBus = SLACK_BUS_SELECTOR_FALLBACK.select(busesByIndex, excludedSlackBuses.size() + maxSlackBusCount);
+                selectedSlackBus = SLACK_BUS_SELECTOR_FALLBACK.select(selectableBus, excludedSlackBuses.size() + maxSlackBusCount);
                 slackBuses = selectedSlackBus.getBuses().stream()
-                        .filter(bus -> !excludedSlackBuses.contains(bus))
                         .limit(maxSlackBusCount)
                         .toList();
             }
--- a/src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
@@ -223,7 +223,7 @@ public class LfNetwork extends AbstractPropertyBag implements PropertyBag {
             if (slackBuses.isEmpty()) { // ultimate fallback
                 selectedSlackBus = SLACK_BUS_SELECTOR_FALLBACK.select(selectableBus, maxSlackBusCount);
                 if (selectedSlackBus.getBuses().isEmpty()) {
-                    throw new PowsyblException("Slack bus selection process has failed after multiple fallback");
+                    throw new PowsyblException("No slack bus could be selected");
                 }
                 slackBuses = selectedSlackBus.getBuses();
             }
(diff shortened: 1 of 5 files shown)


src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -211,14 +211,14 @@ private void invalidateSlackAndReference() {
...
-                    .filter(bus -> !excludedSlackBuses.contains(bus))
-                    .toList();
+            List<LfBus> selectableBus = busesByIndex.stream()
+                .filter(bus -> !excludedSlackBuses.contains(bus))
+                .toList();
+            SelectedSlackBus selectedSlackBus = slackBusSelector.select(selectableBus, maxSlackBusCount);
+            slackBuses = selectedSlackBus.getBuses();
             if (slackBuses.isEmpty()) { // ultimate fallback
-                selectedSlackBus = SLACK_BUS_SELECTOR_FALLBACK.select(busesByIndex, excludedSlackBuses.size() + maxSlackBusCount);
+                selectedSlackBus = SLACK_BUS_SELECTOR_FALLBACK.select(selectableBus, excludedSlackBuses.size() + maxSlackBusCount);
```
SylvestreSakti, line 220: Hello, a small remark here : There is no excluded slack buses in `selectableBus` so the limit can be modified to just `maxSlackBusCount` ?

Is there a use case that can cover this line ? Maybe using a NameSlackBusSelector ? It seems that Sonar wants more coverage on this file :)

obrix: Yes you are right thanks. I am thinking of an example to test fallback mechanism right now.

SylvestreSakti:
@obrix Yes it seems hard to test this corner case. I agree with you that it can be merged as it is.

To clarify : do committers have the right to merge against SonarCloud ? Or should it be done by the owner ?

SylvestreSakti (approved):

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -214,16 +215,17 @@ private void invalidateSlackAndReference() {
 
     public void updateSlackBusesAndReferenceBus() {
         if (slackBuses == null && referenceBus == null) {
-            SelectedSlackBus selectedSlackBus = slackBusSelector.select(busesByIndex, maxSlackBusCount);
-            slackBuses = selectedSlackBus.getBuses().stream()
-                    .filter(bus -> !excludedSlackBuses.contains(bus))
-                    .toList();
+            List<LfBus> selectableBus = busesByIndex.stream()
+                .filter(bus -> !excludedSlackBuses.contains(bus))
+                .toList();
```
jeandemanged, line 220: rename `selectableBus` to `selectableBuses` 

also wouldn't it beneficial performance wise to do this only if excludedSlackBuses is not empty ? (avoiding an unnecessary copy of the list)

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
```diff
@@ -214,16 +215,17 @@ private void invalidateSlackAndReference() {
...
+            slackBuses = selectedSlackBus.getBuses();
             if (slackBuses.isEmpty()) { // ultimate fallback
-                selectedSlackBus = SLACK_BUS_SELECTOR_FALLBACK.select(busesByIndex, excludedSlackBuses.size() + maxSlackBusCount);
-                slackBuses = selectedSlackBus.getBuses().stream()
-                        .filter(bus -> !excludedSlackBuses.contains(bus))
-                        .limit(maxSlackBusCount)
-                        .toList();
+                selectedSlackBus = SLACK_BUS_SELECTOR_FALLBACK.select(selectableBus, maxSlackBusCount);
+                if (selectedSlackBus.getBuses().isEmpty()) {
+                    throw new PowsyblException("Slack bus selection process has failed after multiple fallback");
```
jeandemanged, line 226:
```suggestion
                    throw new PowsyblException("No slack bus could be selected");
```

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/network/LfNetwork.java
obrix: I guess for big network this can have an impact. Fixed.

jeandemanged (approved):

obrix:
> @obrix Yes it seems hard to test this corner case. I agree with you that it can be merged as it is.
> 
> To clarify : do committers have the right to merge against SonarCloud ? Or should it be done by the owner ?

@SylvestreSakti I think once we agreed committer can do it, anyway we managed to cover it, thanks for the review!

PR #1099, opened 2024-10-11
[WIP] (OLD : see #1122 and #1123) DC Area Interchange Control Outer Loop

vmouradian:
HAS BEEN SPLITTED INTO 2 PRs :  

#1122 
#1123 

feature

AIC not supported for DC Loadflow

AIC outerloop is supported for DC Loadflow

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
+++ b/src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
@@ -15,7 +15,6 @@ import com.powsybl.openloadflow.dc.equations.DcEquationType;
 import com.powsybl.openloadflow.dc.equations.DcVariableType;
 import com.powsybl.openloadflow.equations.*;
 import com.powsybl.openloadflow.lf.LoadFlowEngine;
-import com.powsybl.openloadflow.lf.outerloop.OuterLoopResult;
 import com.powsybl.openloadflow.lf.outerloop.OuterLoopStatus;
 import com.powsybl.openloadflow.network.DisabledNetwork;
 import com.powsybl.openloadflow.network.LfBus;
@@ -54,7 +53,7 @@ public class DcLoadFlowEngine implements LoadFlowEngine<DcVariableType, DcEquati
 
         private int outerLoopTotalIterations = 0;
 
-        private OuterLoopResult lastOuterLoopResult = OuterLoopResult.stable();
+        private OuterLoopStatus lastOuterLoopStatus = OuterLoopStatus.STABLE;
     }
 
     @Override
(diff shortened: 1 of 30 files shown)


src/test/java/com/powsybl/openloadflow/ac/AreaInterchangeControlTest.java
```diff
@@ -107,19 +150,38 @@ void remainingMismatchDistributeOnReferenceGenerator() {
...
+
+    private void remainingMismatchThrow() {
         parametersExt.setSlackDistributionFailureBehavior(OpenLoadFlowParameters.SlackDistributionFailureBehavior.THROW);
         Network network = MultiAreaNetworkFactory.createOneAreaBase();
         network.getGenerator("g1").setMinP(90); // the generator should go down to 70MW to meet the interchange target
         CompletionException thrown = assertThrows(CompletionException.class, () -> loadFlowRunner.run(network, parameters));
         assertEquals("Failed to distribute interchange active power mismatch", thrown.getCause().getMessage());
     }
 
+    static Stream<Arguments> slackDistributionFailureBehaviors() {
```
jeandemanged, line 176: could be simplified to
return Arrays.stream(OpenLoadFlowParameters.SlackDistributionFailureBehavior.values()).map(Arguments::of);

but seems unused - to be deleted ?

vmouradian: Yes it was unused, removed in [59fd886](https://github.com/powsybl/powsybl-open-loadflow/pull/1099/commits/59fd8860e4c54a93d7506096321ac2b0f04d5d2e)

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -162,18 +183,32 @@ public DcLoadFlowResult run() {
...
 
         initStateVector(network, equationSystem, new UniformValueVoltageInitializer());
 
-        if (parameters.isDistributedSlack()) {
-            distributeSlack(network, network.getBuses(), parameters.getBalanceType(), parameters.getNetworkParameters().isUseActiveLimits());
+        double slackBusActivePowerMismatch = getActivePowerMismatch(network.getBuses());
+        double distributedActivePower = 0.0;
+        if (parameters.isDistributedSlack() || parameters.isAreaInterchangeControl()) {
+            distributedActivePower = distributeSlack(network, network.getBuses(), parameters.getBalanceType(), parameters.getNetworkParameters().isUseActiveLimits());
         }
```
jeandemanged, line 212: It was like this before, but here we do not account for slack distribution failure behavior parameter.
Could be in another PR, in this case just add a `// FIXME`.
Important note: in the case of AIC, failing here may not be relevant. Here is done only global slack distribution. But we may have the case of distribution failure when distributing mismatch globally, and at the same time no failure when distributing per area later on in the outerloops.

jeandemanged (changes requested):
Documentation to be updated.

This PR is a quite long, we might want to break it in multiple pieces:
- DC outerloop refactoring
- AreaInterchangeControl

vidaldid-rte:
@jeandemanged  @vmouradian  shouldn't this one be closed now  that it is split in 2 parts ?

vmouradian:
> @jeandemanged @vmouradian shouldn't this one be closed now that it is split in 2 parts ?

Yes, I think we can close it now

PR #1105, opened 2024-10-18
fix: split in sub-reports for a list of elements that could be too big

TheMaskedTurtle:

Split a report where a message can be really big into smaller sub-reports

We're creating a report with a message based on a map that gathers all the network elements that are in error. This list can be really big and we feel this is not a good practice to do so.

Instead of one report gathering all the detailed information we split it into sub-reports with TRACE level and one summary message with ERROR level

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/util/Reports.java
+++ b/src/main/java/com/powsybl/openloadflow/util/Reports.java
@@ -606,14 +606,25 @@ public final class Reports {
     }
 
     public static void reportNewtonRaphsonBusesOutOfRealisticVoltageRange(ReportNode reportNode, Map<String, Double> busesOutOfRealisticVoltageRange, double minRealisticVoltage, double maxRealisticVoltage) {
-        reportNode.newReportNode()
-                .withMessageTemplate("newtonRaphsonBusesOutOfRealisticVoltageRange", "${busCountOutOfRealisticVoltageRange} buses have a voltage magnitude out of the configured realistic range [${minRealisticVoltage}, ${maxRealisticVoltage}] p.u.: ${busesOutOfRealisticVoltageRange}")
+        ReportNode voltageOutOfRangeReport = reportNode.newReportNode()
+            .withMessageTemplate("busesOutOfVoltageRealisticRange", "Buses have a voltage magnitude out of the configured realistic range")
+            .add();
+
+        voltageOutOfRangeReport.newReportNode()
+                .withMessageTemplate("busesOutOfVoltageRealisticRangeSummary", "${busCountOutOfRealisticVoltageRange} buses have a voltage magnitude out of the configured realistic range [${minRealisticVoltage}, ${maxRealisticVoltage}]")
                 .withUntypedValue("busCountOutOfRealisticVoltageRange", busesOutOfRealisticVoltageRange.size())
                 .withUntypedValue("minRealisticVoltage", minRealisticVoltage)
                 .withUntypedValue("maxRealisticVoltage", maxRealisticVoltage)
                 .withUntypedValue("busesOutOfRealisticVoltageRange", busesOutOfRealisticVoltageRange.toString())
                 .withSeverity(TypedValue.ERROR_SEVERITY)
                 .add();
+
+        busesOutOfRealisticVoltageRange.forEach((id, voltage) -> voltageOutOfRangeReport.newReportNode()
+            .withMessageTemplate("busesOutOfVoltageRealisticRangeDetail", "Bus ${id} has an unrealistic voltage magnitude: ${voltage} pu")
+            .withUntypedValue("id", id)
+            .withUntypedValue("voltage", voltage)
+            .withSeverity(TypedValue.TRACE_SEVERITY)
+            .add());
     }
 
     public static void reportAngleReferenceBusAndSlackBuses(ReportNode reportNode, String referenceBus, List<String> slackBuses) {
--- a/src/test/resources/saReport.txt
+++ b/src/test/resources/saReport.txt
@@ -37,7 +37,9 @@
             Outer loop ReactiveLimits
             AC load flow completed successfully (solverStatus=CONVERGED, outerloopStatus=STABLE)
          + Post-contingency simulation 'NGEN_NHV1'
-            1 buses have a voltage magnitude out of the configured realistic range [0.5, 1.5] p.u.: {VLHV1_0=1.5105659472379578}
+            + Buses have a voltage magnitude out of the configured realistic range
+               1 buses have a voltage magnitude out of the configured realistic range [0.5, 1.5]
+               Bus VLHV1_0 has an unrealistic voltage magnitude: 1.5105659472379578 pu
             AC load flow completed with error (solverStatus=UNREALISTIC_STATE, outerloopStatus=STABLE)
          + Post-contingency simulation 'NHV2_NLOAD'
             + Outer loop DistributedSlack


src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -606,14 +606,25 @@ public static void reportMaxVoltageChangeStateVectorScaling(ReportNode reportNod
...
+        ReportNode voltageOutOfRangeReport = reportNode.newReportNode()
+            .withMessageTemplate("busesOutOfVoltageRealisticRange", "Buses have a voltage magnitude out of the configured realistic range")
+            .add();
+
+        voltageOutOfRangeReport.newReportNode()
+                .withMessageTemplate("busesOutOfVoltageRealisticRangeSummary", "${busCountOutOfRealisticVoltageRange} buses have a voltage magnitude out of the configured realistic range [${minRealisticVoltage}, ${maxRealisticVoltage}]")
                 .withUntypedValue("busCountOutOfRealisticVoltageRange", busesOutOfRealisticVoltageRange.size())
                 .withUntypedValue("minRealisticVoltage", minRealisticVoltage)
                 .withUntypedValue("maxRealisticVoltage", maxRealisticVoltage)
                 .withUntypedValue("busesOutOfRealisticVoltageRange", busesOutOfRealisticVoltageRange.toString())
```
jeandemanged, line 618: remove this, not used anymore in message template

jeandemanged (changes requested):

src/test/resources/saReport.txt
```diff
@@ -37,7 +37,9 @@
             Outer loop ReactiveLimits
             AC load flow completed successfully (solverStatus=CONVERGED, outerloopStatus=STABLE)
          + Post-contingency simulation 'NGEN_NHV1'
-            1 buses have a voltage magnitude out of the configured realistic range [0.5, 1.5] p.u.: {VLHV1_0=1.5105659472379578}
+            + Buses have a voltage magnitude out of the configured realistic range
+               1 buses have a voltage magnitude out of the configured realistic range [0.5, 1.5]
+               Bus VLHV1_0 has an unrealistic voltage magnitude: 1.5105659472379578 pu
```
jeandemanged, line 42: wouldn't it be more concise this way ?

```suggestion
            + 1 buses have a voltage magnitude out of the configured realistic range [0.5, 1.5]
               Bus VLHV1_0 has an unrealistic voltage magnitude: 1.5105659472379578 pu
```

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/util/Reports.java
TheMaskedTurtle: You're right !

src/test/resources/saReport.txt
TheMaskedTurtle: Actually I agree but the current way of organizing reports is that the report container does not hold the severity. And I want the whole information to be appearing as an error and the detailed information appearing as trace. That's why I did it this way. I agree it makes information redundant but I think it's done like that in several places.

jeandemanged: > the report container does not hold the severity

Report containers can have severity I pushed a commit with a proposal, let me know if this works for you.

Here in JSON serialization:

```json
{
  "version" : "2.1",
  "dictionaries" : {
    "default" : {
      "acLfComplete" : "AC load flow completed ${successText} (solverStatus=${solverStatus}, outerloopStatus=${outerloopStatus})",
      "angleReferenceBusSelection" : "Angle reference bus: ${referenceBus}",
      "busesOutOfVoltageRealisticRangeDetail" : "Bus ${id} has an unrealistic voltage magnitude: ${voltage} pu",
      "busesOutOfVoltageRealisticRangeSummary" : "${busCountOutOfRealisticVoltageRange} buses have a voltage magnitude out of the configured realistic range [${minRealisticVoltage}, ${maxRealisticVoltage}]",
      "lfNetwork" : "Network CC${networkNumCc} SC${networkNumSc}",
      "loadFlow" : "Load flow on network '${networkId}'",
      "networkBalance" : "Network balance: active generation=${activeGeneration} MW, active load=${activeLoad} MW, reactive generation=${reactiveGeneration} MVar, reactive load=${reactiveLoad} MVar",
      "networkInfo" : "Network info",
      "networkSize" : "Network has ${busCount} buses and ${branchCount} branches",
      "slackBusSelection" : "Slack bus: ${slackBus}",
      "testReport" : "Test Report"
    }
  },
  "reportRoot" : {
    "messageKey" : "testReport",
    "children" : [ {
      "messageKey" : "loadFlow",
      "values" : {
        "networkId" : {
          "value" : "sim1"
        }
      },
      "children" : [ {
        "messageKey" : "lfNetwork",
        "values" : {
          "networkNumCc" : {
            "value" : 0
          },
          "networkNumSc" : {
            "value" : 0
          }
        },
        "children" : [ {
          "messageKey" : "networkInfo",
          "children" : [ {
            "messageKey" : "networkSize",
            "values" : {
              "busCount" : {
                "value" : 4
              },
              "branchCount" : {
                "value" : 4
              },
              "reportSeverity" : {
                "value" : "INFO",
                "type" : "SEVERITY"
              }
            }
          }, {
            "messageKey" : "networkBalance",
            "values" : {
              "activeGeneration" : {
                "value" : 607.0
              },
              "activeLoad" : {
                "value" : 600.0
              },
              "reactiveGeneration" : {
                "value" : 0.0
              },
              "reactiveLoad" : {
                "value" : 200.0
              },
              "reportSeverity" : {
                "value" : "INFO",
                "type" : "SEVERITY"
              }
            }
          }, {
            "messageKey" : "angleReferenceBusSelection",
            "values" : {
              "referenceBus" : {
                "value" : "VLHV1_0"
              },
              "reportSeverity" : {
                "value" : "INFO",
                "type" : "SEVERITY"
              }
            }
          }, {
            "messageKey" : "slackBusSelection",
            "values" : {
              "slackBus" : {
                "value" : "VLHV1_0"
              },
              "reportSeverity" : {
                "value" : "INFO",
                "type" : "SEVERITY"
              }
            }
          } ]
        }, {
          "messageKey" : "busesOutOfVoltageRealisticRangeSummary",
          "values" : {
            "busCountOutOfRealisticVoltageRange" : {
              "value" : 4
            },
            "minRealisticVoltage" : {
              "value" : 0.99
            },
            "maxRealisticVoltage" : {
              "value" : 1.01
            },
            "reportSeverity" : {
              "value" : "ERROR",
              "type" : "SEVERITY"
            }
          },
          "children" : [ {
            "messageKey" : "busesOutOfVoltageRealisticRangeDetail",
            "values" : {
              "id" : {
                "value" : "VLGEN_0"
              },
              "voltage" : {
                "value" : 1.0208333333333333
              },
              "reportSeverity" : {
                "value" : "TRACE",
                "type" : "SEVERITY"
              }
            }
          }, {
            "messageKey" : "busesOutOfVoltageRealisticRangeDetail",
            "values" : {
              "id" : {
                "value" : "VLHV1_0"
              },
              "voltage" : {
                "value" : 1.0582636574158686
              },
              "reportSeverity" : {
                "value" : "TRACE",
                "type" : "SEVERITY"
              }
            }
          }, {
            "messageKey" : "busesOutOfVoltageRealisticRangeDetail",
            "values" : {
              "id" : {
                "value" : "VLHV2_0"
              },
              "voltage" : {
                "value" : 1.0261840057810543
              },
              "reportSeverity" : {
                "value" : "TRACE",
                "type" : "SEVERITY"
              }
            }
          }, {
            "messageKey" : "busesOutOfVoltageRealisticRangeDetail",
            "values" : {
              "id" : {
                "value" : "VLLOAD_0"
              },
              "voltage" : {
                "value" : 0.9838500227734096
              },
              "reportSeverity" : {
                "value" : "TRACE",
                "type" : "SEVERITY"
              }
            }
          } ]
        }, {
          "messageKey" : "acLfComplete",
          "values" : {
            "successText" : {
              "value" : "with error"
            },
            "solverStatus" : {
              "value" : "UNREALISTIC_STATE"
            },
            "outerloopStatus" : {
              "value" : "STABLE"
            },
            "reportSeverity" : {
              "value" : "ERROR",
              "type" : "SEVERITY"
            }
          }
        } ]
      } ]
    } ]
  }
}
```

TheMaskedTurtle: I mean I know it can have a severity but this is not how it is done in most cases currently. And in GridSuite, this case is not well represented in the frontend. This is due to how new report model was migrated, we kept the same structure as before (with no severity on containers). I agree that your solution makes a better use of the current API but it's not aligned with the rest in terms of API usage. But actually that's an important topic we could discuss about it in next PowSyBl workshop. 
For now, I think it would be better to stick with the current logic, what do you think ?

geofjamg: This logs does not only appear because or remote voltage issue. It has been added when we didn't had any remote voltages in our data, just to avoid LU crash in stop the resolution earlier in case of any bad case.

vidaldid-rte: After closer look at the PR:
- I do not have a strong opinion on report architecture at this time. 
- On the topic of irrealistic voltage
    - I expect the isse to disapper after https://github.com/powsybl/powsybl-open-loadflow/issues/1124 is implemeted and https://github.com/powsybl/powsybl-open-loadflow/pull/1087 is merged.
    - If ever you want to proceed with this PR, I think that the list of busses with irrealistic voltage is a useful information and should be displayed  so that the user knows what to fix (usually change their target V of a remote voltage control or move them to local voltage control). So if possible the summary should include the list of affected busses. And you could leave at TRACE level the individual events in the NR.

vidaldid-rte (approved):
I don't think that TRACE is the right severity level for the message displaying the bus names, since users need to know them in order to fix the problem as explained in a previous comment.
But if no application developer is shouting I have no reason to block at this time. Hence my approval.

src/test/resources/saReport.txt
TheMaskedTurtle: @jeandemanged after discussion with powsybl team, it is ok to merge with your proposition :)

jeandemanged: Hi @TheMaskedTurtle , thanks for the update, ok great :)

jeandemanged (approved):

PR #1107, opened 2024-10-21
Refactor to allow external AC solvers

jeandemanged:

No

Feature

We can only choose between the Newton-Raphson and Newton-Krylov AC solvers provided by powsybl-open-loadflow.

It is possible to provide additional AC Solvers externally via a plugin mechanism.

AC Solvers must provide the following:
- An implementation of `com.powsybl.openloadflow.ac.solver.AcSolver` for the solver itself
- An implementation of `com.powsybl.openloadflow.ac.solver.AcSolverParameters` for the solver parameters
- An implementation of `com.powsybl.openloadflow.ac.solver.AcSolverFactory` responsible for the solver and solver parameters construction, correctly configured to be found by `java.util.ServiceLoader` (typically using `com.google.auto.service.AutoService` in PowSyBl).

The existing Newton-Raphson and Newton-Krylov AC solvers provided by powsybl-open-loadflow are now also using this mechanism.

OpenLoadFlow specific parameter `acSolverType` allows to choose the `AcSolverFactory` to be used, which will be the one selected via `AcSolverFactory#getName()`

To ease the tests for external solvers, OLF now also publishes a [`test-jar`](https://maven.apache.org/plugins/maven-jar-plugin/test-jar-mojo.html) with test networks and test utilities, which can be used as follows in Maven:

```xml
        <dependency>
            <groupId>com.powsybl</groupId>
            <artifactId>powsybl-open-loadflow</artifactId>
            <version>${powsybl-open-loadflow.version}</version>
            <classifier>tests</classifier>
            <scope>test</scope>
        </dependency>
```

- [x] No

**Other information**:

Use case: e.g. integrate and use non-linear solvers such as open-source solver Ipopt or commercial solver Knitro.


src/main/java/com/powsybl/openloadflow/ac/AcLoadFlowParameters.java
```diff
@@ -30,9 +30,7 @@ public class AcLoadFlowParameters extends AbstractLoadFlowParameters<AcLoadFlowP
 
     private AcEquationSystemCreationParameters equationSystemCreationParameters = new AcEquationSystemCreationParameters();
 
-    private NewtonRaphsonParameters newtonRaphsonParameters = new NewtonRaphsonParameters();
-
-    private NewtonKrylovParameters newtonKrylovParameters = new NewtonKrylovParameters();
+    private AcSolverParameters acSolverParameters = new NewtonRaphsonParameters();
```
jeanneArchambault, line 33: By default, we have NR parameters for Ac Solver Parameters?

jeanneArchambault:
@jeandemanged FYI, when we set a solver factory, it doesn't set the corresponding parameters :
image
When running through context (for example here in `LfBranchDisconnectionTest`), we need to set both the factory and the corresponding parameters :
image

src/main/java/com/powsybl/openloadflow/ac/AcLoadFlowParameters.java
jeandemanged: Indeed, would this be a problem ? Why would NR need NK parameters and vice-versa ?

jeanneArchambault: Yes It makes sense, I was making sure I understood the principle

jeandemanged:
> Right now, I still need to set in my TU both the factory and the parameter

@jeanneArchambault hopefully better with f3184636d97d4eef9b364fa1a97e5814929e64e1 ?

jeanneArchambault:
> yes, perfect, thank you @jeandemanged !

src/main/java/com/powsybl/openloadflow/ac/solver/AcSolverParameters.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2023, RTE (http://www.rte-france.com)
+ * Copyright (c) 2024, Artelys (http://www.artelys.com)
  * This Source Code Form is subject to the terms of the Mozilla Public
```
vidaldid-rte, line 3: Isn't there a way to add a copyright instead of replacing it ?
Same for author below.

vidaldid-rte (approved):

PR #1108, opened 2024-10-22
Retry LU decomposition if incremental mode fails

vidaldid-rte:

No

Bug Fix

In some rare cases, the NR fails with a MatrixException - klu_refactor error KLU_SINGULAR
  This happens when the Jacobian Matrix uptes the LU in incremental mode.  
  But, in those cases, if the LU matrix is updated in non incremental mode the operation succeeds.

At this time, this has only be seen in large model and it has not been possible to isolate a simple case with the issue. 

When the Jacobian updates occurs in incremental mode, if the LU update fails a second try is attempted in non incremental mode.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/equations/LUDecompositionTest.java
+++ b/src/test/java/com/powsybl/openloadflow/equations/LUDecompositionTest.java
@@ -1,0 +1,121 @@
+package com.powsybl.openloadflow.equations;
+
+import com.powsybl.iidm.network.test.EurostagTutorialExample1Factory;
+import com.powsybl.math.matrix.DenseMatrix;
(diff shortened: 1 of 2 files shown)


src/test/java/com/powsybl/openloadflow/equations/LUDecompositionTest.java
```diff
@@ -0,0 +1,132 @@
+package com.powsybl.openloadflow.equations;
```
geofjamg, line 1: Could you add the licence header and authorship ?

vidaldid-rte: Done

geofjamg (approved):

PR #1109, opened 2024-10-29
Fix high memory usage with DC Sensitivity Analysis and DC Woodbury Security Analysis

SylvestreSakti:

Last release of OLF (1.13.0) introduced a memory usage issue : DC Sensitivity Analysis is consuming too much memory (around twice more than previous release with some user tests)

DC Sensitivity is consuming around twice more memory with the woodbury engine extraction from sensitivity analysis. This memory consumption is due to creation of new DenseMatrix objects in the new implementation allocating more memory space.

A small fix has been introduced to this implementation by replacing new DenseMatrix objects allocation by shallow copies of already used DenseMatrix objects.  The WoodburyEngine class has been modified to directly modify the contingency state matrix given as input, instead of creating new matrix. 

This fix reduces the memory usage to the same level as the previous release.
The same fix has been also applied for DC Security Analysis.

- [ ] Yes
- [x] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,12 +93,12 @@ dependencies to respectively have access to network model, IEEE test networks an
 <dependency>
     <groupId>com.powsybl</groupId>
     <artifactId>powsybl-iidm-impl</artifactId>
-    <version>6.5.0</version>
+    <version>6.5.1</version>
 </dependency>
 <dependency>
     <groupId>com.powsybl</groupId>
     <artifactId>powsybl-ieee-cdf-converter</artifactId>
-    <version>6.5.0</version>
+    <version>6.5.1</version>
 </dependency>
 <dependency>
     <groupId>org.slf4j</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -54,7 +54,7 @@
         <slf4jtoys.version>1.6.3</slf4jtoys.version>
         <asciitable.version>0.3.2</asciitable.version>
 
-        <powsybl-core.version>6.5.0</powsybl-core.version>
+        <powsybl-core.version>6.5.1</powsybl-core.version>
     </properties>
 
     <build>
--- a/src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
@@ -90,9 +90,9 @@ public class AcloadFlowEngine implements LoadFlowEngine<AcVariableType, AcEquati
             if (outerLoopResult.status() == OuterLoopStatus.UNSTABLE) {
                 LOGGER.debug("Start outer loop '{}' iteration {}", outerLoop.getName(), runningContext.outerLoopTotalIterations);
 
-                ReportNode nrReportNode = context.getNetwork().getReportNode();
+                ReportNode reportNode = context.getNetwork().getReportNode();
                 if (context.getParameters().isDetailedReport()) {
-                    nrReportNode = Reports.createDetailedSolverReporterOuterLoop(nrReportNode,
+                    reportNode = Reports.createDetailedSolverReporterOuterLoop(reportNode,
                             solver.getName(),
                             context.getNetwork().getNumCC(),
                             context.getNetwork().getNumSC(),
@@ -101,7 +101,7 @@ public class AcloadFlowEngine implements LoadFlowEngine<AcVariableType, AcEquati
                 }
 
                 // if not yet stable, restart solver
-                runningContext.lastSolverResult = solver.run(new PreviousValueVoltageInitializer(), nrReportNode);
+                runningContext.lastSolverResult = solver.run(new PreviousValueVoltageInitializer(), reportNode);
 
                 runningContext.nrTotalIterations.add(runningContext.lastSolverResult.getIterations());
                 runningContext.outerLoopTotalIterations++;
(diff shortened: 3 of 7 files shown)


Hadrien-Godard (approved):

p-arvy (approved):

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -464,23 +465,45 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
...
         }
     }
+
+    /**
+     * Copy all the values that are in an originalMatrix and paste it in the copyMatrix (without allocating new memory spaces)
+     * The dimensions of both matrices must be the same
+     */
+    // TODO : implement this method for DenseMatrix in powsybl-core ?
+    private static void matrixShallowCopy(DenseMatrix originalMatrix, DenseMatrix copyMatrix) {
+        if (originalMatrix.getRowCount() == copyMatrix.getRowCount() && originalMatrix.getColumnCount() == copyMatrix.getColumnCount()) {
```
vidaldid-rte, line 500: The term shallow copy is incorrect. See definition here https://developer.mozilla.org/en-US/docs/Glossary/Shallow_copy
In a shallow copy in a shallow copy when you change either the source or the copy, you may also cause the other object to change too.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -438,13 +438,14 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
...
+                DenseMatrix baseFlowStates = calculateFlowStates(loadFlowContext, participatingElements, new DisabledNetwork(), reportNode);
+                DenseMatrix flowStates = (DenseMatrix) baseFlowStates.copy(new DenseMatrixFactory());
                 // compute the pre-contingency factor states
-                DenseMatrix factorsStates = calculateFactorStates(loadFlowContext, factorGroups, participatingElements);
+                DenseMatrix baseFactorStates = calculateFactorStates(loadFlowContext, factorGroups, participatingElements);
+                DenseMatrix factorStates = (DenseMatrix) baseFactorStates.copy(new DenseMatrixFactory());
 
                 // calculate sensitivity values for pre-contingency network
-                calculateSensitivityValues(validFactorHolder.getFactorsForBaseNetwork(), factorsStates, flowStates, null, resultWriter, new DisabledNetwork());
+                calculateSensitivityValues(validFactorHolder.getFactorsForBaseNetwork(), factorStates, flowStates, null, resultWriter, new DisabledNetwork());
```
vidaldid-rte, line 449: Why don't you use baseFactorStates here ? Isn't the function read-only ?

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
SylvestreSakti: Should we speak about "deep copy" ? (but without creating a new container)

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -438,13 +438,14 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
                         : Collections.emptyList();
 
                 // run DC load on pre-contingency network
-                DenseMatrix flowStates = calculateFlowStates(loadFlowContext, participatingElements, new DisabledNetwork(), reportNode);
-
+                DenseMatrix baseFlowStates = calculateFlowStates(loadFlowContext, participatingElements, new DisabledNetwork(), reportNode);
+                DenseMatrix flowStates = (DenseMatrix) baseFlowStates.copy(new DenseMatrixFactory());
                 // compute the pre-contingency factor states
```
vidaldid-rte, line 443: It seems useless to copu the values here since they will be restored to the base value at each iteration. You could just create it and avoid here the iteration on all non null values (can be important on large matrixes).

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
SylvestreSakti: Yes `calculateSensisitivityValues` (as well as `createBranchSensitivityValue`) is read-only. At this line, it can take baseFactorSates and baseFlowStates indeed.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
vidaldid-rte: It would also be clearer to rename the variable flowStatesWorkingMatrix to make it clear that is is a reused matrix that is modified at each iteration.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
vidaldid-rte: copyValues light be a bettre term and make clear you expect the same matrix dimensions.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -464,23 +465,45 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
...
+    // TODO : implement this method for DenseMatrix in powsybl-core ?
+    private static void matrixShallowCopy(DenseMatrix originalMatrix, DenseMatrix copyMatrix) {
+        if (originalMatrix.getRowCount() == copyMatrix.getRowCount() && originalMatrix.getColumnCount() == copyMatrix.getColumnCount()) {
+            for (int columnIndex = 0; columnIndex < originalMatrix.getColumnCount(); columnIndex++) {
+                for (int rowIndex = 0; rowIndex < originalMatrix.getRowCount(); rowIndex++) {
+                    copyMatrix.set(rowIndex, columnIndex, originalMatrix.get(rowIndex, columnIndex));
+                }
+            }
+        }
+    }
```
vidaldid-rte, line 507: You should throw an exception (IllegalArgumentException) if the matri sizes do not match instead of silently ignoring the error.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -231,7 +232,6 @@ private void calculateSensitivityValuesForAContingency(DcLoadFlowContext loadFlo
             DenseMatrix postContingencyFlowStates = engine.run(newFlowStates);
             DenseMatrix postContingencyFactorStates = engine.run(newFactorStates);
             calculateSensitivityValues(factors, postContingencyFactorStates, postContingencyFlowStates, contingency, resultWriter, disabledNetwork);
```
vidaldid-rte, line 234: Since nexFlowStates and newFactorSates are modified by run, and since they can be the input values factorSate and flowState the javadoc of calculateSensitivityValuesForAContingency should explicit that the matrices may be modified.

src/main/java/com/powsybl/openloadflow/sensi/WoodburyEngine.java
```diff
@@ -84,12 +84,12 @@ private void setAlphas(DenseMatrix states, int columnState) {
 
     /**
      * Calculate post-contingency states values using pre-contingency states values and some flow transfer factors (alphas).
-     * @return a matrix of post-contingency voltage angle states.
+     * @return a matrix of post-contingency voltage angle states (which is the same modified pre-contingency states java object).
      */
     public DenseMatrix run(DenseMatrix preContingencyStates) {
         Objects.requireNonNull(preContingencyStates);
```
vidaldid-rte, line 90: The API xould be clearer with something like
public void toPostContingencyStates(DenseMatrix preContingencyStates);

It would clarify that the input matrix is modified.

SylvestreSakti:
Thank you for the review @vidaldid-rte 
All the remarks have been taken into account and can be reviewed in this PR for the next release : https://github.com/powsybl/powsybl-open-loadflow/pull/1111

PR #1111, opened 2024-11-04
Refactor DC Woodbury

SylvestreSakti:

No

This PR follows this PR: https://github.com/powsybl/powsybl-open-loadflow/pull/1109
Some variable names and comments have been modified to ensure a better maintanability of the code

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
@@ -76,6 +76,7 @@ public class WoodburyDcSecurityAnalysis extends DcSecurityAnalysis {
      * In case of connectivity break, a pre-computation has been done in {@link #calculatePostContingencyStatesForAContingencyBreakingConnectivity}
      * to reset active power flow of hvdc lines on which one bus is lost.
      * If connectivity, a generator, a load or a phase tap changer is lost due to the contingency, the pre contingency flowStates are overridden.
+     * @return the post contingency states for the contingency
      */
     private double[] calculatePostContingencyStatesForAContingency(DcLoadFlowContext loadFlowContext, DenseMatrix contingenciesStates, double[] flowStates,
                                                                    PropagatedContingency contingency, Map<String, ComputedContingencyElement> contingencyElementByBranch,
@@ -92,7 +93,6 @@ public class WoodburyDcSecurityAnalysis extends DcSecurityAnalysis {
         DisabledNetwork disabledNetwork = new DisabledNetwork(disabledBuses, disabledBranches);
 
         WoodburyEngine engine = new WoodburyEngine(loadFlowContext.getParameters().getEquationSystemCreationParameters(), contingencyElements, contingenciesStates);
-        double[] postContingencyStates;
         double[] newFlowStates = flowStates;
         if (contingency.getGeneratorIdsToLose().isEmpty() && contingency.getLoadIdsToLose().isEmpty()) {
 
@@ -108,7 +108,7 @@ public class WoodburyDcSecurityAnalysis extends DcSecurityAnalysis {
             if (!disabledBuses.isEmpty() || !lostPhaseControllers.isEmpty()) {
                 newFlowStates = DcLoadFlowEngine.run(loadFlowContext, disabledNetwork, reportNode);
             }
-            postContingencyStates = engine.run(newFlowStates);
+            engine.computeToPostContingencyStates(newFlowStates);
         } else {
             // if we have a contingency including the loss of a DC line or a generator or a load
             // save base state for later restoration after each contingency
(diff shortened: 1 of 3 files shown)


src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -496,13 +499,15 @@ public void analyse(Network network, List<PropagatedContingency> contingencies,
...
+    private static void matrixCopyValues(DenseMatrix originalMatrix, DenseMatrix copyMatrix) {
         if (originalMatrix.getRowCount() == copyMatrix.getRowCount() && originalMatrix.getColumnCount() == copyMatrix.getColumnCount()) {
             for (int columnIndex = 0; columnIndex < originalMatrix.getColumnCount(); columnIndex++) {
                 for (int rowIndex = 0; rowIndex < originalMatrix.getRowCount(); rowIndex++) {
                     copyMatrix.set(rowIndex, columnIndex, originalMatrix.get(rowIndex, columnIndex));
                 }
             }
+        } else {
+            throw new IllegalArgumentException();
         }
```
vidaldid-rte, line 511: Maybe a message in the exception. For example. Invalid matrix size for copyMatrix. x columns and r rows expected. Received x' columns and r' rows.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
```diff
@@ -108,7 +108,7 @@ private double[] calculatePostContingencyStatesForAContingency(DcLoadFlowContext
             if (!disabledBuses.isEmpty() || !lostPhaseControllers.isEmpty()) {
                 newFlowStates = DcLoadFlowEngine.run(loadFlowContext, disabledNetwork, reportNode);
             }
-            postContingencyStates = engine.run(newFlowStates);
+            engine.computeToPostContingencyStates(newFlowStates);
         } else {
```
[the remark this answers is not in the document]
vidaldid-rte: In this file(WoodburyDCSecurityAnalysis), line 285, can you rename baseContingencyStates to workingContingencyStates to highlight that the array is reused for each contingency ?

vidaldid-rte:
There are still type of contingenies in DcSensitivyAnalysis where a new DenseMatrix is created  (For example in caculateFactorStates. This can lead to the same memory issues in cases where many contingency fall in such category.

It would be good to have a resizableDenseMatrix that can grow its directBuffer if needed when resized, and could be used as a working matrix for all contingencies. 
In the same idea, for WoodburyDCSecurityAnalysis, we could have a large reusable array of double that could grow when needed.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
SylvestreSakti: Yes, I modified this exception to a MatrixException (same check is done and same exception is used when summing two matrices in powybl-core)

SylvestreSakti:
> There are still type of contingenies in DcSensitivyAnalysis where a new DenseMatrix is created (For example in caculateFactorStates. This can lead to the same memory issues in cases where many contingency fall in such category.
> 
> It would be good to have a resizableDenseMatrix that can grow its directBuffer if needed when resized, and could be used as a working matrix for all contingencies. In the same idea, for WoodburyDCSecurityAnalysis, we could have a large reusable array of double that could grow when needed.
> 
> But this is maybe out of the scope of this PR since it needs an evolution i core.

Yes, I agree with that, a "ResizableDenseMatrix" would help using less memory. But indeed may be out of this scope (and would need a modification in powsybl-core)

vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
```diff
@@ -201,8 +201,10 @@ protected SecurityAnalysisResult runSimulations(LfNetwork lfNetwork, List<Propag
             // this is a difference with dc security analysis
             cleanContingencies(lfNetwork, propagatedContingencies);
 
+            // compute the pre-contingency states
             double[] preContingencyStates = DcLoadFlowEngine.run(context, new DisabledNetwork(), reportNode);
-            double[] baseContingencyStates = preContingencyStates.clone();
+            // create workingContingencyStates that will be a working copy of pre-contingency states
+            double[] workingContingencyStates = preContingencyStates.clone();
```
jeandemanged, line 207: Use `System.arraycopy` instead of clone ?

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
SylvestreSakti: Done ! I did not know about this subtle difference but seems slightly better indeed (https://stackoverflow.com/questions/7179251/is-there-any-reason-to-prefer-system-arraycopy-over-clone)

vidaldid-rte: That was true for Java5 in 2006 but if you run the bench today you see no difference after the JIT has comleted.

Hadrien-Godard (approved):

jeandemanged (approved):

PR #1113, opened 2024-11-05
bump parent pom to version 20

vidaldid-rte:

Upgrade pom parent version.

Updates build configuration.

No change expected. Just keep up to date with powsybl build chain and configuration.

- [ ] Yes
- [X] No

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
     <parent>
         <groupId>com.powsybl</groupId>
         <artifactId>powsybl-parent</artifactId>
-        <version>18</version>
+        <version>20</version>
         <relativePath/>
     </parent>
 


jeandemanged:
we should probably do the same on powsybl-core.
I guess it was not done because the changes are related to liquibase/hibernate and springboot which are not used in core nor olf.

@olperr1 your advice ?

vidaldid-rte:
> we should probably do the same on powsybl-core. I guess it was not done because the changes [v18..v20](https://github.com/powsybl/powsybl-parent/compare/v18...v20) are related to liquibase/hibernate and springboot which are not used in core nor olf. Although we can see in the diff a change in git commit id plugin...
> 
> @olperr1 your advice ?

@jeandemanged  FYI Olivier asked al projects to do this in our internal powsybl meeting this monday, if it was possible. It is changing the build plugin versions, nothing visible at runtime. But if helps us stay current. Also good to do this early in a release cycle.

olperr1:
@jeandemanged: Yes, I will open the PR on powsybl-core.

jeandemanged (approved):

PR #1115, opened 2024-11-06
[WIP] Automatic fix of incompatible target voltages

geofjamg:

- [ ] Yes
- [ ] No

**Other information**:


src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -276,6 +276,8 @@ public enum FictitiousGeneratorVoltageControlCheckMode {
 
     public static final String AREA_INTERCHANGE_P_MAX_MISMATCH_PARAM_NAME = "areaInterchangePMaxMismatch";
 
+    public static final String FIX_TARGET_VOLTAGE_INCOMPATIBILITY_PARAM_NAME = "fixTargetVoltageIncompatibility";
+
```
vidaldid-rte, line 280: To be documented.
And maybe addiotional parameters given like TARGET_VOLTAGE_PLAUSIBILITY_THRESHOLD

However in my test networks, increasing CONTROLLED_BUS_NEIGHBORS_EXPLORATION_DEPTH had a terrible algotirhmic cost and didn't improve quality (defined as more convergence)

src/main/java/com/powsybl/openloadflow/TargetVoltageCompatibilityChecker.java
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ */
+public class TargetVoltageCompatibilityChecker {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(TargetVoltageCompatibilityChecker.class);
+
+    private final LfNetwork network;
+
```
vidaldid-rte, line 36: Would it make sense to extend the API so that it can also be used with a Network (not an LfNetwork) and returns the IDM view of IIDM incompatibleControlled Buses - for example pairs of list of controllers  that are incomaptible) ?

src/main/java/com/powsybl/openloadflow/adm/AdmittanceMatrix.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public AdmittanceMatrix(EquationSystem<AdmittanceVariableType, AdmittanceEquationType> equationSystem, Matrix matrix) {
+        this.equationSystem = Objects.requireNonNull(equationSystem);
+        this.matrix = Objects.requireNonNull(matrix);
+    }
+
+    public static AdmittanceMatrix create(AdmittanceEquationSystem ySystem, MatrixFactory matrixFactory) {
+        Objects.requireNonNull(matrixFactory);
+        try (var j = new JacobianMatrix<>(ySystem.getEquationSystem(), matrixFactory)) {
+            return new AdmittanceMatrix(ySystem.getEquationSystem(), j.getMatrix());
```
vidaldid-rte, line 42: Wouldn't it be cleaner to inherit from JacobianMatrix ?
Here if a future evolution of JacobianMatrix closes the Matrix object the could will fail because it assumes today that the matrix is still alive.

src/main/java/com/powsybl/openloadflow/adm/AdmittanceMatrix.java
```diff
@@ -0,0 +1,94 @@
...
+     * Get equivalent impedance between 2 buses.
+     */
+    public Complex getZ(LfBus bus1, LfBus bus2) {
+        Objects.requireNonNull(bus1);
+        Objects.requireNonNull(bus2);
+        var i1xEq = equationSystem.getEquation(bus1.getNum(), AdmittanceEquationType.BUS_ADM_IX).orElseThrow();
+        var i1yEq = equationSystem.getEquation(bus1.getNum(), AdmittanceEquationType.BUS_ADM_IY).orElseThrow();
+        var i2xEq = equationSystem.getEquation(bus2.getNum(), AdmittanceEquationType.BUS_ADM_IX).orElseThrow();
+        var i2yEq = equationSystem.getEquation(bus2.getNum(), AdmittanceEquationType.BUS_ADM_IY).orElseThrow();
+        if (e == null) {
```
vidaldid-rte, line 60: This is definitely not MT safe.
Can you document the limitation in the class ?

src/main/java/com/powsybl/openloadflow/TargetVoltageCompatibilityChecker.java
```diff
@@ -0,0 +1,139 @@
...
+                .filter(bus -> !bus.isDisabled()
+                        && bus.isVoltageControlled()
+                        && bus.getVoltageControls().stream().anyMatch(vc -> !vc.isDisabled()))
+                .collect(Collectors.toSet());
+
+        List<Pair<LfBus, LfBus>> incompatibleControlledBuses = new ArrayList<>();
+
+        var ySystem = AdmittanceEquationSystem.create(network, new VariableSet<>());
+        try (var y = AdmittanceMatrix.create(ySystem, parameters.getMatrixFactory())) {
+            for (LfBus controlledBus : controlledBuses) {
```
vidaldid-rte, line 90: Don't you compute the impedance between each pair of bus twice ? if you iterate on all buses then on all neighbours ?

Also, would it accerlerate things if there was an option to check a pair only if at least one of the two buses is remotely controlled ? [This is where we have convergence issues]

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
geofjamg: "And maybe addiotional parameters given like TARGET_VOLTAGE_PLAUSIBILITY_THRESHOLD" => yes
for CONTROLLED_BUS_NEIGHBORS_EXPLORATION_DEPTH indeed it should not be configurable, a value of 2 or 3 is probably enough.

src/main/java/com/powsybl/openloadflow/adm/AdmittanceMatrix.java
geofjamg: Admittance matrix has nothing to do with Jacobian matrix even if we use it to compute admittance.

src/main/java/com/powsybl/openloadflow/adm/AdmittanceMatrix.java
geofjamg: Indeed but I don't think it is expected to be thread safe as all of other similar OLF linear algebra utilities.

src/main/java/com/powsybl/openloadflow/TargetVoltageCompatibilityChecker.java
geofjamg: Yes, done. 
```java
Network network = ...;
LoadFlowParameters parameters = new LoadFlowParameters();
var elementIds = RemoteVoltageTargetChecker.findElementsToDiscardFromVoltageControl(network, parameters);
System.out.println(elementIds);
```

src/main/java/com/powsybl/openloadflow/adm/AdmittanceEquationSystem.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private static void createBranchEquations(Collection<LfBranch> branches, VariableSet<AdmittanceVariableType> variableSet, EquationSystem<AdmittanceVariableType, AdmittanceEquationType> equationSystem) {
+        for (LfBranch branch : branches) {
+            LfBus bus1 = branch.getBus1();
+            LfBus bus2 = branch.getBus2();
+            PiModel piModel = branch.getPiModel();
+            if (FastMath.abs(piModel.getX()) < LfNetworkParameters.LOW_IMPEDANCE_THRESHOLD_DEFAULT_VALUE) {
+                if (bus1 != null && bus2 != null) {
+                    LOGGER.warn("Non impedant branches ({}) not supported in admittance matrix",
```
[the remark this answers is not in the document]
geofjamg: Fixed, I cut the impedance using the `LfNetworkParameters.LOW_IMPEDANCE_THRESHOLD_DEFAULT_VALUE` which is ok for what we are doing with admittance matrix.

src/main/java/com/powsybl/openloadflow/TargetVoltageCompatibilityChecker.java
geofjamg: Yes, we are checking potentially twice. It could be improved indeed.

geofjamg:
@vidaldid-rte @SylvestreSakti I added in addition to "incompatible target v" check a new "unrealistic target v" which rely on target voltage to calculate voltage sensitivities (already existing in OLF) to estimate to controller bus voltage given a controlled bus targe voltage and discard when too far from 1pu. On my real test case it works very well.

src/main/java/com/powsybl/openloadflow/RemoteVoltageTargetChecker.java
```diff
@@ -0,0 +1,270 @@
...
+        this.network = Objects.requireNonNull(network);
+        this.equationSystem = Objects.requireNonNull(equationSystem);
+        this.j = Objects.requireNonNull(j);
+    }
+
+    public static Set<String> findElementsToDiscardFromVoltageControl(Network network, LoadFlowParameters parameters) {
+        return findElementsToDiscardFromVoltageControl(network, parameters, new SparseMatrixFactory());
+    }
+
+    public static Set<String> findElementsToDiscardFromVoltageControl(Network network, LoadFlowParameters parameters, MatrixFactory matrixFactory) {
```
vidaldid-rte, line 68: Is it possible to add a reason ?
   Incompatible with XXX
   Irrealistic

(It seems at this time there is even no log to understand the discard recommendation)

src/main/java/com/powsybl/openloadflow/RemoteVoltageTargetChecker.java
```diff
@@ -0,0 +1,270 @@
...
+                break;
+            }
+            if (b != bus) {
+                neighbors.add(b);
+            }
+        }
+        return neighbors;
+    }
+
+    private void checkIncompatibleTargets(RemoteVoltageTargetCheckerParameters parameters,
```
vidaldid-rte, line 142: Very good idea !

src/main/java/com/powsybl/openloadflow/TargetVoltageCompatibilityChecker.java
vidaldid-rte: Thanks !
Closing that one. (but I added an ohter commet on the method suggesting to add an explanation)

src/main/java/com/powsybl/openloadflow/adm/AdmittanceMatrix.java
vidaldid-rte: Well - you reused the matrix factory part. And steal the matrix of a closable - in addition a matrix with a DirectBuffer in it. That works today, and in your code structure but is not really a reusable object. Not a big deal and not reason to disapprove. But I can't stay silent when I see this :-)

src/main/java/com/powsybl/openloadflow/adm/AdmittanceMatrix.java
vidaldid-rte: If I didn-t know the underlyings I would expect it to be threadsafe. Maybe a general problem to OLF. 
But running it in MT mode requires a lot of expertis today (see previous PR an MT security analysis).
Anyway, out of scope of ths feature.

src/main/java/com/powsybl/openloadflow/RemoteVoltageTargetChecker.java
vidaldid-rte: Corrected

src/main/java/com/powsybl/openloadflow/TargetVoltageCompatibilityChecker.java
vidaldid-rte: Note that the formula does not give the same result for b1, b2 than for b2,b1. See log extract from the new unit test.

Indicator: vl4_1/vl4_2 dv = 0.010000000000000009 dz = 1.8319485389994188E-5 indicator = 545.8668618203571
Indicator: vl4_2/vl4_1 dv = 0.010000000000000009 dz = 2.0915710639439877E-5 indicator = 478.1095021052466

The small dz has also different values on Mac OS (looks like numerical errors end up in large differences for the indicator when impedance distance is small)

vidaldid-rte (dismissed):

docs/loadflow/parameters.md
```diff
@@ -596,6 +596,17 @@ When set to `false`:
...
+If true, runs a preprocessing algorithm to identify remote voltage settings that may cause convergence issues and 
+removes them automatically from voltage control. Two voltage targets are considered incompatible, when they are separated 
+by small impedances, and have quite different voltage targets. In some situtations, this could lead to a failure in the Newton-Rapshon convergence.
+
+**Note** : In situations where the network configuration can be modified (for example when preparing a forecast network configuration) a preferred 
+solution is to identify problematic voltage settings using the RemoteVoltageTargetChecker.findElementsToDiscardFromVoltageControl
+that can be run on a Network. This java method returns the problematic voltage settings and the reason. Users can then use this result
+to fix the network data.
+
+The default value is 'false'
```
jeandemanged: backticks and dot at the end

```suggestion
The default value is `false`.
```

src/main/java/com/powsybl/openloadflow/ac/AcLoadFlowParameters.java
```diff
@@ -56,6 +58,8 @@ public class AcLoadFlowParameters extends AbstractLoadFlowParameters<AcLoadFlowP
 
     private double minNominalVoltageRealisticVoltageCheck = DEFAULT_MIN_NOMINAL_VOLTAGE_REALISTIC_VOLTAGE_CHECK;
 
+    private boolean fixRemoteVoltageTarget = FIX_REMOTE_VOLTAGE_TARGET_DEFAULT_VALUE;
```
jeandemanged:
```suggestion
    private boolean fixRemoteVoltageTarget = FIX_REMOTE_VOLTAGE_TARGETS_DEFAULT_VALUE;
```

src/main/java/com/powsybl/openloadflow/network/AbstractElement.java
```diff
@@ -71,6 +71,11 @@ public void removeEvaluable(Evaluable evaluable) {
         // nothing by default
     }
 
+    @Override
+    public void setVoltageControlEnabled(boolean enabled) {
+        // nothing byt default
```
jeandemanged:
```suggestion
        // nothing by default
```

jeandemanged (changes requested):
looks great to me, only cosmetic comments

docs/loadflow/parameters.md
```diff
@@ -596,6 +596,17 @@ When set to `false`:
 
 The default value is `true`.
 
+**fixRemoteVoltageTarget**
+If true, runs a preprocessing algorithm to identify remote voltage settings that may cause convergence issues and 
+removes them automatically from voltage control. Two voltage targets are considered incompatible, when they are separated 
+by small impedances, and have quite different voltage targets. In some situtations, this could lead to a failure in the Newton-Rapshon convergence.
```
SylvestreSakti: Without getting too much into details, some explaination could be given on the used indicator (that is also displayed in the report)
 
```suggestion
by small impedances, and have quite different voltage targets. This incompatibility is calculated through the admittance matrix of the network, that is used to compute an indicator between each voltage control : the greater the voltage difference is (given the impedance), the higher the indicator value will be, meaning an incompatible voltage target. In some situtations, this could lead to a failure in the Newton-Rapshon convergence.

```

SylvestreSakti (changes requested):
I'm also ok with this PR. Still, despite being the main use case, is this feature to be used only for **remote** voltage control ? For me it can be used for any type of voltage control, which would need a renaming of the **~Remote~VoltageTargetChecker** and the associated parameter name.

SylvestreSakti (dismissed):

PR #1116, opened 2024-11-07
Fix injections ignored on slack buses when using multi slacks

vmouradian:

Bug fix

When using multi slack, loads or gens on slack buses are ignored because value for P in target vector is 0. This results in the existing injections to be "mangled"/reported in the slack injection incorrectly.

Take P injection into account on slack buses equations for the multi slack case

- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
@@ -1037,13 +1037,13 @@ public class AcEquationSystemCreator {
             for (int i = 1; i < slackBuses.size(); i++) {
                 LfBus slackBus = slackBuses.get(i);
                 // example for 2 slack buses
-                // 0 = slack_p1 - slack_p2
-                // 0 = slack_p1 - slack_p3
+                // 0 = slack_p2 - slack_p1
+                // 0 = slack_p3 - slack_p1
                 equationSystem.createEquation(slackBus, AcEquationType.BUS_DISTR_SLACK_P)
-                        .addTerms(createActiveInjectionTerms(firstSlackBus, equationSystem.getVariableSet()))
-                        .addTerms(createActiveInjectionTerms(slackBus, equationSystem.getVariableSet()).stream()
+                        .addTerms(createActiveInjectionTerms(firstSlackBus, equationSystem.getVariableSet()).stream()
                                 .map(EquationTerm::minus)
-                                .collect(Collectors.toList()));
+                                .toList())
+                        .addTerms(createActiveInjectionTerms(slackBus, equationSystem.getVariableSet()));
                 // to update open/close terms activation
                 for (LfBranch branch : slackBus.getBranches()) {
                     updateBranchEquations(branch);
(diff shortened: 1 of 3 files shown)


src/test/java/com/powsybl/openloadflow/ac/MultipleSlackBusesTest.java
```diff
@@ -107,4 +110,25 @@ void nonImpedantBranchTest() {
...
+        LoadFlowResult.ComponentResult componentResult = result.getComponentResults().get(0);
+        List<LoadFlowResult.SlackBusResult> slackBusResults = componentResult.getSlackBusResults();
+
+        TwoWindingsTransformer t2wtLoad = network.getTwoWindingsTransformer("NHV2_NLOAD");
+        Load load = network.getLoad("LOAD");
+
+        assertEquals(-600.710, t2wtLoad.getTerminal2().getP(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(600.0, load.getTerminal().getP(), LoadFlowAssert.DELTA_POWER);
+
+        // still small difference due to NR conv epsilon per eq
```
jeandemanged, line 129: maybe just change the parameter so we can see same values ?

src/test/java/com/powsybl/openloadflow/ac/MultipleSlackBusesTest.java
```diff
@@ -107,4 +110,25 @@ void nonImpedantBranchTest() {
         assertEquals(-0.005, slackBusResults.get(0).getActivePowerMismatch(), LoadFlowAssert.DELTA_POWER);
         assertEquals(-0.005, slackBusResults.get(1).getActivePowerMismatch(), LoadFlowAssert.DELTA_POWER);
     }
+
+    @Test
+    void targetVectorBug() {
```
jeandemanged, line 115:
```suggestion
    void testSlackBusWithInjection() {
```

jeandemanged (changes requested):

jeandemanged (approved):
ty

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -1036,14 +1036,14 @@ private void createMultipleSlackBusesEquations(EquationSystem<AcVariableType, Ac
             LfBus firstSlackBus = slackBuses.get(0);
             for (int i = 1; i < slackBuses.size(); i++) {
                 LfBus slackBus = slackBuses.get(i);
-                // example for 2 slack buses
-                // 0 = slack_p1 - slack_p2
-                // 0 = slack_p1 - slack_p3
+                // example for 3 slack buses
+                // target_p2 = slack_p2 - slack_p1 + inj_p2
+                // target_p3 = slack_p3 - slack_p1 + inj_p2
```
SylvestreSakti, line 1041: Not sure about my remark but shouldn't we rather have : 
```
// target_p2 = slack_p2 - slack_p1
// target_p3 = slack_p3 - slack_p1
```

jeandemanged: you are absolutely right ... thanks / fixed

SylvestreSakti (approved):

PR #1118, opened 2024-11-07
DC Load Flow reference bus and multi slack support

### LINKED ISSUE #1117: Bug : DC used as voltage init with mutliple slack buses

When using FULL_VOLTAGE voltage initializer with AC LF. The network can be created with mutliple slcak buses.
But then the DC LF used for phase init handles that in an incorrect way: all slack buses have no P balance equation and have phi = 0.
Gives false results and can even crash if a non-impedant branch connects two slack buses

Handle multi slack as in AC : one reference bus with phi = 0 and slack equally spread over all slack buses

### Describe the steps

_No response_

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

vmouradian:

Handles multi slack for DC load flow
Fixes #1117

Feature and bugfix

see #1117

DC Load Flow properly account for the selected reference bus - also important for the DC Values Voltage Initializer before the Newton-Raphson.
Only first slack bus has p balance equation deactivated.
For other slack buses the corresponding value in target vector is updated : (finalTarget = targetP + totalSlack/nbSlackBuses) 

- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/AbstractLfNetworkLoaderPostProcessor.java
+++ b/src/main/java/com/powsybl/openloadflow/network/AbstractLfNetworkLoaderPostProcessor.java
@@ -36,4 +36,9 @@ public abstract class AbstractLfNetworkLoaderPostProcessor implements LfNetworkL
     public void onAreaAdded(Object element, LfArea lfArea) {
         // to implement
     }
+
+    @Override
+    public void onLfNetworkLoaded(Object element, LfNetwork lfNetwork) {
+        // to implement
+    }
 }
--- a/src/main/java/com/powsybl/openloadflow/network/LfNetworkLoaderPostProcessor.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfNetworkLoaderPostProcessor.java
@@ -37,4 +37,6 @@ public interface LfNetworkLoaderPostProcessor {
     void onInjectionAdded(Object element, LfBus lfBus);
 
     void onAreaAdded(Object element, LfArea lfArea);
+
+    void onLfNetworkLoaded(Object element, LfNetwork lfNetwork);
 }
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
@@ -1028,6 +1028,7 @@ public class LfNetworkLoaderImpl implements LfNetworkLoader<Network> {
             lfNetwork.writeGraphViz(debugDir.resolve("lfnetwork-" + dateStr + ".dot"), parameters.getLoadFlowModel());
         }
 
+        postProcessors.forEach(pp -> pp.onLfNetworkLoaded(network, lfNetwork));
         return lfNetwork;
     }
 


src/test/java/com/powsybl/openloadflow/ac/MultipleSlackBusesTest.java
jeandemanged, line 1: * Do not put DC tests with AC tests, DC tests should be in package `com.powsybl.openloadflow.dc` e.g. existing `DcLoadFlowTest`
* make use of LoadFlowAssert utility class methods, e.g.:

```diff
- assertEquals(0, nGen.getAngle(), LoadFlowAssert.DELTA_ANGLE);
+ assertAngleEquals(0.0, nGen);
```

or

```diff
- assertEquals(301.75, l12A.getTerminal1().getP(), LoadFlowAssert.DELTA_POWER);
+ assertActivePowerEquals(301.75, l12A.getTerminal1());
```

etc

src/main/java/com/powsybl/openloadflow/dc/DcTargetVector.java
```diff
@@ -24,7 +25,12 @@ public class DcTargetVector extends TargetVector<DcVariableType, DcEquationType>
     public static void init(Equation<DcVariableType, DcEquationType> equation, LfNetwork network, double[] targets) {
         switch (equation.getType()) {
             case BUS_TARGET_P:
-                targets[equation.getColumn()] = network.getBus(equation.getElementNum()).getTargetP();
+                LfBus bus = network.getBus(equation.getElementNum());
+                targets[equation.getColumn()] = bus.getTargetP();
```
jeandemanged, line 29: move `targets[equation.getColumn()] = bus.getTargetP();` in an `else` (=not slack) block below ?

src/main/java/com/powsybl/openloadflow/dc/DcTargetVector.java
```diff
@@ -24,7 +25,12 @@ public class DcTargetVector extends TargetVector<DcVariableType, DcEquationType>
...
+                // Only used for multi slack (BUS_TARGET_P equation is disabled for first slack bus)
+                if (bus.isSlack()) {
+                    targets[equation.getColumn()] += DcLoadFlowEngine.getActivePowerMismatch(network.getBuses()) / network.getSlackBuses().size();
```
jeandemanged, line 32: compute this value only once at the beginning of the init method e.g.

```java
final double slackActivePower = DcLoadFlowEngine.getActivePowerMismatch(network.getBuses()) / network.getSlackBuses().size();
```

jeandemanged (changes requested):
I would be nice also to explain multi slack handling in documentation.
In DC we can take shortcuts because it is a lossless model. In AC it is a different story.

src/main/java/com/powsybl/openloadflow/dc/DcTargetVector.java
jeandemanged: nononon I am wrong, forget it

src/main/java/com/powsybl/openloadflow/dc/DcTargetVector.java
jeandemanged: ah no this would be wrong, forget it.

src/main/java/com/powsybl/openloadflow/dc/equations/DcEquationSystemCreator.java
```diff
@@ -38,9 +38,11 @@ private void createBuses(EquationSystem<DcVariableType, DcEquationType> equation
...
             var p = equationSystem.createEquation(bus, DcEquationType.BUS_TARGET_P);
             bus.setP(p);
-            if (bus.isSlack()) {
+            if (bus.isReference()) {
                 equationSystem.createEquation(bus, DcEquationType.BUS_TARGET_PHI)
                         .addTerm(equationSystem.getVariable(bus.getNum(), DcVariableType.BUS_PHI).createTerm());
+            }
+            if (bus == network.getSlackBus()) {
                 p.setActive(false);
             }
```
jeandemanged, line 47: add a comment explaining the why only first slack bus, in line with what is done on target vector

jeandemanged (changes requested):

geofjamg:
> I would be nice also to explain multi slack handling in documentation. In DC we can take shortcuts because it is a lossless model. In AC it is a different story.

So what is the purpose of this feature. In AC, the only aim was to allow convergence for very specific model (large and very imbalanced that was diverging at first NR). But in DC with just a simple linear system solving what could be the point?

src/main/java/com/powsybl/openloadflow/dc/DcTargetVector.java
vmouradian: In my comprehension, this method is called for each equation. Which means that this value would be calculated for each equation instead of just for the multi-slack ones as it is in the current version

jeandemanged: oh right sorry

src/test/java/com/powsybl/openloadflow/dc/DcMultipleSlackBusesTest.java
jeandemanged, line 1: add copyright & authorship

jeandemanged (changes requested):

src/test/java/com/powsybl/openloadflow/dc/DcMultipleSlackBusesTest.java
vmouradian: added in [d139083](https://github.com/powsybl/powsybl-open-loadflow/pull/1118/commits/d139083460fa932836efc415ee42566967713fef)

jeandemanged (approved):
ty

src/test/java/com/powsybl/openloadflow/dc/DcMultipleSlackBusesTest.java
SylvestreSakti, line 1: Can we imagine parametrizing the tests in `AcMultipleSlackBusesTests` rather than having a whole class `DcMultipleSlackBusesTest` that implements the same cases ? (To do so maybe a condition can be added in the test to differentiate asserts values between AC and DC)
By doing so, when adding new tests in the future, we would be sure to implement them in both AC and DC mode.

vmouradian: I agree, here a proposal for this : [2b39a11](https://github.com/powsybl/powsybl-open-loadflow/pull/1118/commits/2b39a11fc4db2b3be6756d3d8926993b738fb6a7), what do you think ?

SylvestreSakti: Thank you ! It is good for me. https://github.com/powsybl/powsybl-open-loadflow/pull/1116 has been merged so the values can be changed for the corresponding test.

src/main/java/com/powsybl/openloadflow/dc/DcTargetVector.java
```diff
@@ -24,7 +25,12 @@ public class DcTargetVector extends TargetVector<DcVariableType, DcEquationType>
     public static void init(Equation<DcVariableType, DcEquationType> equation, LfNetwork network, double[] targets) {
         switch (equation.getType()) {
             case BUS_TARGET_P:
-                targets[equation.getColumn()] = network.getBus(equation.getElementNum()).getTargetP();
+                LfBus bus = network.getBus(equation.getElementNum());
+                targets[equation.getColumn()] = bus.getTargetP();
+                // Only used for multi slack (BUS_TARGET_P equation is disabled for first slack bus)
+                if (bus.isSlack()) {
```
annetill, line 31: Hi @jeandemanged and @vmouradian, reading this line with @geofjamg today, I'am not sure your implementation is resilient to disabled buses (slack or not). I think a test case with a connectivity loss (including a participating element) could be imagine to verify this. Think about the case where we loose a slack bus, I don't remember what we support...

PR #1119, opened 2024-11-08
Check too far voltage remote control

SylvestreSakti:

No

This change introduces a new feature that checks if a controller bus is too far from controlled bus when using voltage remote control (introducing a new parameter which is `maxVoltageRemoteControlDistance`)

There is no check of distance between buses

If a controller bus has a distance (number of branches) to its controlled bus that is higher than `maxVoltageRemoteControlDistance`, then the voltage remote control switches to local voltage control.

(`maxVoltageRemoteControlDistance = 0` (or a negative value) deactivates the distance check. Default value for this parameter is `2`)

- [ ] Yes
- [x] No


vidaldid-rte:
As discussed, can you also run this test during a contingence propagation (and make sure the status is correctly restored at the end).
The goal is to avoid a divergence in an AS or a Sensi computation if a line loss extends the distance between a group and its regulated point and make the next NR fail.

src/main/java/com/powsybl/openloadflow/util/BusDistance.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Breadth first search algorithm to compute distance between two LfBus in a network
+     * @param bus1              first LfBus (from which the breadth first search starts)
+     * @param bus2              second LfBus
+     * @param maxDistanceSearch the algorithm searches until this range and stops after this limit (or if every bus have been checked)
+     * @return                  measured distance (number of branches) or Integer.MAX_VALUE if bus2 is not found
+     */
+    public static int distanceBetweenBuses(LfBus bus1, LfBus bus2, int maxDistanceSearch) {
```
geofjamg, line 32: This is an unusual way to implement a BFS. We can do it much simpler without these kind of set manipulation (busesToCheck.removeAll(checkedBuses)). Also we should take care for a BFS to have an orderer neighbors traversal (be careful with HashSets)

geofjamg (changes requested):

docs/loadflow/parameters.md
```diff
@@ -140,6 +140,12 @@ If set to false, any existing voltage remote control is converted to a local con
 according to the nominal voltage ratio between the remote regulated bus and the equipment terminal bus.  
 The default value is `true`.
 
+**maxVoltageRemoteControlDistance**
+Defines the maximum distance allowed between the remote regulated bus and the voltage controller (requires `voltageRemoteControl` to be set to `true`).
+This distance is measured in number of branches separating the two buses. If the distance is higher than this parameter, then the voltage remote control is converted to a local control, rescaling the target voltage according to the nominal voltage ratio between the remote regulated bus and the equipment terminal bus.
+If `maxVoltageRemoteControlDistance` is set to `0`, there is no distance checking.
+The default value is `0`.
+
```
vidaldid-rte, line 148: I'm not sure of the default value.
Should we privilege backward compatibility (default = 0)  or robustness (default = 2 ) ?

src/main/java/com/powsybl/openloadflow/util/BusDistance.java
vidaldid-rte: Can you add a unit test for the method with various cases (exected distance, farther than maxDistance  etc..)

src/main/java/com/powsybl/openloadflow/network/LfNetworkParameters.java
```diff
@@ -597,6 +602,22 @@ public LfNetworkParameters setAreaInterchangeControlAreaType(String areaIntercha
...
     }
 
+    public int getMaxVoltageRemoteControlDistance() {
+        return maxVoltageRemoteControlDistance;
+    }
+
+    public static int checkMaxVoltageRemoteControlDistance(int maxVoltageRemoteControlDistance) {
+        if (maxVoltageRemoteControlDistance < 0) {
+            throw new IllegalArgumentException("Invalid value for parameter maxVoltageRemoteControlDistance: " + maxVoltageRemoteControlDistance);
+        }
```
vidaldid-rte, line 612: The message is not clear. It should explain the rules (>=0 ?)

But do we need this ? Can't we say that <=0 means no check ?

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -127,14 +128,24 @@ private static void createVoltageControls(List<LfBus> lfBuses, LfNetworkParamete
...
+                if (keepVoltageRemoteControl) {
                     controlledBus.getGeneratorVoltageControl().ifPresentOrElse(
                         vc -> updateGeneratorVoltageControl(vc, controllerBus, controllerTargetV),
                         () -> createGeneratorVoltageControl(controlledBus, controllerBus, controllerTargetV, voltageControls, parameters));
                 } else {
                     // if voltage remote control deactivated and remote control, set local control instead
-                    LOGGER.warn("Remote voltage control is not activated. The voltage target of {} with remote control is rescaled from {} to {}",
-                            controllerBus.getId(), controllerTargetV, controllerTargetV * controllerBus.getNominalV() / controlledBus.getNominalV());
                     controlledBus.getGeneratorVoltageControl().ifPresentOrElse(
                         vc -> updateGeneratorVoltageControl(vc, controllerBus, controllerTargetV), // updating only to check targetV uniqueness
```
vidaldid-rte, line 150: What is the difference with the if case ? I do not understand how the voltage control is transformed to a local control in this case...

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -71,6 +71,16 @@ public static void reportNotUniqueControlledBus(ReportNode reportNode, String ge
...
     }
 
+    public static void reportTooFarControlledBus(ReportNode reportNode, String controllerBusId, String controlledBusId, int maxVoltageRemoteControlDistance) {
+        reportNode.newReportNode()
+                .withMessageTemplate("tooFarControllerBus", "Remote voltage controlled bus ${controlledBusId} is too far from controller bus ${controllerBusId} (maximum distance is ${maxVoltageRemoteControlDistance}). Switching to local voltage control")
+                .withUntypedValue(CONTROLLED_BUS_ID, controlledBusId)
+                .withUntypedValue(CONTROLLER_BUS_ID, controllerBusId)
+                .withUntypedValue("maxVoltageRemoteControlDistance", maxVoltageRemoteControlDistance)
+                .withSeverity(TypedValue.ERROR_SEVERITY)
+                .add();
```
vidaldid-rte, line 81: For the log -- do we keep ERROR or change to WARNING ?
(In my definition, and ERROR is something that causes a failure - and this is not the case here)

docs/loadflow/parameters.md
geofjamg: Yes default to 2 would be a good start.

vidaldid-rte:
@geofjamg  @SylvestreSakti  shouldn't this be put into https://github.com/powsybl/powsybl-open-loadflow/pull/1115
This way it could be used:
   - to detect incorrect voltage controls
   - to remove (or switch to local ?) incorrect remote voltage controls

geofjamg:
> @geofjamg @SylvestreSakti shouldn't this be put into #1115 This way it could be used:
> 
> * to detect incorrect voltage controls
> * to remove (or switch to local ?) incorrect remote voltage controls

This is somehow related to #1115 indeed but #1115 is already quite large and we should keep it as separatly.
Regarding #1115, it will be a useful feature only if we are able to discard a majority a inconsistencies (and making bad cases mosttly converging). If not the case I would go for abandonning it. Anyway I didn't finished this work and there I still have some ideas to test.

vidaldid-rte:
> > @geofjamg @SylvestreSakti shouldn't this be put into #1115 This way it could be used:
> > 
> > * to detect incorrect voltage controls
> > * to remove (or switch to local ?) incorrect remote voltage controls
> 
> This is somehow related to #1115 indeed but #1115 is already quite large and we should keep it separatly. Regarding #1115, it will be a useful feature only if we are able to discard a majority a inconsistencies (and making bad cases mosttly converging). If not the case I would go for abandonning it. Anyway I didn't finished this work and there I still have some ideas to test.

OK. Let's put this on hold for now. My note was more related to the idea of creating a service as we have in #115 to detect inconsistent voltage control in an IIDM network. If we do that then I wonder if we should not show all voltage controls that are removed by OLF (such as this one - too far control) but also others existing (unrealistic / qmax-qmin to small etc...)

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -127,14 +128,24 @@ private static void createVoltageControls(List<LfBus> lfBuses, LfNetworkParamete
                     }
                 });
 
-                if (parameters.isGeneratorVoltageRemoteControl() || controlledBus == controllerBus) {
+                boolean keepVoltageRemoteControl = false;
+                if (!parameters.isGeneratorVoltageRemoteControl() && controlledBus != controllerBus) {
+                    LOGGER.warn("Remote voltage control is not activated. The voltage target of {} with remote control is rescaled from {} to {}",
+                            controllerBus.getId(), controllerTargetV, controllerTargetV * controllerBus.getNominalV() / controlledBus.getNominalV());
+                } else if (parameters.getMaxVoltageRemoteControlDistance() != 0 && distanceBetweenBuses(controlledBus, controllerBus, parameters.getMaxVoltageRemoteControlDistance()) > parameters.getMaxVoltageRemoteControlDistance()) {
+                    LOGGER.error("Voltage controlled bus '{}' is too far from controller bus '{}' (maximum distance is {}). The bus switches to local voltage control and target voltage is rescaled from {} to {}",
```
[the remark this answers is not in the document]
SylvestreSakti: I reproduced the severity of this check : checkUniqueTargetVControllerBus, but after some thoughts, I think a warning severity is indeed better (since this is not a topological error of the network but a criteria exceeding)

src/main/java/com/powsybl/openloadflow/network/LfNetworkParameters.java
SylvestreSakti: Yes this is simpler to say that <= 0 means no check

src/main/java/com/powsybl/openloadflow/util/Reports.java
SylvestreSakti: Same answer for severity, I changed it to `WARN_SEVERITY`

docs/loadflow/parameters.md
SylvestreSakti: I had to modify one unit test (`NotSameNumberVariableEquationIssueTest.test`) with this default value because there are buses that have a distance of 4. I prefered to directly modify the values of the assert instead of changing the parameter in the test which would have made it less clear.

src/test/java/com/powsybl/openloadflow/util/BusDistanceTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @BeforeEach
+    void setUp() {
+        network = LfNetwork.load(VoltageControlNetworkFactory.createWithGeneratorFarFromRemoteControl(),
+                new LfNetworkLoaderImpl(),
+                new LfNetworkParameters()).get(0);
+    }
+
+    @Test
+    void testBusDistance() {
+        bus1 = network.getBusById("vl1_0"); // Bus "b1"
```
ColasGael, line 40: Maybe also test the equality case: when `bus1 == bus2` ?

src/main/java/com/powsybl/openloadflow/util/BusDistance.java
```diff
@@ -0,0 +1,57 @@
...
+ * @author Sylvestre Prabakaran {@literal <sylvestre.prabakaran at rte-france.com>}
+ */
+public final class BusDistance {
+
+    private BusDistance() {
+    }
+
+    /**
+     * Breadth first search algorithm to compute distance between two LfBus in a network
+     * @param bus1              first LfBus (from which the breadth first search starts)
```
ColasGael, line 28: The search actually starts from `bus2`.

src/main/java/com/powsybl/openloadflow/util/BusDistance.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param bus2              second LfBus
+     * @param maxDistanceSearch the algorithm searches until this range and stops after this limit (or if every bus have been checked)
+     * @return                  measured distance (number of branches) or Integer.MAX_VALUE if bus2 is not found
```
ColasGael, line 31: And here, we try to find `bus1`.

SylvestreSakti: You're right. Thanks !

src/main/java/com/powsybl/openloadflow/util/BusDistance.java
geofjamg: Ok

vidaldid-rte:
> As discussed, can you also run this test during a contingence propagation (and make sure the status is correctly restored at the end). The goal is to avoid a divergence in an AS or a Sensi computation if a line loss extends the distance between a group and its regulated point and make the next NR fail.

As discussed - let's keep that refinement for later, it we see a problem with that.

src/test/java/com/powsybl/openloadflow/ac/GeneratorRemoteControlTest.java
```diff
@@ -975,4 +975,33 @@ void testGeneratorRemoteReactivePowerControlOutsideReactiveLimits() {
...
+
+        // First check with low maxRemoteVoltageControlDistance that detects too far controller buses
+        parametersExt.setMaxVoltageRemoteControlDistance(2);
+        LoadFlowResult result1 = loadFlowRunner.run(network, VariantManagerConstants.INITIAL_VARIANT_ID, LocalComputationManager.getDefault(), parameters, reportNode);
+        assertTrue(result1.isFullyConverged());
+        assertReactivePowerEquals(-157.5988, g1.getTerminal());
+        assertReactivePowerEquals(-189.2926, g2.getTerminal());
+        assertReactivePowerEquals(41.6880, g3.getTerminal());
+        LoadFlowAssert.assertReportEquals("/tooFarVoltageRemoteControlReport.txt", reportNode);
+
```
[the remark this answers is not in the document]
SylvestreSakti: Done.

Adding these asserts and showing the expected values enlighten the limits of the used formula to switch to local voltage control. We may find a better formula than just using the ratio of nominal voltages to compute the local target voltage. That can be discussed in the future.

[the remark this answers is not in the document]
vidaldid-rte: Thanks

vidaldid-rte (approved):

jeandemanged (changes requested):
Please clarify that this is only for generator remote voltage control:
- in MR title
- in parameter name
- in documentation

And also precising in the doc that generator means here: generators + svcs + vsc converters.

SylvestreSakti:
> Please clarify that this is only for generator remote voltage control:
> 
> * in MR title
> * in parameter name
> * in documentation
> 
> And also precising in the doc that generator means here: generators + svcs + vsc converters.

Thanks for this precision. Done !

jeandemanged (approved):

vidaldid-rte (commented):
I delay my approval since we are not certain of the right default. I would like to better understand what is a legitimate distance.

PR #965, opened 2024-01-23
Fix flow computation of HVDC connected at only one side

vidaldid-rte:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

Yes : https://github.com/powsybl/powsybl-open-loadflow/issues/953
This PR fixes the disconnection in situation N and completes https://github.com/powsybl/powsybl-open-loadflow/pull/961

**What kind of change does this PR introduce?**

When one of the LfBus is missing in the LfNetwork, the HVDC status is determined by the IIDM Network.
If that bus is isolated, the HVDC does not flow any active power.
The determination is  made before all LfBus are created for LCC stations - and thus contingencies that disconnect or reconnect an HDVC link with LCC station is still NOT SUPPORTED.  

**What is the current behavior?**

Active power flows through an HVDC even if one of the stations is dandling in situation N.

**What is the new behavior (if this is a feature change)?**
If one of the stations of an HVDC line is dandling in situation N, no active power flows through it. 

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [X] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

PR #1120, opened 2024-11-14
fictive load on isolated bus keeps hvdc isolated

vidaldid-rte:

No

In https://github.com/powsybl/powsybl-open-loadflow/pull/965 a simple criteria was added to detect an HVDC disconnection on one side : 
   A station is isolated if its merged bus is not connected to any other bus neither to a load or a group.

This criteria was meant to detect quickly contingencies that disconnect a line, while allowing simple models where one side is merely simplified as a group or a load.
But in real networks it happens that a fictive load is added to a station.

This fix considers that if an HVDC's stations contains only fictive loads it is still considered isolated.

**NOTE that this criteria for automatic HVDC disconnection in a contingncy is not documented at this time.** There is currently no good place to document how contingenices are propagated, when the propagation is not obvious.

If a contingency isolates a bus that is connected to a fictive load and an HVDC station, the HVDC is considered working.

If a contingency isolates a bus that is connected to a fictive load and an HVDC station, the HVDC is is stopped.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
@@ -413,6 +413,13 @@ public abstract class AbstractLfBus extends AbstractElement implements LfBus {
     }
 
     @Override
+    public double getNonFictitiousLoadTargetP() {
+        return loads.stream()
+                .mapToDouble(load -> load.getNonFictitiousLoadTargetP() * load.getLoadModel().flatMap(lm -> lm.getExpTermP(0).map(LfLoadModel.ExpTerm::c)).orElse(1d))
+                .sum();
+    }
+
+    @Override
     public double getLoadTargetQ() {
         return loads.stream()
                 .mapToDouble(load -> load.getTargetQ() * load.getLoadModel().flatMap(lm -> lm.getExpTermQ(0).map(LfLoadModel.ExpTerm::c)).orElse(1d))
--- a/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
@@ -297,7 +297,7 @@ public abstract class AbstractLfBus extends AbstractElement implements LfBus {
     }
 
     void addLccConverterStation(LccConverterStation lccCs, LfNetworkParameters parameters) {
-        if (!HvdcConverterStations.isHvdcDanglingInIidm(lccCs, parameters)) {
+        if (!HvdcConverterStations.isHvdcDanglingInIidm(lccCs)) {
             // Note: Load is determined statically - contingencies or actions that change an LCC Station connectivity
             // will continue to give incorrect result
             getOrCreateLfLoad(null, parameters).add(lccCs, parameters);
--- a/src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
@@ -58,14 +58,18 @@ public final class HvdcConverterStations {
     }
 
     private static boolean isIsolated(Terminal terminal, LfNetworkParameters parameters) {
-        Bus bus = parameters.isBreakers() ? terminal.getBusBreakerView().getBus() : terminal.getBusView().getBus();
+        Bus bus = terminal.getBusView().getBus();
         if (bus == null) {
             return true;
         }
 
-        // The criteria should as close as possible to Networks.isIsolatedBusForHvdc - only connected to the station
+        // The criteria should as close as possible to Networks.isIsolatedBusForHvdc - only connected to the station or a fictitious load
         return bus.getConnectedTerminalStream()
                 .map(Terminal::getConnectable)
-                .noneMatch(c -> !(c instanceof HvdcConverterStation<?> || c instanceof BusbarSection));
+                .noneMatch(c -> !(c instanceof HvdcConverterStation<?> || c instanceof BusbarSection || isFictitiousLoad(c)));
+    }
+
+    private static boolean isFictitiousLoad(Connectable c) {
+        return c instanceof Load load && LfLoadImpl.isLoadFictitious(load);
     }
 }
--- a/src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
@@ -9,7 +9,6 @@ package com.powsybl.openloadflow.network.impl;
 
 import com.powsybl.iidm.network.*;
 import com.powsybl.iidm.network.util.HvdcUtils;
-import com.powsybl.openloadflow.network.LfNetworkParameters;
 
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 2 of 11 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -412,6 +412,13 @@ public double getLoadTargetP() {
                 .sum();
     }
 
+    @Override
+    public double getNonFictitiousLoadTargetP() {
```
geofjamg, line 416: Are you sure there is no performance impact to compute this on the fly ?

src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
```diff
@@ -45,27 +44,31 @@ public static boolean isVsc(Identifiable<?> identifiable) {
...
-                return isIsolated(otherTerminal, parameters);
+                return isIsolated(otherTerminal);
             }).orElse(true); // it means there is no HVDC line connected to station
         }
     }
 
-    private static boolean isIsolated(Terminal terminal, LfNetworkParameters parameters) {
-        Bus bus = parameters.isBreakers() ? terminal.getBusBreakerView().getBus() : terminal.getBusView().getBus();
+    private static boolean isIsolated(Terminal terminal) {
+        Bus bus = terminal.getBusView().getBus();
```
geofjamg, line 60: I would rather agree on this change, but has it been motivated by something failing?

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
vidaldid-rte: The laternative I see would be 
- to compute incrementally this value for *all* buses in each event modifying the loads.
    Seems complex and gain is not obvious
- to cache it at bus creation, ignoring deconnection events, actions etc..
     Might cause bugs

Here it is computed;
   at each restore, for each HVDC station
   at contingency propagation for each HVDC
   and you pay an iteration price if the bus has several loads.

This being said I'll run some mesures of this cost on a large network.

src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
vidaldid-rte: No. Just the fact that when I wrote it first I didn't fully understand the difference betwwen BusView and BusBreakerVew

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
vidaldid-rte: It has ZERO performance impact the way it is used today.
The method was never called in a sensitivity analysis of a large network with 100 contingencies.
The reason is that restoreInitialTopology that will trigger the call for each HVDC link is called in very rare and targeted occasions - when some switches are moved. So I suggest to keep it simple stupid at this time.

src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
geofjamg: Indead it seems to be only called at Lf Network loading so only one time.

geofjamg: So maybe we should revert this change and think about it in another one.

vidaldid-rte: Oh la la... So it can breaks something in a topology I have seen where HVDC disconnection matters. I'll show you next time we meet.

geofjamg: So in that case, you just need to add a unit test that shows the issue.

vidaldid-rte: Done. That was a bit trickier that I thought - and the bug is more seldom that I though because:
   - the breaker mode is activated in very rare occasions
   - the criteria is only used for an HVDC part that is not in the LfNetwork
But the bug existed.

src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
```diff
@@ -45,27 +44,31 @@ public static boolean isVsc(Identifiable<?> identifiable) {
...
         if (bus == null) {
             return true;
         }
 
-        // The criteria should as close as possible to Networks.isIsolatedBusForHvdc - only connected to the station
+        // The criteria should as close as possible to Networks.isIsolatedBusForHvdc - only connected to the station or a fictitious load
         return bus.getConnectedTerminalStream()
                 .map(Terminal::getConnectable)
-                .noneMatch(c -> !(c instanceof HvdcConverterStation<?> || c instanceof BusbarSection));
+                .noneMatch(c -> !(c instanceof HvdcConverterStation<?> || c instanceof BusbarSection || isFictitiousLoad(c)));
```
SylvestreSakti, line 68: Just for readability : Can we remove the double negation ?

`return !bus.getConnectedTerminalStream().map(Terminal::getConnectable).anyMatch(c -> c instanceof HvdcConverterStation<?> || c instanceof BusbarSection || isFictitiousLoad(c));`

src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
```diff
@@ -45,27 +44,31 @@ public static boolean isVsc(Identifiable<?> identifiable) {
...
-    private static boolean isIsolated(Terminal terminal, LfNetworkParameters parameters) {
-        Bus bus = parameters.isBreakers() ? terminal.getBusBreakerView().getBus() : terminal.getBusView().getBus();
+    private static boolean isIsolated(Terminal terminal) {
+        Bus bus = terminal.getBusView().getBus();
         if (bus == null) {
             return true;
         }
 
-        // The criteria should as close as possible to Networks.isIsolatedBusForHvdc - only connected to the station
+        // The criteria should as close as possible to Networks.isIsolatedBusForHvdc - only connected to the station or a fictitious load
```
SylvestreSakti, line 65:
```suggestion
        // The criteria should be as close as possible to Networks.isIsolatedBusForHvdc - only connected to the station or a fictitious load
```

src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
vidaldid-rte: Your formula is not equivalent. You want to find nothing else than a station, a busbar or a fictive load.
But your statement says that if their is no station (which never happens by the way) or no busbar or no fictive load, the bus is isolated.

src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
vidaldid-rte: Done

src/main/java/com/powsybl/openloadflow/network/impl/HvdcConverterStations.java
SylvestreSakti: Sorry, you are right indeed :+1:

SylvestreSakti (approved):

PR #1123, opened 2024-11-18
DC area interchange control outerloop

vmouradian:

No

Feature

Area interchange control is not supported for DC load flow

Area interchange is supported, via a new DC outerloop, that shares most of the logic with the AC version.

Introduces : 
- refactor, classes/interfaces to share code between DistributedSlack and AC/DC AIC outerloops
- Result for distributed active power in DC LF

- [ ] Yes
- [x] No

**Other information**:


src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
```diff
@@ -1,67 +1,62 @@
-/**
- * Copyright (c) 2024, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
- * SPDX-License-Identifier: MPL-2.0
- */
-package com.powsybl.openloadflow.ac.outerloop;
+package com.powsybl.openloadflow.lf.outerloop;
```
vidaldid-rte, line 1: This file has nolonger copyright notice nor author

vidaldid-rte:
Can you add something in the doc that says that Area Interchange control works in DC and AC ? 
And with the limitations if any.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -1934,11 +1936,15 @@ public static DcLoadFlowParameters createDcParameters(LoadFlowParameters paramet
...
                 .setEquationSystemCreationParameters(equationSystemCreationParameters)
+                .setSlackDistributionFailureBehavior(parametersExt.getSlackDistributionFailureBehavior())
                 .setMatrixFactory(matrixFactory)
                 .setDistributedSlack(parameters.isDistributedSlack())
+                .setAreaInterchangeControl(parametersExt.isAreaInterchangeControl())
                 .setBalanceType(parameters.getBalanceType())
                 .setSetVToNan(true)
-                .setMaxOuterLoopIterations(parametersExt.getMaxOuterLoopIterations());
+                .setMaxOuterLoopIterations(parametersExt.getMaxOuterLoopIterations())
+                .setSlackBusPMaxMismatch(parametersExt.getSlackBusPMaxMismatch())
```
vidaldid-rte, line 1946: A naive question. Is the mismatch really needed given that the equations are linear ? Don't you just need an epsilon suitable for double comparison instead ?

src/main/java/com/powsybl/openloadflow/ac/AcOuterLoopContext.java
```diff
@@ -18,12 +18,23 @@
...
+    private int iteration; // current iterations of this single outer loop type
+
     private AcSolverResult lastSolverResult;
 
     AcOuterLoopContext(LfNetwork network) {
         super(network);
     }
 
+    @Override
+    public int getIteration() {
```
vidaldid-rte, line 30: Do not override the getters/setters. This is managed by the base class now.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcActivePowerDistributionOuterLoop.java
```diff
@@ -0,0 +1,26 @@
...
+import com.powsybl.openloadflow.lf.outerloop.ActivePowerDistributionOuterLoop;
+
+/**
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public interface AcActivePowerDistributionOuterLoop extends ActivePowerDistributionOuterLoop<AcVariableType, AcEquationType, AcLoadFlowParameters, AcLoadFlowContext, AcOuterLoopContext> {
+    @Override
+    default double getSlackBusActivePowerMismatch(AcOuterLoopContext context) {
+        return context.getLastSolverResult().getSlackBusActivePowerMismatch();
+    }
```
vidaldid-rte, line 25: Default methods in an interface are not meant to factorize code usually. But to extend an interface without breaking change. Why not moving the implementations to the two subclasses ?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
vmouradian: I think that there can be some (really) corner cases (with tielines not considered for interchange and a sum of interchange targets not exactly equal to 0), where using the slackBusPMax will make the outer loop tolerate a situation that would have failed if using a small epxilon.
I admit that I never created such test case... whether this is truly possible remains to be proven

src/main/java/com/powsybl/openloadflow/ac/AcOuterLoopContext.java
```diff
@@ -18,12 +18,23 @@
  */
 public class AcOuterLoopContext extends AbstractOuterLoopContext<AcVariableType, AcEquationType, AcLoadFlowParameters, AcLoadFlowContext> {
 
+    private int iteration; // current iterations of this single outer loop type
```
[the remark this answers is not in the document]
vmouradian: oh ! yes it was an oversigth. fixed in [269b938](https://github.com/powsybl/powsybl-open-loadflow/pull/1123/commits/269b93899538fd33f1e73cb50ffdbf2bcc87082e)

src/main/java/com/powsybl/openloadflow/ac/AcOuterLoopContext.java
vmouradian: same, fixed in [269b938](https://github.com/powsybl/powsybl-open-loadflow/pull/1123/commits/269b93899538fd33f1e73cb50ffdbf2bcc87082e)

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcActivePowerDistributionOuterLoop.java
vmouradian: I agree, I was a bit too much on a factorizing spree when I dit this 😅
fixed in [4830f6d](https://github.com/powsybl/powsybl-open-loadflow/pull/1123/commits/4830f6d3d4c9d0d2f8a3097ce827722c68f2098d)

src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
```diff
@@ -267,4 +246,13 @@ private Map<String, Double> allocateSlackDistributionParticipationFactors(LfNetw
         return areaSlackDistributionParticipationFactor;
     }
 
+    @Override
+    public OpenLoadFlowParameters.SlackDistributionFailureBehavior getSlackDistributionFailureBehavior(O context) {
+        OpenLoadFlowParameters.SlackDistributionFailureBehavior slackDistributionFailureBehavior = context.getLoadFlowContext().getParameters().getSlackDistributionFailureBehavior();
+        if (OpenLoadFlowParameters.SlackDistributionFailureBehavior.DISTRIBUTE_ON_REFERENCE_GENERATOR == slackDistributionFailureBehavior) {
+            logger.error("Distribute on reference generator is not supported in AcAreaInterchangeControlOuterLoop, falling back to FAIL mode");
```
m-guibert, line 253: The log mentions AcAreaInterchangeControlOuterLoop, but we are in the abstract class so we could also be in a DC AIC loop

src/test/java/com/powsybl/openloadflow/ac/AreaInterchangeControlTest.java
```diff
@@ -56,49 +58,88 @@ void twoAreasWithUnpairedDanglingLine() {
...
         runLfTwoAreas(network, -40, 40, -35, expectedIterationCount);
+        parameters.setDc(true);
+        runLfTwoAreas(network, -40, 40, -35, 0);
     }
 
     @Test
-    void remainingMismatchLeaveOneSlackBus() {
+    void acRemainingMismatchLeaveOnSlackBus() {
+        var mainComponentResult = remainingMismatchLeaveOnSlackBus();
+        assertEquals(-10, mainComponentResult.getDistributedActivePower(), 1e-3);
```
m-guibert, line 85: @vmouradian: why don't we have a similar assertion on DC unit tests?

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcAreaInterchangeControlOuterLoop.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public class AcAreaInterchangeControlOuterLoop
+        extends AbstractAreaInterchangeControlOuterLoop<AcVariableType, AcEquationType, AcLoadFlowParameters, AcLoadFlowContext, AcOuterLoopContext>
+        implements AcOuterLoop, AcActivePowerDistributionOuterLoop {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AcAreaInterchangeControlOuterLoop.class);
+
+    public static final String NAME = "AreaInterchangeControl";
```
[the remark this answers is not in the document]
vidaldid-rte: If renamed, that would be a breaking change for those who use the explicit outerloop construction for example.

src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
vmouradian: fixed in [068919d](https://github.com/powsybl/powsybl-open-loadflow/pull/1123/commits/068919d5b58978d567ee7ab4eb96fa8f7b8c7d41)

src/test/java/com/powsybl/openloadflow/ac/AreaInterchangeControlTest.java
vmouradian: I forgot to update those tests indeed. They are factorized and made identical for AC and DC in [d3da3ee](https://github.com/powsybl/powsybl-open-loadflow/pull/1123/commits/d3da3ee22652e2dfd91b79fb7012617f3e825bb3)

src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
vmouradian: Fixed in [2c979b1](https://github.com/powsybl/powsybl-open-loadflow/pull/1123/commits/2c979b1a0230fb2d7485702af79f924a36f879f8) (wrong commit message)

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
vmouradian: In the outerloop we have two phases : 

1) Check for each area if the interchange is closer to the target than AreaInterchangePmaxMismatch, and distribute the difference if not

When all areas are at their target : 

2) Check if some mismatch remains on the slack bus (with SlackBusPMaxMismatch) and distribute the difference : 
    - to the buses without Area first
    - to the buses with area if some mismatch still remains
   
There can be mismatch at stage 2) if the sum of the targets is not exactly 0 or if we have interconnections not accounted in the interchange so there is a difference between areas target and actual interchanges.
This situations make that some P mismatch will be present on the slack bus. 

Let's take the example where we have a network with all buses having an Area and where this slack mismatch is under SlackBusPMaxMismatch. 

- In case we use SlackBusPMaxMismatch as a threshold, this mismatch is tolerated and the LF is sucessful.

- In case we use a small epsilon, two scenarios : 
   - (most likely) : the mismatch is distributed to all network, and all Areas stay close enough to their interchange targets -> LF success and we just made 1 more iteration of outer loop
   - (ultra corner case imo) : the mismatch is distributed but some Area's interchange was very close to [target + AreaInterchangePmaxMismatch] (possible in DC if unconsidered interconnections) and this distribution of slack makes the Area be not balanced anymore -> adds at least 2 iterations to the OL or even more because the scenario could happen again.
   
To summarize, the important point is that the possibility of unconsidered interconnections adds uncertainty on the amount of power distributed by an Area, that can be different from its P imbalance.
In that context, I feel more confident to keep some tolerance on the slack mismatch threshold. 
But I admit that the probability of having serious issues due to this with that outerloop is very small, and would be ready to use an epsilon instead if you prefer, because indeed the benefits seem to outweigth the risks.

src/main/java/com/powsybl/openloadflow/ac/outerloop/AcAreaInterchangeControlOuterLoop.java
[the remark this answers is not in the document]
m-guibert: Ok, thank you for the answer!

vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -1923,7 +1923,10 @@ public static DcLoadFlowParameters createDcParameters(LoadFlowParameters paramet
                 .setSvcVoltageMonitoring(false)
                 .setMaxSlackBusCount(parametersExt.getMaxSlackBusCount())
                 .setLinePerUnitMode(parametersExt.getLinePerUnitMode())
-                .setReferenceBusSelector(ReferenceBusSelector.fromMode(parametersExt.getReferenceBusSelectionMode()));
+                .setLinePerUnitMode(parametersExt.getLinePerUnitMode())
```
m-guibert, line 1926: This new line is a duplicate of the unchanged line above.

jeandemanged (approved):

PR #1125, opened 2024-11-20
Refactor separation of post contingency calculations in Woodbury DC Security Analysis

p-arvy:

No.

The distinction between post contingency result computation is refactored in Woodbury DC Security Analysis.

In current behavior, code is duplicated. The only difference between the two calculations is the method used to calculate post contingency states (conditionned by post contingency connectivity result).

The code is mutualized. 

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
@@ -41,6 +41,7 @@ import com.powsybl.security.results.PreContingencyResult;
 import com.powsybl.security.strategy.OperatorStrategy;
 
 import java.util.*;
+import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


vidaldid-rte (changes requested):
This is much better ! Thanks.  Can you remove the WIP ?

p-arvy:
> This is much better ! Thanks. Can you remove the WIP ?

Thank you for the review. Done.

vidaldid-rte (approved):
As discussed with the team I would find it easier to read if the BiConsumer postContingencyResultAdder was instead defined as a method (because here we start reading runSimulations with this thing out of the calling context.
But OK to merge since this solves the issue of duplicated code.

p-arvy:
> As discussed with the team I would find it easier to read if the BiConsumer postContingencyResultAdder was instead defined as a method (because here we start reading runSimulations with this thing out of the calling context. But OK to merge since this solves the issue of duplicated code.

Okay, thanks for the comment. I moved the BiConsumer `postContingencyResultAdder` into a method to simplify reading. This should also help to better manage the computation logic when we'll add actions.

PR #1129, opened 2024-11-25
Fix secondary synchronous components in main CC not computed in SA

obrix:

No

Bugfix
Handle properly secondary synchronous component from main connected component in SA

Secondary synchronous component of main CC were ignored by the network selection on which we computed sa.

All component are now handled correctly and in a way similar as the load flow 👍 

MAIN component mode will compute SA on main connected and all its synchronous component.
ALL component mode will compute SA on all connected component and all synchronous component. 

- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
@@ -251,15 +251,21 @@ public abstract class AbstractSecurityAnalysis<V extends Enum<V> & Quantity, E e
     SecurityAnalysisResult runSimulationsOnAllComponents(LfNetworkList networks, List<PropagatedContingency> propagatedContingencies, P parameters,
                                                          SecurityAnalysisParameters securityAnalysisParameters, List<OperatorStrategy> operatorStrategies,
                                                          List<Action> actions, List<LimitReduction> limitReductions, LoadFlowParameters lfParameters) {
-        // run simulation on main connected component network first
-        SecurityAnalysisResult result = selectValidMainComponent(networks)
-                .map(n -> runSimulations(n, propagatedContingencies, parameters, securityAnalysisParameters, operatorStrategies, actions, limitReductions))
-                .orElse(createNoResult());
+
+        List<LfNetwork> networkToSimulate = new ArrayList<>(getNetworkToSimulate(networks, lfParameters.getConnectedComponentMode()));
+
+        if (networkToSimulate.isEmpty()) {
+            return createNoResult();
+        }
+
+        // run simulation on first lfNetwork to initialize results structures
+        LfNetwork firstNetwork = networkToSimulate.remove(0);
+        SecurityAnalysisResult result = runSimulations(firstNetwork, propagatedContingencies, parameters, securityAnalysisParameters, operatorStrategies, actions, limitReductions);
 
         List<PostContingencyResult> postContingencyResults = result.getPostContingencyResults();
         List<OperatorStrategyResult> operatorStrategyResults = result.getOperatorStrategyResults();
         NetworkResult mergedPreContingencyNetworkResult = result.getPreContingencyResult().getNetworkResult();
-        List<LimitViolation> preContingenctViolations = result.getPreContingencyResult().getLimitViolationsResult().getLimitViolations();
+        List<LimitViolation> preContingencyViolations = result.getPreContingencyResult().getLimitViolationsResult().getLimitViolations();
 
         Map<String, PostContingencyResult> postContingencyResultMap = new LinkedHashMap<>();
         Map<String, OperatorStrategyResult> operatorStrategyResultMap = new LinkedHashMap<>();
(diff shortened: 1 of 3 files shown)


vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -262,38 +261,56 @@ SecurityAnalysisResult runSimulationsOnAllComponents(LfNetworkList networks, Lis
...
+    }
+
+    static boolean isMainCCandMainSC(LfNetwork n) {
+        return n.getNumCC() == ComponentConstants.MAIN_NUM && n.getNumSC() == ComponentConstants.MAIN_NUM;
+    }
+
+    static boolean isMainCCandSecondarySC(LfNetwork n) {
+        return n.getNumCC() == ComponentConstants.MAIN_NUM && n.getNumSC() != ComponentConstants.MAIN_NUM;
+    }
+
```
jeandemanged, line 313: `AbstractSecurityAnalysis` doesn't seem to be the right place for these methods, move them to `LfNetworkList` ?

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -262,38 +261,56 @@ SecurityAnalysisResult runSimulationsOnAllComponents(LfNetworkList networks, Lis
...
         PreContingencyResult mergedPrecontingencyResult =
                 new PreContingencyResult(result.getPreContingencyResult().getStatus(),
                         new LimitViolationsResult(preContingenctViolations),
                         mergedPreContingencyNetworkResult);
         return new SecurityAnalysisResult(mergedPrecontingencyResult, postContingencyResults, operatorStrategyResults);
     }
 
+    static Optional<LfNetwork> selectValidMainComponent(LfNetworkList networks) {
+        return networks.getList().stream().filter(n -> isMainCCandMainSC(n) && n.getValidity().equals(LfNetwork.Validity.VALID)).findFirst();
+    }
```
jeandemanged, line 297: method name `selectValidMainComponent` is a confusing since we don't know if this means main CC or main SC.

It would be clearer with one method `List<LfNetwork> getNetworksToSimulate(LoadFlowParameters.ConnectedComponentMode mode)`

- in mode MAIN: returns all valid CC0 SCx
- in mode ALL: retuens all valid CCx SCy

LfNetworkList would be the right place for this method.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
jeandemanged, line 1: The new tests should actually run an SA and check what results are present/absent depending on connected component mode.

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
obrix: That is not what needs to be done here. I understand your approach but : 
We need a way here to find a "single" main component (usually CC0|SC0) to initiate the result structure. Here your MAIN mode would yield a list of several component on which we should again select the first one.

In fact thos methods are highly linked to the security analysis flow and only are this way because of the constraints on merged results building. 
This is also why I kept that in AbstractSecurityAnalysis, this is not supposed to be used elsewhere and will probably disappear when we rethink a more suitable structure for merged result someday.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
obrix: I think having unit tests here is a good thing, and there is already test running full SA with several components.
I can add a full test with a SA for the case that was not handled though

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
jeandemanged: I think this is exactly what needs to be done. Expectation is:
- in mode MAIN: runs all valid CC0 SCx
- in mode ALL: runs all valid CCx SCy
But the code is much more complicated, only to accommodate technical issues.
Also there is the assumption that CC0 SC0 always exists, which is as you wrote "usually" true / "usually" works, but can be wrong, I added a test case to demonstrate this.
Could be done later on if you prefer, but I find the code hard to follow.

jeandemanged (approved):
ty

src/main/java/com/powsybl/openloadflow/sa/AbstractSecurityAnalysis.java
```diff
@@ -251,49 +251,64 @@ SecurityAnalysisReport runSync(SecurityAnalysisParameters securityAnalysisParame
...
-                new PreContingencyResult(result.getPreContingencyResult().getStatus(),
-                        new LimitViolationsResult(preContingenctViolations),
-                        mergedPreContingencyNetworkResult);
+            new PreContingencyResult(result.getPreContingencyResult().getStatus(),
+                new LimitViolationsResult(preContingencyViolations),
+                mergedPreContingencyNetworkResult);
         return new SecurityAnalysisResult(mergedPrecontingencyResult, postContingencyResults, operatorStrategyResults);
     }
 
+    static List<LfNetwork> getNetworkToSimulate(LfNetworkList networks, LoadFlowParameters.ConnectedComponentMode mode) {
```
jeandemanged, line 299:
```suggestion
    static List<LfNetwork> getNetworksToSimulate(LfNetworkList networks, LoadFlowParameters.ConnectedComponentMode mode) {
```

jeandemanged (changes requested):
sorry one more small change please

jeandemanged (approved):

PR #1134, opened 2024-11-26
add ViolationLocation to LimitViolationManager

### LINKED ISSUE #1133: Add the violation location in voltage LimitViolations

The current implementation of violation locations in open-load-flow is missing a feature added in [powsybl-core #3153](https://github.com/powsybl/powsybl-core/pull/3153) and [powsybl-core #3178](https://github.com/powsybl/powsybl-core/pull/3178)

The expected behavior is to enhance the Limitviolation reporting by providing ViolationLocation

### Describe the motivation

Feature needed to make it easier for the user to read security analysis results

### Extra Information

_No response_

jamal-khey:

Fixes #1133 

- [ ] Yes
- [x] No

This PR is a continuation of the [powsybl-core violation location](https://github.com/powsybl/powsybl-core/pull/3178) for open-load-flow. Follow the same migration steps as described in that pull request.

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
@@ -202,7 +202,7 @@ public class LfBusImpl extends AbstractLfBus {
                 if (nodesByBus.containsKey(getBus().getId())) {
                     nodes = nodesByBus.get(getBus().getId()).stream().toList();
                 }
-                yield new NodeBreakerViolationLocation(getVoltageLevelId(), nodes);
+                yield nodes.isEmpty() ? null : new NodeBreakerViolationLocation(getVoltageLevelId(), nodes);
             }
             case BUS_BREAKER -> {
                 // are we in breaker mode ?
@@ -211,10 +211,11 @@ public class LfBusImpl extends AbstractLfBus {
                     yield new BusBreakerViolationLocation(List.of(getBus().getId()));
                 } else {
                     // Bus is a merged bus from thebus view
-                    yield new BusBreakerViolationLocation(busBreakerView
+                    List<String> busIds = busBreakerView
                             .getBusStreamFromBusViewBusId(getBus().getId())
                             .map(Identifiable::getId)
-                            .sorted().toList());
+                            .sorted().toList();
+                    yield busIds.isEmpty() ? null : new BusBreakerViolationLocation(busIds);
                 }
             }
         };
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
@@ -15,10 +15,7 @@ import com.powsybl.openloadflow.network.*;
 import com.powsybl.openloadflow.util.PerUnit;
 import com.powsybl.security.results.BusResult;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import java.util.stream.Collectors;
 
 /**
@@ -42,6 +39,10 @@ public class LfBusImpl extends AbstractLfBus {
 
     private final List<String> bbsIds;
 
+    private final TopologyKind topologyKind;
+
+    private List<Integer> nodes = new ArrayList<>();
+
     protected LfBusImpl(Bus bus, LfNetwork network, double v, double angle, LfNetworkParameters parameters,
                         boolean participating) {
         super(network, v, angle, parameters.isDistributedOnConformLoad());
@@ -61,6 +62,8 @@ public class LfBusImpl extends AbstractLfBus {
         } else {
             bbsIds = Collections.emptyList();
         }
+
+        topologyKind = bus.getVoltageLevel().getTopologyKind();
     }
 
     private static void createAsym(Bus bus, LfBusImpl lfBus) {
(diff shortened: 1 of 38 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -42,6 +40,10 @@ public class LfBusImpl extends AbstractLfBus {
 
     private final List<String> bbsIds;
 
+    private final TopologyKind topologyKind;
+
+    private List<Integer> nodes = new ArrayList<>();
```
EtienneLt, line 45: should it be better to initialize in the constructor when the powsybl core bus is present 
something like : 
```java
if (topologyKind == TopologyKind.NODE_BREAKER) {
            nodes =  Arrays.stream(bus.getVoltageLevel().getNodeBreakerView().getNodes()).boxed().toList();
        }
```

EtienneLt: and juste get the node in getNodes()
```java
public List<Integer> getNodes() {
     return nodes;
}
```

vidaldid-rte: Because getNodes is called only in specific situations and a few times for a few buses, I would prefer to remove the field.
No mempry cost nor initialization costs in the majority of cases where it is not used.
Computed and allocated only when building a violation result in node breaker mode.

src/main/java/com/powsybl/openloadflow/sa/LimitViolationManager.java
```diff
@@ -180,16 +176,25 @@ private void detectBusViolations(LfBus bus) {
...
     }
 
+    public static ViolationLocation createViolationLocation(LfBus bus) {
+        List<Integer> nodes = bus.getNodes();
+        if (nodes.isEmpty()) {
+            return null;
+        }
+
+        return new NodeBreakerViolationLocation(bus.getVoltageLevelId(), nodes);
+    }
```
vidaldid-rte, line 196: In Bus Breaker topology you should return a BusBreakerViolationDetection.

src/main/java/com/powsybl/openloadflow/network/impl/LfBusImpl.java
```diff
@@ -190,4 +194,16 @@ public double getTargetQ() {
...
+        if (topologyKind == TopologyKind.NODE_BREAKER) {
+            Map<String, Set<Integer>> nodesByBus = Networks.getNodesByBus(getBus().getVoltageLevel());
+            if (nodesByBus.containsKey(getBus().getId())) {
+                nodes = nodesByBus.get(getBus().getId()).stream().toList();
+            }
+
+        }
+        return nodes;
+    }
 }
```
vidaldid-rte, line 209: Can you add a test for this method, that verifies that the result is as expected ?

src/main/java/com/powsybl/openloadflow/sa/LimitViolationManager.java
```diff
@@ -180,16 +176,25 @@ private void detectBusViolations(LfBus bus) {
...
+
     /**
      * Detect violation limits on one voltage angle limit and add them to the given list
      * @param limit voltage angle limit of interest
```
vidaldid-rte, line 200: Can you add a test in NodeBreaker and BusBreaker that verifies that the results are as expected ?

src/main/java/com/powsybl/openloadflow/sa/LimitViolationManager.java
jamal-khey: done, but i had to add a new method `getTopologyKind`  for all sub-class of LfBus

vidaldid-rte: I pushed a commit that moves getLocation in LfBus. It reduces the numbre of new methods and encapsulate all the topolog logic into the LfBus.
I also added the case of the breaker LfNetwork (happens in Security Analysis if switches need to be open or closed in contingencies or actions).
Let me know if it works for you.

jamal-khey: Looks good! Moving getLocation to LfBus is better.

jamal-khey: i just added a commit for test coverage

vidaldid-rte: I added more tests to check the values, including in OLF breaker mode.

vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBus.java
```diff
@@ -75,4 +76,9 @@ public void updateState(LfNetworkStateUpdateParameters parameters) {
 
         super.updateState(parameters);
     }
+
+    @Override
+    public ViolationLocation getViolationLocation() {
+        return null;
+    }
```
EtienneLt, line 83: to prevent from implementing should it be better to have a default implementation in the interface lfBus ?

src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
```diff
@@ -70,4 +71,9 @@ public void updateState(LfNetworkStateUpdateParameters parameters) {
 
         super.updateState(parameters);
     }
+
+    @Override
+    public ViolationLocation getViolationLocation() {
+        return null;
+    }
```
EtienneLt, line 78: same here

EtienneLt (commented):
except for this possible refactory it is good for me

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBus.java
EtienneLt: ok

EtienneLt (approved):

PR #1135, opened 2024-11-27
Reduce irrealistic voltage created by remote voltage control 

### LINKED ISSUE #1124: Reduce irrealistic voltage created by remote voltage control

Remote voltage control is implemented by a QPV algorithm.
The controlled bus is QPV. THe controller is P only.
The equations expects that the controller delivers enough active power to the controlled bus to maintain the target voltage.
AN outerloop moves the controller to Q status if the  injections is out of its reactive limits.

But mechanically, this modifies that voltage of the controller bus - sometimes below the realistiv controls
Sometimes also, because the controllers voltage is a quadratic function of Q at the other end of al line, Q min is never reached at the conroller bus even with very low Q delivered at controlled bus.

When voltage targets are not too irrealistic do not produce irrealistic voltage at the stator.

Note: the method can not protect againts completelly irrealistic voltage targets that can create divergence (imposible to find a configuration hat keep the volrage at controller bus) or produce negative voltage at controller bus and LU factorisation issues. 
For irrealisti voltage targets, that method should be used in combination with an incorrect voltage target detector such as https://github.com/powsybl/powsybl-open-loadflow/pull/1115

### Describe the steps

Proposed solution:
- If ReactiveControlLimit outerloop is used:
-   deactivate irrealistic voltage control until ReactoveLimitOuterLoop is stable (idea similar to https://github.com/powsybl/powsybl-open-loadflow/pull/1087)
-   in ReactiveLimitOuterloop set the remote controller bus with irrealistiv voltage to PQ and Qmin/Qmax. Reset their voltage to 1.

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

vidaldid-rte:

Fixes #1124

A new parameter, **voltageRemoteControlRobustMode** is introduced. With value true by default.
When the robust mode is set to true, 
 - The enforcement of realistic voltage limit is performed only after the reactive limit outerloop is stabilized

the reactive limit outerloop is modified as follow:

If the bus of a generator with remote voltage control on , and reactivie limits, is approaches the realistic voltage limit:
    - If the bus is outside its reactive limits, it is move PQ (as usual) and its voltage is set to 1pu for the next NR run
    - If the bus is inside its reactive limites, it is removed from voltage control (and Q is set to initial targetQ). Its voltage is set to 1pu for the next NR run.

The Newton Raphson fails if a bus reaches a low voltage - a situation that can happen with remoted voltage control due to the PQV model.

With **voltageRemoteControlRobustMode** set to true (the default value) the resolution of networks with stressed or inconsistent remote voltage targets is more robust.

- [] Yes
- [ X ] No


src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
```diff
@@ -207,45 +215,91 @@ private static boolean switchPqPv(List<PqToPvBus> pqToPvBuses, ContextData conte
      *  - if Q equals to Qmin
```
SylvestreSakti, line 215: This comment should be updated :
```suggestion
     * (1) A bus PV bus can be switched to PQ in 3 cases:
     *  - if Q equals to Qmax
     *  - if Q equals to Qmin
     *  - if the controller bus is a remote controller bus and V is outside realistic voltage limits
```

src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
```diff
@@ -265,12 +319,19 @@ private static boolean switchReactiveControllerBusPq(List<ControllerBusToPqBus>
...
+                    case MAX_Q:
+                        LOGGER.trace("Remote reactive power controller bus '{}' -> PQ, q={} > maxQ={}", controllerBus.getId(), bus.q * PerUnit.SB,
+                                bus.qLimit * PerUnit.SB);
+                        break;
+                    case MIN_Q:
+                        LOGGER.trace("Remote reactive power controller bus '{}' -> PQ, q={} < minQ={}", controllerBus.getId(), bus.q * PerUnit.SB,
+                                bus.qLimit * PerUnit.SB);
+                        break;
+                    case MIN_V, MAX_V:
+                        LOGGER.trace("Remote reactive power controller bus '{}' -> PQ = targetQ - v outside realistic voltage limits",
```
SylvestreSakti, line 332: Same remark as above

SylvestreSakti (commented):
Some minor remarks on the code and I did some typo fix.

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -106,6 +108,16 @@ private void runOuterLoop(AcOuterLoop outerLoop, AcOuterLoopContext outerLoopCon
                 && runningContext.lastSolverResult.getStatus() == AcSolverStatus.CONVERGED
                 && runningContext.outerLoopTotalIterations < context.getParameters().getMaxOuterLoopIterations());
 
+        if (!canCheckUnrealistic && runningContext.lastUnrealisticStateFixingLoop == outerLoop && runningContext.lastSolverResult.getStatus() == AcSolverStatus.CONVERGED) {
```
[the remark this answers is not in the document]
vidaldid-rte: It is equivalent if you take into account that canCheckUnrealistic is defined today by "runningContext.lastUnrealisticStateFixingLoop == null"

But I prefer to explicit the condition to avoid side effect if that definition is changed.

The meaning is:
  if canCheckUnrealistic is false, then the check was not performed in solver.run
  And if we are the last outerloop that tolerates unrealistic voltage, then we must check if the loop has converged.
  
  I prefer to keep it written this way instead of assuming relationships between these variables.

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -106,6 +108,16 @@ private void runOuterLoop(AcOuterLoop outerLoop, AcOuterLoopContext outerLoopCon
...
+            // This is time to check the unrealistic state
+
+            if (solver.isStateUnrealisticForSolver(context.getNetwork().getReportNode())) {
```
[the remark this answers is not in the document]
vidaldid-rte: I wanted to explicit the conditions.
My logic is:
First condition; we have to check now the irrealistic values.
Second condition:
   We test the values (it is not only a test since it generates a report by the way and has really to be done if previou sconditions are true)
      and f they are irrealistic we update the result.

src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
```diff
@@ -137,18 +139,24 @@ private boolean switchPvPq(List<ControllerBusToPqBus> pvToPqBuses, int remaining
...
+                        case MAX_Q :
+                            LOGGER.trace("Switch bus '{}' PV -> PQ, q={} > maxQ={}", controllerBus.getId(), pvToPqBus.q * PerUnit.SB,
+                                    pvToPqBus.qLimit * PerUnit.SB);
+                            break;
+                        case MIN_Q:
+                            LOGGER.trace("Switch bus '{}' PV -> PQ, q={} < minQ={}", controllerBus.getId(), pvToPqBus.q * PerUnit.SB,
+                                    pvToPqBus.qLimit * PerUnit.SB);
+                            break;
+                        case MIN_V, MAX_V:
+                            LOGGER.trace("Switch bus '{}' PV -> PQ, q={} = targetQ - v outside realistic voltage limits", controllerBus.getId(), pvToPqBus.qLimit * PerUnit.SB);
```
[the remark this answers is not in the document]
vidaldid-rte: Why not adding the voltage. But my intention was mostly to tell that the bus is switch to PQ with Q = target Q which is unusual. Displaying targetQ is important to me because it explain the results (and I think it disappears from your suggested message and is replaced by the  of the bus when in irrealistic conditions)

src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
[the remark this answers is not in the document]
vidaldid-rte: I rewrote the part to avoid making Sonar believe this is just boolean arithmetic...

src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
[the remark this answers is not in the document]
vidaldid-rte: Trace Message updated with bus V and limits.

src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
vidaldid-rte: Info added but targetQ kept.

src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
```diff
@@ -349,4 +417,9 @@ public OuterLoopResult check(AcOuterLoopContext context, ReportNode reportNode)
         }
         return new OuterLoopResult(this, status);
     }
+
+    @Override
+    public boolean canFixUnrealisticState() {
+        return true;
```
SylvestreSakti, line 423: Shouldn't we check unrealistic state if robust mode is deactivated ?

If voltageRemoteControlRobustMode is set to `false`, I would put the value canFixUnrealisticState at `false` for this outerloop.

SylvestreSakti: After discussion, this value can stay to `true` because when the parameter `voltageRemoteControlRobustMode` is deactivated, `canFixUnrealisticState()` is not called.

vidaldid-rte: @SylvestreSakti after more thought I still think the outerloop should return true whether or not robustMode is true.
Here is why:
   Even passively, the outerloop can fix some irrealistic voltage (as noticed by Sebastien in PR #1087 
   This is also the case of the TrasformerVoltageControlOuterloop that can passively fix irrealistic voltages (that it may introduce when unstable when continuous tapratio are not within their bounds)

    So if in the future, someone wanted to delay voltage control after this outerloop, even if the ReactiveLimitOuterloop does nothing about robustification ( reset unrealistiv V to 1, or to set PQ controller buses within diagram but outside realistic voltage), he could implement this using this method.

SylvestreSakti (approved):

src/main/java/com/powsybl/openloadflow/ac/solver/AcSolver.java
```diff
@@ -17,5 +17,10 @@ public interface AcSolver {
 
     String getName();
 
-    AcSolverResult run(VoltageInitializer voltageInitializer, ReportNode reportNode);
+    AcSolverResult run(VoltageInitializer voltageInitializer, ReportNode reportNode, boolean canCheckUnrealistic);
```
jeandemanged, line 20: no strong opinion here... asking @geofjamg also for advice...
My understanding is that now the AcSolver needs to know things happening in AcLoadFlowEngine and OuterLoops.
Would it make sense instead to:
- move `canCheckUnrealistic` to `AcLoadFlowContext`
- change the API to provide the `AcSolver` with the `AcLoadFlowContext` 
?

src/main/java/com/powsybl/openloadflow/network/LfBus.java
```diff
@@ -21,7 +21,19 @@ public interface LfBus extends LfElement {
 
     enum QLimitType {
         MIN_Q,
-        MAX_Q
+        MAX_Q,
+        // Remote voltage control bus that have not reached a Q Limit but a realistic V limit
+        MIN_V,
+        MAX_V;
```
jeandemanged, line 27: minor: rename to `MIN/MAX_REALISTIC_V` ? (just for extra clarity)

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerReactivePowerControlTest.java
jeandemanged, line 1: do you need these changes ?

src/test/java/com/powsybl/openloadflow/ac/GeneratorRemoteControlPQSwitchTest.java
```diff
@@ -0,0 +1,274 @@
...
+        // Activate trace logs
+        LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
+        loggerContext.getLogger(ReactiveLimitsOuterLoop.class).setLevel(Level.TRACE);
+    }
+
+    @AfterEach
+    void restoreLogger() {
+        LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
+        loggerContext.getLogger(ReactiveLimitsOuterLoop.class).setLevel(null);
+    }
```
jeandemanged, line 170: do you need to keep this ?

src/test/java/com/powsybl/openloadflow/ac/GeneratorRemoteControlPQSwitchTest.java
```diff
@@ -0,0 +1,274 @@
...
+        parametersExt.setMinRealisticVoltage(0.8);
+        parametersExt.setMaxRealisticVoltage(1.2);
+        parametersExt.setVoltageRemoteControlRobustMode(robustMode);
+        parameters.setUseReactiveLimits(true);
+        g1.newMinMaxReactiveLimits().setMinQ(-3000).setMaxQ(6000).add();
+        g1.setTargetQ(10);
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        if (robustMode) {
+            assertTrue(result.isFullyConverged());
+            assertReactivePowerEquals(-10, g1.getTerminal());  // The group is set to initial targetQ
```
jeandemanged, line 227: group  -> generator (also 2 times below)

src/test/java/com/powsybl/openloadflow/ac/GeneratorRemoteControlPQSwitchTest.java
```diff
@@ -0,0 +1,274 @@
...
+    private Network network;
+    private Bus b1;
+    private Generator g1;
+    private LoadFlow.Runner loadFlowRunner;
+    private LoadFlowParameters parameters;
+    private OpenLoadFlowParameters parametersExt;
+
+    /**
+     * (401kV)                            (402kv)
+     *  b2 ----------------l23-------------b3---l3 (100mW)
```
jeandemanged, line 51: kv -> kV
mW -> MW

jeandemanged (changes requested):
ok for me, except one API design topic (AcSolver) that needs to be though about maybe.
Otherwise just minor comments.

src/main/java/com/powsybl/openloadflow/ac/solver/AcSolver.java
geofjamg: Indeed unrealistic state management could be seen as a hack compared to a clean modeller/solver design where in theory the solver should not known anything regarding the physics behing the equations to solve.
Looking at the code (to confirm) we should be able to move unrealistic state management out of the solver (check is done at the very end of the solve) and have all these code in the AcLoadFlowEngine which would be a cleaner design. Also the second solver Newton krylov would also befinit from it which is not the case here.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -463,6 +468,8 @@ public enum ReportedFeatures {
 
     private boolean voltageRemoteControl = VOLTAGE_REMOTE_CONTROL_DEFAULT_VALUE;
 
+    private boolean voltageRemoteControlRobustMode = VOLTAGE_REMOTE_CONTROL_ROBUST_MODE_DEFAULT_VALUE;
```
geofjamg, line 471: Instead of adding this new parameter, we could just replace voltageRemoteControl by a 3 values enum. That would simplify the configuration.

geofjamg (changes requested):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
vidaldid-rte: I see issues with that:
    - it is a breaking change
    - it is harder for users to use an enum value than using boolean values
    - how to make it the default ?

Now to brainstorm on the enum idea, we could consider
voltageRemoteControl:
     Off
     Robust
     Fragile ?

vidaldid-rte: Another proposal to kepp compatibility and have robust by default

false
true    (with robust)
reactiveLimitsOnly  (the original algo)

geofjamg: To keep backward compatibility you just have to keep voltageRemoteControl and setter and getter that set the corresponding mode and manage it when loading from platform config / json.
Values could be 
ON,
ON_ROBUST,
OFF

But maybe it would not worth the dev effort.

src/main/java/com/powsybl/openloadflow/ac/solver/AcSolver.java
jeandemanged: > Looking at the code (to confirm) we should be able to move unrealistic state management out of the solver (check is done at the very end of the solve) and have all these code in the AcLoadFlowEngine which would be a cleaner design. Also the second solver Newton krylov would also befinit from it which is not the case here.

Thanks for feedback, indeed if possible this would be the ideal solution. As I see it unrealistic state management is in NR solver only for historical reasons and should probably be moved out.

geofjamg: Yes, it is an already existing design issue even if not as visible as with this PR because only part of the solver API because of the return status.

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
geofjamg: After more thought let's keep it as it is as for backward compatibility especially for pypowsybl we would need to keep old parameter and make a hack for that.

src/test/java/com/powsybl/openloadflow/ac/GeneratorRemoteControlPQSwitchTest.java
vidaldid-rte: In reality the PV/PQ switches should probably be in the reports. 
It is useful information today. And with robust mode it would provide an explanation for PQ groups not a reactive limit bounds.

@geofjamg @jeandemanged  keep the info in logs today should I use this PR to move the PQ changes to the reports ?

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
jeandemanged: I'd go for OFF/ON/ROBUST
Robust can be the default.
For backward compat we may map `false` to OFF and `true` to ROBUST (in OpenLoadFlowParameters#update).
Note: be careful of pypowsybl users which may provide `True`/`False` and not `true`/`false`.

src/test/java/com/powsybl/openloadflow/ac/GeneratorRemoteControlPQSwitchTest.java
vidaldid-rte: logger activation removed.

src/test/java/com/powsybl/openloadflow/ac/GeneratorRemoteControlPQSwitchTest.java
vidaldid-rte: Done

src/test/java/com/powsybl/openloadflow/ac/GeneratorRemoteControlPQSwitchTest.java
vidaldid-rte: Done

src/main/java/com/powsybl/openloadflow/network/LfBus.java
vidaldid-rte: Done

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerReactivePowerControlTest.java
jeandemanged: oh sonar was the reason behind ... worst case this can be bypassed for merging, sonar most of the time points real issues, but occasionally may be too much a pain for just nothing.

vidaldid-rte: @geofjamg @jeandemanged 
Reporting the PV P switches is a bit large. I speparated into a different PR to merge after that one (https://github.com/powsybl/powsybl-open-loadflow/pull/1182)
This means that we may have to shut down Sonar alert on test coverage for that PR, or restore the LOGGER trace activation hack.

src/main/java/com/powsybl/openloadflow/ac/solver/AcSolver.java
vidaldid-rte: FYI, I'll create a separate PR to extract the voltage verification from Newton Raphson.

vidaldid-rte: @geofjamg  @jeandemanged  the PR to move the irrealistic voltage verification is here https://github.com/powsybl/powsybl-open-loadflow/pull/1188 . To be merged in main before that one on irrealistic voltage (and then I will adapt that one).

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
vidaldid-rte: @jeandemanged  I would vote like Geoffroy and keep as it is. I do not see an easy way to change voltageRemoteControl to an enum without creating a breaking change in one of the following:
         JSON,  the Java parameter API or in pypowsybl.

It believe that the way to change a parameter from bool to enum would be to deprecate the old one and create a new one that would be enum based. But it seems overkill here.

jeandemanged: I am ok with this no problem, my last message was sent almost same time as Geoffroy's one, I was just slower by 1 minute :)
Agree to keep as it is.

src/main/java/com/powsybl/openloadflow/ac/solver/AcSolver.java
vidaldid-rte: Merged. Requesting new review.

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowTransformerReactivePowerControlTest.java
vidaldid-rte: After more thought I reintroduced the trace log for the two tests.
I think it is good that the code that generates trace logs in the ReactiveLimitsOuterLoop is run at least once per build. This ensures that this won't break if someone needs to activate this log level for the class. So for me it justifies that we merge that.
And for this PR that was freshly refactored (moving up the realistic check) it is good to have a solide test coverage measure and easily spot was is left untested.

jeandemanged (changes requested):
Thank you, just minor polishing to do.

Please also update the PR description:
- not a breaking change anymore
- mention the new parameter and its default value

jeandemanged (approved):

PR #1138, opened 2024-11-29
Fix DC Load Flow bus imbalance

### LINKED ISSUE #1137: DC Loadflow: Wrong bus P balance if PST connected.  (equation eval ignores constant terms)
## Short explanation :

When using getP() on LfBus object. The result is the buse's BUS_TARGET_P equation.
But when evaluating with eval(), the right-hand-side part of the terms is not taken into account.
Which results in a wrong result.
This can happen in DC LF for phase shifting or hvdc branches.

## Detailed explanation
In case of a BUS_TARGET_P equation in DC LF.
Equation terms are used for branches' getP1()/getP2() and the whole equation is used for buse's getP().

But : 
branch.getP2().eval() = ClosedBranchSide2...term::eval =  variable part + rhs part 
bus.getP().eval() = Equation::eval = sum for term in terms { term.eval() - term.rhs()}

so bus.getP().eval() != sum of branch.getP2().eval() 

in case of DC LF, for a branch P2 :
variable part = (phi_2 - phi_1) / X_12
rhs part = (A_2 - a_1) / X_12

Which means that if a branch has a pst in DC -> it has a non null rhs (Aij/Xij) and the bus.getP().eval calculation is false beacuse a part of the active power is not taken into account.

## Case where it creates a bug :

lfBus.getP().eval() is used to calculate the flow on non impedant branches after the LF (the flow on the NI branch will be the P balance of the bus minus the flows on impedant branches).

So if a bus is connected to a branch with phase shifting AND a non impedant branch. The flow will be incorrectly calculated on the non impedant branch. Resulting in a P imbalance in the bus.

Calculate the actual flow balance of a bus when using lfBus.getP().eval()

### Describe the steps

_No response_

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

vmouradian:

fixes #1137

Bug fix

To solve this I have been facing the following issue.
Both branch.getP1() and bus.getP() are used fo two things :
- To use eval() and get the physical value for active power
- To cast it to Equation or EquationTerm and modify it or the equation system

But also :
bus.getP() -> Equation class  (.eval() = sum of term.eval - sum of term.rhs())
branch.getP1 -> ClosedBranchSide1DcFlowEquationTerm  (.eval contains rhs)

So I cannot just change the eval() of ClosedBranchSide1DcFlowEquationTerm because it is used to calculate the flow on the branch (needs to have rhs included) but also is summed to calculate P balance of bus in Equation eval() (needs to not have rhs since it is already added by the Equation eval() method)

The easiest solution I found is to create a BusPowerEquation class that extends Equation but that just calculates the eval() differently (sums the terms' eval() without substracting rhs)

Maybe a more long-term solution would be to create a layer of abstraction to make clearer what is an equation or equation term and what is a representation of some physical value

- [ ] Yes
- [ ] No

**Other information**:


src/main/java/com/powsybl/openloadflow/equations/EquationSystem.java
```diff
@@ -186,6 +188,14 @@ private Equation<V, E> addEquation(Pair<Integer, E> p) {
         return equation;
     }
 
+    private Equation<V, E> newEquation(int num, E type) {
```
geofjamg, line 191: This is really a terrible hack. It breaks the genericity if EquationSystem that is not supposed to know equations type (this is the purpose of the generic type). It creates bad cyclic dependencies between modules.

Not clear to me even at looking the issue to figure what really is the bug. Could you try to create a clear test case that show the issue (and not changing an existing one)? 

Anyway even if we really have an issue I'm sure we can do better to fix it.

geofjamg (dismissed):

src/main/java/com/powsybl/openloadflow/equations/EquationSystem.java
Hadrien-Godard: Correct me @vmouradian if I am wrong or if I forget something.

The issue is that, when you have a non impedant branch connected to a bus, you should recompute the flow on this branch. Surprisingly, we do not use the dummy_p variable value, instead we use the balance of the bus (calling bus.getP())
Flow on the branch = bus (im)balance = - sum (flows on other bus terminals)

When you have a PST on one branch connected to this bus, the flow generated by the shifting angle is not taken into account in DC when you call bus.getP(). -> this is the bug.

geofjamg: > Surprisingly, we do not use the dummy_p variable value, instead we use the balance of the bus (calling bus.getP())
> Flow on the branch = bus (im)balance = - sum (flows on other bus terminals)

So this is a hack on a hack ? :-) 
I think using dummy_p for getting flow on a branch only work for an isolated non impedant branch. When there is mutliple non impedant branches connected we have to rely on `ZeroImpedanceFlows`

geofjamg: > When you have a PST on one branch connected to this bus, the flow generated by the shifting angle is not taken into account in DC when you call bus.getP(). -> this is the bug.

I don't understand where is the bug in bus.getP(), we already take into account the RHS containing the phase shifting part.

jeandemanged: @vmouradian please:
- revert the changes on the existing tests
- add new tests showing in which case the issue happens

and then it will be much easier to discuss on the solution

Hadrien-Godard: I think that a better and easier method is to code an eval method in AbstractClosedBranchDcFlowEquationTerm without calling the "calculateSensi" one. The method should only compute ph2 - ph1 (or the opposite) times the "power" attribute.

geofjamg: Indeed the eval function was not supposed to take into account the RHS (shift) part. This is probably a regression introduced by sensi calculation. So let's fix the real problem.

geofjamg: > I think that a better and easier method is to code an eval method in AbstractClosedBranchDcFlowEquationTerm without calling the "calculateSensi" one. The method should only compute ph2 - ph1 (or the opposite) times the "power" attribute.

We can just add an option to calculateSensi to take into account the a1 variable or not depending on the context and call it without in the eval method.

vmouradian: If we just change the value given by AbstractClosedBranchDcFlowEquationTerm.eval() to (phi2 - ph1) / X,  this will fix the issue for the lfBus.getP(), but then the value for active power on pst branches will be wrong. (lfbranch.getP().eval() will be wrong -> not taking A2 into account)

vmouradian: What if we just consider that the eval() method of an equation term should contain the rhs part ? 
If we do so we simply need to remove [these lines](https://github.com/powsybl/powsybl-open-loadflow/blob/8822d5f16d09de29e7f3034650a1ec625894bd95/src/main/java/com/powsybl/openloadflow/equations/Equation.java#L159-L161) to fix the bug

jeandemanged: > What if we just consider that the result of eval() method for an equation term should contain the rhs part ? If we do so, we simply need to remove [these lines](https://github.com/powsybl/powsybl-open-loadflow/blob/8822d5f16d09de29e7f3034650a1ec625894bd95/src/main/java/com/powsybl/openloadflow/equations/Equation.java#L159-L161) to fix the bug +using the eval() like that is what is currently done in practice

I am not qualified enough to tell, but I can tell you that while OLF unit tests would be happy, on real test cases I get less imbalance KO but still many KO-s.

vidaldid-rte: > What if we just consider that the result of eval() method for an equation term should contain the rhs part ? If we do so, we simply need to remove [these lines](https://github.com/powsybl/powsybl-open-loadflow/blob/8822d5f16d09de29e7f3034650a1ec625894bd95/src/main/java/com/powsybl/openloadflow/equations/Equation.java#L159-L161) to fix the bug +using the eval() like that is what is currently done in practice

You can't do that in DC because there is an assumption by the DC engine that eval is linear, not affine.

vidaldid-rte: Maybe the solution is in DCEquationSystemCreator. For an non impedant branch, you may need to transfer the BUS_TARGET_P of one side to the other side. (maybe once all equations are created to be sure the get all RHS defined).

vmouradian: > You can't do that in DC because there is an assumption by the DC engine that eval is linear, not affine.

Ok thanks, I was not aware of that. Do you know where we can see this ?

vmouradian: > Maybe the solution is in DCEquationSystemCreator. For an non impedant branch, you may need to transfer the BUS_TARGET_P of one side to the other side. (**edit** - should be done in a way that works with securty analysis - disconnection of the element bringing the RHS - disconnection of the zero impedance line or one of its bus etc...)

I do not realy understand. The issue is that the eval() of the BUS_TARGET_P (or any other)  equations is wrong if one of the terms has rhs in his eval() (like pst or hvdc in DC). 
In my comprehension, your suggestion is transferring the equation from one side of the NI branch to the other, and still in my comprehension this should not solve the issue. Does it ?

vidaldid-rte: If I remember well, the DC engine doesn't call eval. SO any constant term has t be in the RHS.
Anyway, since this goes deep into the equation system we need a better coverage for this change thatn includes AC tests, sensi tests etc...
I wil try to spend time on this newt week.
For me the ideal solution would only touch the zero impedance equation logic.

vidaldid-rte: There is a contract in the equation system that says eval = rhs + target vector
If we break this contract we do not know the effect.
This is why I would like to be sure we fully understand what's going on.

vmouradian: For the moment i can suggest 3 solutions : 
- Remove the rhs lines in the Equation::eval()  (reset to [9d5a6ec](https://github.com/powsybl/powsybl-open-loadflow/pull/1138/commits/9d5a6ec04ea4cb85ca75b38589ac22d3af6941dc))
- Not count the rhs for Impedant branches in ZeroImpedanceFlows (see [e70857d](https://github.com/powsybl/powsybl-open-loadflow/pull/1138/commits/e70857d8154f2c1767a966edec5494a5e7ecd5ab)
- Make equation terms' eval() affine and create intermediary class that can give the equation and compute the flow with rhs() (will dev it soon)

vidaldid-rte: I start to think that there are some confusions in the current equation system object model.

In the current model, an equation term represent a physical quantity.
The physical quantity is (in the current confusing model):   eval() - rhs()
   This is what is used in the eval implementation of an equation that returns the pysical quantity 

But for the DC engine, we must explicitely separate the linear part from the constant part so taht we can rite

TARGET vector - RHS vector = MATRIX*(vector of V and phi)
And this separation is done by using for a term:
   eval as the linear part
   rhs as the constant part

On the other hand, most of the code uses eval of a term to get a physical quantity. But the pysical quantity is, with this model  term.eval - term.rhs

I'm still not sure how to fix that but I would tend to something like;
    eval returns the physical quantity for both equations (as it does today) and the terms
    create an interface for DC terms that force to define the linear part and the constant part
           use this to create the DC target vector and the DC matrix
           and use this for eval that should return the physical quantity

vidaldid-rte: > For the moment i can suggest 3 solutions :
> 
>     * Remove the rhs lines in the Equation::eval()  (reset to [9d5a6ec](https://github.com/powsybl/powsybl-open-loadflow/pull/1138/commits/9d5a6ec04ea4cb85ca75b38589ac22d3af6941dc))
> 
>     * Not count the rhs for Impedant branches in ZeroImpedanceFlows (see [e70857d](https://github.com/powsybl/powsybl-open-loadflow/pull/1138/commits/e70857d8154f2c1767a966edec5494a5e7ecd5ab))
> 
>     * Make equation terms' eval() affine and create intermediary class that can give the equation and compute the flow with rhs() (will dev it soon)

I will try another one:
    implement evallhs for use in the equation solver that wants to have   lhs = rhs + target
    eval that returns the full quantity
I'll let you know if it leads me somewhere. I think that currently the mangement of rhs is inconsistent in many places.

vidaldid-rte: @vmouradian @geofjamg  I created a proposal here https://github.com/powsybl/powsybl-open-loadflow/pull/1147
I think it clairifes the issue with RHS.

vmouradian: > @vmouradian @geofjamg I created a proposal here https://github.com/powsybl/powsybl-open-loadflow/pull/1147
Thank you ! I is indeed a clearer and cleaner way to fix the problem

vidaldid-rte: Thanks. I merged @geofjamg  the code is now here and can be compared with main.

vidaldid-rte (approved):

PR #1142, opened 2024-12-04
Fix busbar section minimal tripping

flo-dup:

No

Bug fix

The tripping corresponding to a bbs contingency with no propagation adds the traversed terminals in the neigbours list.

The tripping corresponding to a bbs contingency with no propagation adds the visited terminals in the traversed terminals set.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/ContingencyTripping.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/ContingencyTripping.java
@@ -78,7 +78,7 @@ public class ContingencyTripping {
         Objects.requireNonNull(network);
         Objects.requireNonNull(bbs);
 
-        NodeBreakerTraverserFactory minimalTraverserFactory = (stoppingSwitches, neighbourTerminals, traversedTerminals, nbv) ->
+        NodeBreakerTraverserFactory minimalTraverserFactory = (stoppingSwitches, traversedTerminals, neighbourTerminals, nbv) ->
             // To have the minimal tripping ("no propagation") with a busbar section we still need to traverse the
             // voltage level starting from that busbar section, stopping at first switch encountered (which will be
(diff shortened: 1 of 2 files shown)


flo-dup:
I still need to add an unit test!

flo-dup:
To me there's no easy way to include a unit test, could we merge it like this? The unit test would be really complex to write for just having swapped two arguments, I think this is not worth the effort.

The original mistake of adding the traversed terminal to the list of neighbouring terminals (to later visit) instead of the set of traversed terminals (to disconnect) has no impact at the end. Indeed, the neighbouring terminals are later visited and added to the set of traversed terminals.

vidaldid-rte:
> To me there's no easy way to include a unit test, could we merge it like this? The unit test would be really complex to write for just having swapped two arguments, I think this is not worth the effort.
> 
> The original mistake of adding the traversed terminal to the list of neighbouring terminals (to later visit) instead of the set of traversed terminals (to disconnect) has no impact at the end. Indeed, the neighbouring terminals are later visited and added to the set of traversed terminals.

I realized that there is no simple test in ContingencyTrippingTest that goes through the line that uses traversedTerminals (nbv.getOptionalTerminal(nodeAfter).ifPresent(traversedTerminals::add);)
I'll try to catch you to understand how to write one and also understand what should be the expected result.

vidaldid-rte:
> I realized that there is no simple test in ContingencyTrippingTest that goes through the line that uses traversedTerminals (nbv.getOptionalTerminal(nodeAfter).ifPresent(traversedTerminals::add);) I'll try to catch you to understand how to write one and also understand what should be the expected result.

Test added. As you said, it worked also on main without your fix. But at least we have a unit test for this tripping that as otherwised only exercised by SA tests.

vidaldid-rte (approved):

PR #1147, opened 2024-12-09
Refacto lhs

### LINKED ISSUE #1137: DC Loadflow: Wrong bus P balance if PST connected.  (equation eval ignores constant terms)

vidaldid-rte:

 This is another approach to fix #1137

Clairify the equation model:

     evallhs = target - rhs
and
    eval = evallhs - rhs

evallhs is used to solve equations
eval is used to get the equation or term quantity

Errors seen with non impedant branches in DC near a PST or an HVDC in AC emulation with P0 non null.

Potental inconsitenscies each time a non null RHS is used

Error fixed

- [ ] Yes
- [ X No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/equations/EquationTerm.java
+++ b/src/main/java/com/powsybl/openloadflow/equations/EquationTerm.java
@@ -186,14 +186,6 @@ public interface EquationTerm<V extends Enum<V> & Quantity, E extends Enum<E> & 
     }
 
     /**
-     * Evaluate the equation term with an alternative state vector.
-     * @return value of the equation term
-     */
-    default double eval(StateVector sv) {
-        throw new UnsupportedOperationException("Not implemented");
-    }
-
-    /**
      * Get partial derivative.
      *
      * @param variable the variable the partial derivative is with respect to
--- a/src/main/java/com/powsybl/openloadflow/equations/EquationTerm.java
+++ b/src/main/java/com/powsybl/openloadflow/equations/EquationTerm.java
@@ -182,7 +182,7 @@ public interface EquationTerm<V extends Enum<V> & Quantity, E extends Enum<E> & 
      * @return value of the equation term
      */
     default double evalLhs() {
-        return eval() + (hasRhs() ? rhs() : 0);
+        return eval() - (hasRhs() ? rhs() : 0);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/main/java/com/powsybl/openloadflow/equations/EquationTerm.java
```diff
@@ -178,11 +178,11 @@ static <V extends Enum<V> & Quantity, E extends Enum<E> & Quantity> EquationTerm
...
 
     /**
-     * Evaluate the equation term with an alternative state vector.
+     * Evaluate equation lhs of the equation term
      * @return value of the equation term
      */
-    default double eval(StateVector sv) {
-        throw new UnsupportedOperationException("Not implemented");
+    default double evalLhs() {
+        return eval() + (hasRhs() ? rhs() : 0);
```
SylvestreSakti, line 185: Should it be - rhs instead of + rhs ?
```suggestion
        return eval() - (hasRhs() ? rhs() : 0);
```

vidaldid-rte: You are right. Done.

PR #1148, opened 2024-12-11
Fix regression slack distribution reports

jeandemanged:

No

Bugfix

We don't have anymore slack distribution failure reports since #1123 

Re-introduce the pre-existing reports. Also some refactoring to easen upcoming #1146 

- [x] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
+++ b/src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
@@ -7,15 +7,17 @@
  */
 package com.powsybl.openloadflow.network.util;
 
+import com.powsybl.commons.PowsyblException;
 import com.powsybl.loadflow.LoadFlowParameters;
+import com.powsybl.openloadflow.OpenLoadFlowParameters;
 import com.powsybl.openloadflow.network.LfBus;
 import com.powsybl.openloadflow.network.LfGenerator;
 import com.powsybl.openloadflow.network.LfNetwork;
+import com.powsybl.openloadflow.util.PerUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -23,6 +25,8 @@ import java.util.stream.Collectors;
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
 public final class ActivePowerDistribution {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ActivePowerDistribution.class);
 
     /**
      * Active power residue epsilon: 10^-5 in p.u => 10^-3 in Mw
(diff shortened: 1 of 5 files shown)


src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
```diff
@@ -104,4 +108,60 @@ public static Step getStep(LoadFlowParameters.BalanceType balanceType, boolean l
                     new GenerationActivePowerDistributionStep(GenerationActivePowerDistributionStep.ParticipationType.REMAINING_MARGIN, useActiveLimits);
         };
     }
+
+    public record ResultAfterFailureBehaviorHandling(boolean failed, String failedMessage, int iteration, double remainingMismatch, boolean movedBuses, double additionalDistributedActivePower) { }
+
+    public static ResultAfterFailureBehaviorHandling handleDistributionFailureBehavior(OpenLoadFlowParameters.SlackDistributionFailureBehavior behavior,
+                                                                                       LfGenerator referenceGenerator,
+                                                                                       double requestedActivePower,
```
SylvestreSakti, line 116: Something like `slackActivePowerInitialMismatch` is more explicit than `requestedActivePower`?

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
```diff
@@ -104,4 +108,60 @@ public static Step getStep(LoadFlowParameters.BalanceType balanceType, boolean l
...
+                                                                                       Result result, String failMessageTemplate) {
+        Objects.requireNonNull(behavior);
+        Objects.requireNonNull(result);
+        ResultAfterFailureBehaviorHandling resultAfterFailureBehaviorHandling;
+        final OpenLoadFlowParameters.SlackDistributionFailureBehavior effectiveBehavior;
+
+        // if requested behavior is to distribute on reference generator, but there is no reference generation, we fall back internally to FAIL mode
+        if (OpenLoadFlowParameters.SlackDistributionFailureBehavior.DISTRIBUTE_ON_REFERENCE_GENERATOR == behavior && referenceGenerator == null) {
+            effectiveBehavior = OpenLoadFlowParameters.SlackDistributionFailureBehavior.FAIL;
```
SylvestreSakti, line 125: A log message may be given to inform that no reference generator is defined.

SylvestreSakti (commented):
Thank you for finding this bug and fixing it ! Should we add a unit test for these report messages to avoid such a regression ?

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
jeandemanged: adding it but proposing DEBUG level, I fear this would be too noisy: it would be issued at each outerloop

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
jeandemanged: changed to `activePowerMismatch` just like in the other `run(...)` method

vidaldid-rte (approved):

Issue #1024, opened 2024-04-29
Support reactive power control of a generator on its own terminal


Hadrien-Godard:
### Describe the current behavior

Currently, if a generator reactive power control is defined on its own terminal, the control is discarded, and following line is logged:
ERROR <TIME> c.p.o.n.i.AbstractLfGenerator | Generator <GEN_NAME> is controlled by an instance of class com.powsybl.iidm.network.impl.GeneratorImpl: not supported

Notice that in IIDM, such a reactive power control is qualified as "remote", although it is local. In CGMES, such a control is not qualified as local or remote (the distinction does not exist), but is converted as a remote one in IIDM.
As "remote" reactive power control is the only way to model reactive power control from a generator in IIDM, we do not consider that this term "remote" prevents us from modeling such a local control in OpenLoadFlow.

### Describe the expected behavior

If a generator remote reactive power control is on its own generator terminal, keep it and do not log an error.
We propose to keep it as a new kind of generator control called local reactive power control.

If parameter usereactivelimits is true:
The initial target Q of the generator should be: max( min(Qmax, controlTargetQ), Qmin).
In the reactive power outer loop, we should take this control into account and re-check the values of Qmin and Qmax (which could change as the generator targetP changes) and so recheck the targetQ of the generator.

Else:
The targetQ of the generator is controlTargetQ.

### Describe the motivation

The motivation is to model reactive power control that are defined at the generator terminal level.
Those controls ensure that the generator follows a target (as the targetQ field does), but here this target could evolve as the Qmin/Qmax bounds on the reactive power output could evolve during the load flow iterations.

Hadrien-Godard:
Pending, since we have to deliberate more on the expected behavior.

PR #1154, opened 2024-12-16
Add a parameter to force targetQ within reactive limits

vidaldid-rte:

No

In some situations group reactive limits are more trusted that group targetQ
This PR adds a parameter **forceTargetQInReactiveLimits** . When this parameter is set to true and a generator's targetQ is outside the limit, the targetQ is set to the nearest possible value in the PQ diagram.

targetQ has more priority than reactive limits

By default, targetQ has more priority than reactive limits. But if you set the **forceTargetQInReactiveLimits** parameter to true, reactiveLimits have more priority than targetQ.

- [ ] Yes
- [X] No


SylvestreSakti:
Thank you for this PR, I added a small doc paragraph for the introduced parameter.

SylvestreSakti (approved):

Hadrien-Godard:
I have two remarks on this PR:
- Why is this parameter linked to "useReactiveLimits"? It would not shock me if both parameters are independant. We might want to test what happens when we prefer reactiveLimits over targetQ but without activating the reactiveLimits outerloop in generator voltage control.
- I am not fully convince by adding a boolean to every generator objects. Maybe we can do better by using directly the global parameter and pass it to getTargetQ().

jeandemanged:
Hi @Hadrien-Godard ,

> I am not fully convince by adding a boolean to every generator objects. Maybe we can do better by using directly the global parameter and pass it to getTargetQ().

I kind of like it, having it at individual generator level would address #1024 from what I understand ? Finalizing fixing #1024 would just require filling this boolean for each LfGenerator from an iIDM extension (to be decided if new extension or existing `RemoteReactivePowerControl`)

Hadrien-Godard:
> Hi @Hadrien-Godard ,
> 
> > I am not fully convince by adding a boolean to every generator objects. Maybe we can do better by using directly the global parameter and pass it to getTargetQ().
> 
> I kind of like it, having it at individual generator level would address #1024 from what I understand ? Finalizing fixing #1024 would just require filling this boolean for each LfGenerator from an iIDM extension (to be decided if new extension or existing `RemoteReactivePowerControl`)

Hello @jeandemanged 
Yes you are right! Although it won't fully solve #1024 because there is the case where the active power produced is modified during the LF (for instance due to slack distribution), and then the minQ/maxQ are modified. But it will definitely help to solve the issue to have the parameter at the generator level. Thanks.

vidaldid-rte:
> I have two remarks on this PR:
> 
>     * Why is this parameter linked to "useReactiveLimits"? It would not shock me if both parameters are independant. We might want to test what happens when we prefer reactiveLimits over targetQ but without activating the reactiveLimits outerloop in generator voltage control.
> 
Ah.. My reasoning was the opposite. If useReactiveLimit is set to false you assume an infinite PQ diagram so you don't want to set a constraint to targetQ. 
This being said my use cases so far always have useReactiveLimit set to true.

jeandemanged:
> > I have two remarks on this PR:
> > ```
> > * Why is this parameter linked to "useReactiveLimits"? It would not shock me if both parameters are independant. We might want to test what happens when we prefer reactiveLimits over targetQ but without activating the reactiveLimits outerloop in generator voltage control.
> > ```
> 
> Ah.. My reasoning was the opposite. If useReactiveLimit is set to false you assume an infinite PQ diagram so you don't want to set a constraint to targetQ. This being said my use cases so far always have useReactiveLimit set to true.

I agree with you. Either reactive limits are considered, or they are not. And this is also clarified in the parameter documentation so fine with me.

jeandemanged:
> Hello @jeandemanged Yes you are right! Although it won't fully solve #1024 because there is the case where the active power produced is modified during the LF (for instance due to slack distribution), and then the minQ/maxQ are modified. But it will definitely help to solve the issue to have the parameter at the generator level. Thanks.

Hi @Hadrien-Godard , I added a test with a capability curve, and from what I can see the case where the active power produced is modified during the LF is managed! :)
See `AcLoadFlowGeneratorTest#testGeneratorForceTargetQInCurveDiagram`.

Hadrien-Godard:
> > Hello @jeandemanged Yes you are right! Although it won't fully solve #1024 because there is the case where the active power produced is modified during the LF (for instance due to slack distribution), and then the minQ/maxQ are modified. But it will definitely help to solve the issue to have the parameter at the generator level. Thanks.
> 
> Hi @Hadrien-Godard , I added a test with a capability curve, and from what I can see the case where the active power produced is modified during the LF is managed! :) See `AcLoadFlowGeneratorTest#testGeneratorForceTargetQInCurveDiagram`.

Ok very nice ! Thanks

Hadrien-Godard:
> > > I have two remarks on this PR:
> > > ```
> > > * Why is this parameter linked to "useReactiveLimits"? It would not shock me if both parameters are independant. We might want to test what happens when we prefer reactiveLimits over targetQ but without activating the reactiveLimits outerloop in generator voltage control.
> > > ```
> > 
> > 
> > Ah.. My reasoning was the opposite. If useReactiveLimit is set to false you assume an infinite PQ diagram so you don't want to set a constraint to targetQ. This being said my use cases so far always have useReactiveLimit set to true.
> 
> I agree with you. Either reactive limits are considered, or they are not. And this is also clarified in the parameter documentation so fine with me.

My idea was to be able to test this feature when reactive limits for generator controlling voltage are deactivated (since we like deactivating it in debuging conditions). But I won't fight both of you :).

Hadrien-Godard (approved):

jeandemanged (approved):

PR #1157, opened 2024-12-17
Fix slack distribution outerloop incorrect stable status

jeandemanged:

No

Bug fix

<del>Regression introduced in the #1148 refactoring.</del>
The root problem is not really the refactoring, actually the code before the refactor was hiding a bug. The real problem is the computation of the `movedBuses` indicator in `ActivePowerDistribution`.
In the following situation:
- the slack power to be distributed is small but above threshold, e.g. 0.15 MW vs. 0.10 MW threshold,
- and there are many elements over which we distribute the mismatch (e.g. 1000 elements)
then each element moves by only 0.000015 MW, below epsilon threshold. Consequently `ActivePowerDistribution` reports that no bus has moved.

Next Consequences:
- DistributedSlackOuterloop returns a stable status,
- no new NR is launched,
- result is not balanced at slack bus

`movedBuses` indicator in `ActivePowerDistribution` correctly accounts for many many small changes, by summing them instead of checking individual values.

- [X] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
+++ b/src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
@@ -86,8 +86,10 @@ public final class ActivePowerDistribution {
             iteration++;
         }
 
-        final boolean movedBuses = Math.abs(initialP.entrySet().stream()
-                .mapToDouble(e -> e.getKey().getTargetP() - e.getValue()).sum()) > P_RESIDUE_EPS;
+        // Identify if injections moved significantly, used e.g. to establish stable/unstable outer loop status.
+        // The 0.9 magic factor is to handle potential rounding issues.
+        final boolean movedBuses = initialP.entrySet().stream()
+                .mapToDouble(e -> Math.abs(e.getKey().getTargetP() - e.getValue())).sum() > P_RESIDUE_EPS * 0.9;
 
         return new Result(iteration, remainingMismatch, movedBuses);
     }
--- a/src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
+++ b/src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
@@ -86,8 +86,8 @@ public final class ActivePowerDistribution {
             iteration++;
         }
 
-        final boolean movedBuses = initialP.entrySet().stream()
-                .anyMatch(e -> Math.abs(e.getKey().getTargetP() - e.getValue()) > P_RESIDUE_EPS);
+        final boolean movedBuses = Math.abs(initialP.entrySet().stream()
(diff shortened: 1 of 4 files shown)


src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
```diff
@@ -86,8 +86,8 @@ public Result run(LfGenerator referenceGenerator, Collection<LfBus> buses, doubl
             iteration++;
         }
 
-        final boolean movedBuses = initialP.entrySet().stream()
-                .anyMatch(e -> Math.abs(e.getKey().getTargetP() - e.getValue()) > P_RESIDUE_EPS);
+        final boolean movedBuses = Math.abs(initialP.entrySet().stream()
+                .mapToDouble(e -> e.getKey().getTargetP() - e.getValue()).sum()) > P_RESIDUE_EPS;
```
vidaldid-rte, line 91: If unlucky (mismatch very close to P_RESIDUE_EPS, and roundings you could be below the criteria)
Maybe > P_RESIDUE_EPS*0.9   to avoid rounding effects ?

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
```diff
@@ -86,8 +86,8 @@ public Result run(LfGenerator referenceGenerator, Collection<LfBus> buses, doubl
...
```
SylvestreSakti, line 90: Isn't better to sum the abs values instead of doing the abs value of the sum ?

jeandemanged: in today's distribution all elements go always in the same direction, therefore we can spare a few cpu cycles with a single Math.abs
But maybe we should indeed put abs on each element, no strong opinion here ...

SylvestreSakti: Oh yes that's true. No strong opinion neither, these cpu cycles may not be perceptible, but it can stay as it is.

vidaldid-rte (approved):

SylvestreSakti (approved):

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
jeandemanged: in the end I change to what you suggested, imagine in the future we do fancy distribution on both loads and generators...

src/main/java/com/powsybl/openloadflow/network/util/ActivePowerDistribution.java
jeandemanged: indeed, thanks

Issue #1091, opened 2024-09-25
SecurityAnalysis: Reconfigure slack distribution outer loops per contingency using an extension


jeandemanged:
### Describe the current behavior

The same outer loops are used for basecase and all contingencies

### Describe the expected behavior

Via a contingency extension, e.g. <del>`ContingencyParametersExtension`</del> `ContingencyLoadFlowParameters`, we can configure which contingencies should use either SlackDistribution or AreaInterchangeControl.
The operator strategies associated to the contingency, if any, would use the same outerloops as configured for the contingency.

For a start we propose the contingency extension to be created in OLF (and not core).

The extension would for a start support choosing:
- between SlackDistribution or AreaInterchangeControl,
- which balanceType should be used (on loads, generators, ...),
but should server as a basis for other options in the future.

### Describe the motivation

Customer request. Depending on the contingency type, either simulate FCR or FRR. (precision: OLF by itself will not check the contingency type, this is done outside OLF when creating/setting up the contingency extension).

PR #1158, opened 2024-12-18
Refactor DC Outer Loops creation

vmouradian:

Part of #1091 + adds consistency (imo)

Feature / refactor

DC Outerloops creation is made more consistent with AC ones (created in the OLF parameters class then are contained in DcLoadFlowParameters)

Support of `outerLoopNames` parameter for DC is in this separated PR : #1159.

- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
@@ -1929,7 +1929,6 @@ public class OpenLoadFlowParameters extends AbstractExtension<LoadFlowParameters
                 .setSlackDistributionFailureBehavior(parametersExt.getSlackDistributionFailureBehavior())
                 .setMatrixFactory(matrixFactory)
                 .setDistributedSlack(parameters.isDistributedSlack())
-                .setAreaInterchangeControl(parametersExt.isAreaInterchangeControl())
                 .setBalanceType(parameters.getBalanceType())
                 .setSetVToNan(true)
                 .setOuterLoops(createDcOuterLoops(parameters, parametersExt))
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
@@ -24,13 +24,17 @@ import com.powsybl.openloadflow.ac.equations.AcEquationSystemCreationParameters;
 import com.powsybl.openloadflow.ac.solver.*;
 import com.powsybl.openloadflow.ac.outerloop.AcOuterLoop;
 import com.powsybl.openloadflow.ac.outerloop.ReactiveLimitsOuterLoop;
+import com.powsybl.openloadflow.dc.DcAreaInterchangeControlOuterLoop;
+import com.powsybl.openloadflow.dc.DcIncrementalPhaseControlOuterLoop;
 import com.powsybl.openloadflow.dc.DcLoadFlowParameters;
+import com.powsybl.openloadflow.dc.DcOuterLoop;
 import com.powsybl.openloadflow.dc.DcValueVoltageInitializer;
 import com.powsybl.openloadflow.dc.equations.DcApproximationType;
 import com.powsybl.openloadflow.dc.equations.DcEquationSystemCreationParameters;
 import com.powsybl.openloadflow.graph.GraphConnectivityFactory;
 import com.powsybl.openloadflow.lf.AbstractLoadFlowParameters;
 import com.powsybl.openloadflow.network.*;
+import com.powsybl.openloadflow.network.util.ActivePowerDistribution;
 import com.powsybl.openloadflow.network.util.PreviousValueVoltageInitializer;
 import com.powsybl.openloadflow.network.util.UniformValueVoltageInitializer;
 import com.powsybl.openloadflow.network.util.VoltageInitializer;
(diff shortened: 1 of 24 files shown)


vmouradian:
Support of OuterLoopNames parameters for DC is in this separated PR : #1159.
Can be merged into this branch or to main later

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
```diff
@@ -184,37 +183,37 @@ public DcLoadFlowResult run() {
...
-
-        if (parameters.getNetworkParameters().isPhaseControl()) {
-            DcIncrementalPhaseControlOuterLoop phaseShifterControlOuterLoop = new DcIncrementalPhaseControlOuterLoop();
-            DcOuterLoopContext phaseShifterControlOuterLoopContext = new DcOuterLoopContext(network);
-            outerLoopsAndContexts.add(Pair.of(phaseShifterControlOuterLoop, phaseShifterControlOuterLoopContext));
-            phaseShifterControlOuterLoop.initialize(phaseShifterControlOuterLoopContext);
+        // remove area interchange control outer loop if no area in the network, active power will be distributed with classical slack distribution
+        boolean areaInterchangeControl = outerLoops.stream().anyMatch(DcAreaInterchangeControlOuterLoop.class::isInstance);
+        if (!network.hasArea()) {
+            outerLoops.removeIf(DcAreaInterchangeControlOuterLoop.class::isInstance);
```
jeandemanged, line 191: don't remove the outerloop here.
the outerloop should handle the 'no area' situation by itself.

src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
```diff
@@ -42,7 +42,7 @@ public abstract class AbstractAreaInterchangeControlOuterLoop<
             C extends LoadFlowContext<V, E, P>,
             O extends AbstractOuterLoopContext<V, E, P, C>>
         extends AbstractActivePowerDistributionOuterLoop<V, E, P, C, O>
-        implements OuterLoop<V, E, P, C, O>, ActivePowerDistributionOuterLoop<V, E, P, C, O> {
+        implements ActivePowerDistributionOuterLoop<V, E, P, C, O> {
```
jeandemanged, line 45: since AbstractActivePowerDistributionOuterLoop implements ActivePowerDistributionOuterLoop, this part is redundant and can be removed too

```java
implements ActivePowerDistributionOuterLoop<V, E, P, C, O>
```

jeandemanged (changes requested):

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
```diff
@@ -1851,6 +1855,20 @@ public static AcLoadFlowParameters createAcParameters(LoadFlowParameters paramet
                 .setSolverFactory(solverFactory, parameters);
     }
 
+    static List<DcOuterLoop> createDcOuterLoops(LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt) {
+        List<DcOuterLoop> outerLoops = new ArrayList<>();
+        if (parameters.isPhaseShifterRegulationOn()) {
+            DcIncrementalPhaseControlOuterLoop phaseShifterControlOuterLoop = new DcIncrementalPhaseControlOuterLoop();
```
SylvestreSakti, line 1861: Hello, a small remark here, maybe a little bit out of the scope of this PR : when comparing with `createAcOuterLoops`, it shows that the parameter **phaseShifterControlMode** is not taken into account in DC mode. Maybe it should be mentioned in the doc (or even with a log message, because default value of this param is `CONTINUOUS_WITH_DISCRETISATION` and not `INCREMENTAL`)

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
vmouradian: This was done to reproduce the behavior of DC slack distribution (which is done before equations solving and not via an outerloop).
So in the existing implementation of the AIC outreloops in the "no area" case :
 - The AC one defaults to distributed slack outerloop
 - The DC one does not handles the case because it should be handled by the DC LF engine (distribute slack before the other computations)  

This is something that can be homogenized but I think it will be out of the scope of this refactoring

src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
vmouradian: yes ! fixed in [929a5b8](https://github.com/powsybl/powsybl-open-loadflow/pull/1158/commits/929a5b85adca901e5823c81ae5eaffb28280f374)

src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameters.java
vidaldid-rte: @SylvestreSakti if we want t improve the doc for phaseShifterControlMode I think this should be done in a different PR.
The doc of this paraeter could be improved in the following way I guess (just from code reading) - that are not all related to this outerloop.
    - The parameter is only relevant for AC
    - in Continuous mode, only the phase shifter in active power controller mode are computed through the Newton Raphson. The current limiters are controlled by an outerloop.
    - In continuous mode, if the active target power taget is impossible to reach, the newton raphson will diverge. In incremental mode, the failure occurs in the incrementalPhaseControl outerloop that cannot reach a stable state. This makes it easier to troubleshoot the failure cause.

vidaldid-rte (approved):
Looks OK for me.

jeandemanged (approved):

src/main/java/com/powsybl/openloadflow/dc/DcLoadFlowEngine.java
jeandemanged: homogenization has been done in this PR

PR #1160, opened 2024-12-20
Fix injections ignored on first slack bus when using multiple slacks

SylvestreSakti:

No

A previous PR (https://github.com/powsybl/powsybl-open-loadflow/pull/1116) fixes injections that are ignored in slack buses when using multiple slacks. But actually, the problem is just partially solved because injections are still ignored when being in **the first** slack. This is due to the distribution equations.

Example with 3 slack buses

Before the previous fix (#1116) we had these distribution equation between slacks :
slack_p2 - slack_p1 = 0
slack_p3 - slack_p1 = 0

After the previous fix (#1116) we had these distribution equation between slacks :
slack_p2 - slack_p1 = target_p2
slack_p3 - slack_p1 = target_p3

Now we have these distribution equation between slacks :
slack_p2 - slack_p1 = target_p2 - target_p1
slack_p3 - slack_p1 = target_p3 - target_p1

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/ac/MultipleSlackBusesTest.java
+++ b/src/test/java/com/powsybl/openloadflow/ac/MultipleSlackBusesTest.java
@@ -79,6 +79,15 @@ class MultipleSlackBusesTest {
         return Stream.concat(acStream, dcStream);
     }
 
+    static Stream<Arguments> allModelAndSwitchingTypes() {
+        return Stream.of(
+                Arguments.of(true, true),
+                Arguments.of(true, false),
+                Arguments.of(false, true),
+                Arguments.of(false, false)
+        );
+    }
+
     static Stream<Arguments> allModelTypes() {
         return Stream.of(Arguments.of(true), Arguments.of(false));
     }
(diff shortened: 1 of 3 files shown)


src/test/java/com/powsybl/openloadflow/ac/MultipleSlackBusesTest.java
```diff
@@ -148,20 +157,26 @@ void nonImpedantBranchTest(boolean ac) {
         assertSlackBusResults(slackBusResults, expectedSlackBusMismatch, 2);
     }
 
-    @ParameterizedTest(name = "ac : {0}")
-    @MethodSource("allModelTypes")
-    void loadOnSlackBusTest(boolean ac) {
+    @ParameterizedTest(name = "ac : {0}, switchSlacks : {1}")
+    @MethodSource("allModelAndSwitchingTypes")
+    void loadOnSlackBusTest(boolean ac, boolean switchSlacks) {
```
vidaldid-rte, line 186: Could you add another test that runs one (or more models) with one slack and without multiple and checks that the solution is close enough ? (bus injections, bus voltage, terminal flows...)

SylvestreSakti: Yes, done a few minutes ago, I'll add more asserts to check bus injections, bus voltages as you mentioned

SylvestreSakti: Done

vidaldid-rte (approved):

Hadrien-Godard (approved):

PR #1161, opened 2024-12-27
[WIP] Returning real slack buses mismatches in SlackBusResult when using multiple slacks

SylvestreSakti:

No

When using multi-slack, the returned slack buses result returned the same mismatch for all the slackbuses which is. This approach can be wrong because the mismatches may not be exactly the same between slack buses depending on the solver criterias. 
Moreover, this led to hide a wrong slack distribution use case (fixed in #1160) where the global mismatch was almost zero but unequally distributed between slack buses (because of the presence of a load in the first slack and wrong equation definition in this case).

This PR proposes to display the real mismatch in each slack bus independently.

Currently, the component result has a SlackBusResult list containing the same mismatch for each slack bus which equals to : SlackBusMismatch/nbSlack.

In the new behavior, each SlackBusResult has a mismatch corresponding to SlackBus.getMismatchP(). 
NB : This introduces a functional change which is the handling of SlackbusFailureBehaviour.FAIL which leads to have the initial mismatch as the displayed mismatch in each bus since no distribution has been done.

A cleaner way to do such a change would be to introduce the breaking change of having SlackBusResult list instead of a SlackBusActivePowerMismatch in the LoadFlowResult interface in order to display only data that is contained in the LoadFlow result (instead of computing each MismatchP in the final network to retreive the correct mismatches) but such a major refactoring is not done in this PR. This can be discussed.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
@@ -191,13 +191,13 @@ public class OpenLoadFlowProvider implements LoadFlowProvider {
     private static ReferenceBusAndSlackBusesResults buildReferenceBusAndSlackBusesResults(AbstractLoadFlowResult result) {
         String referenceBusId = null;
         List<LoadFlowResult.SlackBusResult> slackBusResultList = new ArrayList<>();
-        double slackBusActivePowerMismatch = result.getSlackBusActivePowerMismatch() * PerUnit.SB;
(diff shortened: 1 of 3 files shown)


geofjamg:
> A cleaner way to do such a change would be to introduce the breaking change of having SlackBusResult list instead of a SlackBusActivePowerMismatch in the LoadFlowResult interface in order to display only data that is contained in the LoadFlow result (instead of computing each MismatchP in the final network to retreive the correct mismatches) but such a major refactoring is not done in this PR. This can be discussed.

Yes, it would worse an API evolution to get real mismatches for each of the slacks

PR #1162, opened 2024-12-30
Fix performance issue when accessing loading limits from IIDM

geofjamg:

No

Performance issue

Security analysis profiling shows that we spend a lot of time accessing limit violation from the IIDM network. 

There is an issue with the way LfLimit creation from IIDM limits and caching is done. The IIDM access is done at each call even if LfLimit has already been created and cached.
I guess this issue (already ever present) is more visible since we have added operationnal limits in IIDM because we now have an indirection when accessing an operational limit group (using the current limit group and going though a Map access).
This issus is easily solved by passing a Supplier of IIDM limit.

Example of benchmark on a 6000 buses network and a 3000 N-1 line contingencies security analysis:
before this PR: 64s
with this PR: 52s

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfTieLineBranch.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfTieLineBranch.java
@@ -105,11 +105,11 @@ public class LfTieLineBranch extends AbstractImpedantLfBranch {
     public List<LfLimit> getLimits1(final LimitType type, LimitReductionManager limitReductionManager) {
         switch (type) {
             case ACTIVE_POWER:
-                return getLimits1(type, getHalf1().getActivePowerLimits().orElse(null), limitReductionManager);
+                return getLimits1(type, getHalf1()::getActivePowerLimits, limitReductionManager);
             case APPARENT_POWER:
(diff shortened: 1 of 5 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfTieLineBranch.java
```diff
@@ -120,11 +120,11 @@ public List<LfLimit> getLimits1(final LimitType type, LimitReductionManager limi
     public List<LfLimit> getLimits2(final LimitType type, LimitReductionManager limitReductionManager) {
         switch (type) {
             case ACTIVE_POWER:
-                return getLimits2(type, getHalf2().getActivePowerLimits().orElse(null), limitReductionManager);
-            case APPARENT_POWER:
```
p-arvy, line 124: `CURRENT` and `APPARENT_POWER` cases have been swapped here.

geofjamg: Reverted

p-arvy (approved):
Thanks for the dev @geofjamg.

PR #1163, opened 2025-01-03
Load active power target caching

geofjamg:

No

Perfs improvement

When profiling a DC security analysis we get 2.8% if the time spent in calculating bus active power. Generation part is alreadu cached bu not load part.
image

After caching also the load part, we get twice faster.
image

After using a traditionnal loop instead of a stream, we get 0;
image

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
@@ -386,7 +386,10 @@ public abstract class AbstractLfBus extends AbstractElement implements LfBus {
     @Override
     public double getGenerationTargetP() {
         if (generationTargetP == null) {
-            generationTargetP = generators.stream().mapToDouble(LfGenerator::getTargetP).sum();
+            generationTargetP = 0.0;
+            for (LfGenerator generator : generators) {
+                generationTargetP += generator.getTargetP();
+            }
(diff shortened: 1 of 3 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -405,11 +410,20 @@ public void setGenerationTargetQ(double generationTargetQ) {
         }
     }
 
+    @Override
+    public void invalidateLoadTargetP() {
+        loadTargetP = null;
+    }
+
```
vidaldid-rte, line 417: Shouldn't this also be called by AbstractLfBus.addLoad (or safer LfLoadImpl.add) ?

vidaldid-rte: After discussion, ok to consider the addXXX as builfder methods that don't need to call invalidate. The NetworkCache currently only uses setTargetP for  whas is implemented. So lets assume for now that if Load changes are supported in the future it will ony be through setTargetP and not adding loads or line to a bus.

vidaldid-rte (approved):

Hadrien-Godard (approved):

jeandemanged (approved):

PR #1165, opened 2025-01-06
Fix secondary voltage control in case of generator remote voltage control

geofjamg:

No

Bug fix

Secondary voltage control outer failed in case a zone has generators belonging to a shared remote control.   

All generators of a secondary voltage control, have local voltage control.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/ac/outerloop/SecondaryVoltageControlOuterLoop.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/outerloop/SecondaryVoltageControlOuterLoop.java
@@ -275,8 +275,12 @@ public class SecondaryVoltageControlOuterLoop implements AcOuterLoop {
         // replace last row for each of the secondary voltage control block
         for (LfSecondaryVoltageControl secondaryVoltageControl : secondaryVoltageControls) {
             List<LfBus> controllerBuses = secondaryVoltageControl.getEnabledControllerBuses();
-            LfBus lastControllerBus = controllerBuses.get(controllerBuses.size() - 1);
-            int i = controllerBusIndex.get(lastControllerBus.getNum());
+            List<LfBus> controlledBuses = controllerBuses.stream()
+                    .map(bus -> bus.getGeneratorVoltageControl().orElseThrow().getControlledBus())
+                    .distinct()
+                    .toList();
+            LfBus lastControlledBus = controlledBuses.get(controlledBuses.size() - 1);
+            int i = controlledBusIndex.get(lastControlledBus.getNum());
 
             DenseMatrix jVpp = createJvpp(sensitivityContext, secondaryVoltageControl.getPilotBus(), allControlledBuses, controlledBusIndex);
             for (int j = 0; j < b.getColumnCount(); j++) {
--- a/src/main/java/com/powsybl/openloadflow/ac/outerloop/SecondaryVoltageControlOuterLoop.java
+++ b/src/main/java/com/powsybl/openloadflow/ac/outerloop/SecondaryVoltageControlOuterLoop.java
@@ -21,7 +21,6 @@ import com.powsybl.openloadflow.lf.outerloop.OuterLoopResult;
 import com.powsybl.openloadflow.lf.outerloop.OuterLoopStatus;
 import com.powsybl.openloadflow.network.*;
 import org.apache.commons.lang3.mutable.MutableDouble;
-import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/powsybl/openloadflow/ac/outerloop/SecondaryVoltageControlOuterLoop.java
```diff
@@ -181,35 +178,33 @@ private static DenseMatrix createA(List<LfSecondaryVoltageControl> secondaryVolt
...
         return k0;
     }
 
-    private static DenseMatrix createJk(SensitivityContext sensitivityContext, List<LfBus> controllerBuses, Map<Integer, Integer> controllerBusIndex) {
-        int n = controllerBuses.size();
-        DenseMatrix jK = new DenseMatrix(n, n);
-        for (LfBus controllerBusI : controllerBuses) {
+    private static DenseMatrix createJk(SensitivityContext sensitivityContext,
+                                        List<LfBus> controllerBuses, List<LfBus> controlledBuses,
+                                        Map<Integer, Integer> controllerBusIndex, Map<Integer, Integer> controlledBusIndex) {
```
vidaldid-rte, line 191: very subtle difference between controllerBusIndex and controlledBusIndex
Is it possible to use something with more visual constrast ?
For example controllingBusIndex / controlledBusIndex

src/test/java/com/powsybl/openloadflow/ac/SecondaryVoltageControlTest.java
```diff
@@ -428,4 +425,93 @@ void testWithDiscardedGenerator() {
...
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        assertEquals(LoadFlowResult.ComponentResult.Status.CONVERGED, result.getComponentResults().get(0).getStatus());
+        assertEquals(7, result.getComponentResults().get(0).getIterationCount());
+
+        assertVoltageEquals(12.5, b10);
+        assertVoltageEquals(5.548, b991);
+        assertVoltageEquals(4.93, b992);
+        assertReactivePowerEquals(-9.172, g991.getTerminal());
+        assertReactivePowerEquals(4.742, g992.getTerminal());
+    }
```
vidaldid-rte, line 516: Can you add a test where one of the controlling buses (controll units)  reaches Qmin (or Qmax) and is disabled ?

vidaldid-rte: Can you also add a test with an target V at the pilot that would cause an irrealistic voltage in the control units ?
(ie needs so much negative reactive power that control unit's voltage becomes very low).

src/main/java/com/powsybl/openloadflow/network/GeneratorVoltageControl.java
```diff
@@ -80,4 +81,23 @@ public void updateReactiveKeys() {
...
+    public List<GeneratorVoltageControl> split() {
+        if (isLocalControl()) {
+            return List.of(this);
+        } else {
+            List<GeneratorVoltageControl> generatorVoltageControls = new ArrayList<>(controllerElements.size());
+            // create one (local) generator control per controller bus and remove this one
+            controlledBus.setGeneratorVoltageControl(null);
+            for (LfBus controllerBus : controllerElements) {
+                var generatorVoltageControl = new GeneratorVoltageControl(controllerBus, targetPriority, targetValue);
+                generatorVoltageControl.addControllerElement(controllerBus);
```
vidaldid-rte, line 97: moving targetV pu or remote bus to targetV pu at local bus can change a lot the network configuration if the transformer between local and remote as ratedUi different from nominalVi, which is quite frequent.

src/main/java/com/powsybl/openloadflow/network/GeneratorVoltageControl.java
```diff
@@ -80,4 +81,23 @@ public void updateReactiveKeys() {
             controllerBus.setRemoteControlReactivePercent(reactiveKeysSum == 0 ? 0 : reactiveKeys[i] / reactiveKeysSum);
         }
     }
+
+    /**
+     * Returns itself in case of local control, split into several local voltage controls in case of remote control.
+     */
+    public List<GeneratorVoltageControl> split() {
```
vidaldid-rte, line 88: The method name is misleading. 
It appears as a getter, but in reality the method may transform the control if it is remote.

Maybe something like
toLocalVoltageControls  ?

src/main/java/com/powsybl/openloadflow/network/GeneratorVoltageControl.java
geofjamg: Yes but I don't think we can do better as there is no guaranty to have a initial state in IIDM to have a better init value. And at the end of the computation, it win't change anything because the aim of the secondary control is to move these voltage targets to reach the pilot point voltage target.

src/main/java/com/powsybl/openloadflow/ac/outerloop/SecondaryVoltageControlOuterLoop.java
geofjamg: Maybe we should have used controlling instead of controller but this terms controller/controlled are used everywhere in OLF and I don't think we should use a different naming only in this outerloop.
So this is maybe a topic but not for this PR.

src/main/java/com/powsybl/openloadflow/network/GeneratorVoltageControl.java
vidaldid-rte: But it breaks the argument that secondary voltage controls happens after primary voltage control. Here the primary voltage control might be significantly modified. Can't this change the final convergence state ?

src/test/java/com/powsybl/openloadflow/ac/SecondaryVoltageControlTest.java
geofjamg: Not sure it is relevant in this PR. That is going to fail but not because of something related to secondary voltage control outerloop and will me mostly solved by https://github.com/powsybl/powsybl-open-loadflow/pull/1135

src/main/java/com/powsybl/openloadflow/network/GeneratorVoltageControl.java
geofjamg: Yes, in theory it could. Practically I didn't notice it on my test cases. 
Best implementation would be to only convert to local control after first NR (so when SecondaryVoltageControlOuterLoop is first ran). But...
 - this is really not easy (need to handle a new event in the LFNetworkListener and need to update the equation system to remove remote voltage controls equations and replace by local ones)
 - is it worth it ?

src/test/java/com/powsybl/openloadflow/ac/SecondaryVoltageControlTest.java
vidaldid-rte: I'm closing that one. I didn't realize that remote control was deactivated when I wrote this comment, and that groups act individually.

src/main/java/com/powsybl/openloadflow/ac/outerloop/SecondaryVoltageControlOuterLoop.java
vidaldid-rte: tant pis.

src/main/java/com/powsybl/openloadflow/network/GeneratorVoltageControl.java
vidaldid-rte: Let's keep the algorithm as it is then for now.
I added a test to check that voltage configuration is not modified when the parameter is off.
I also documented the limitations on groups, and explained also that this algorithm is a static approximation.

Do you agree with these changes ?

vidaldid-rte: resolved after discussion

vidaldid-rte (approved):

jeandemanged (approved):
Approved, however note that this is a limited review:
I am not a user of this feature. Therefore I can only assess that code/test/doc quality is good as usual, and that no side effect would occur when `secondaryVoltageControl` is disabled. Regarding fitness for purpose you are on your own here.

PR #1169, opened 2025-01-15
[Draft] Improve fast DC AS performances

annetill:

No, because this PR should never be merged but separated in small PRs. 

No, but the fast DC AS is not as fast as expected. For 3000 branch contingencies in the 400-225 kV french network, it is only 50% faster than the classical DC AS.

image

Many trials, to be discussed with @geofjamg and @p-arvy. We stopped the PR because our first goal was to delete totally network saving and restoration. Because of slack distribution, it could lead to a complex duplication of some existing code (see the update of DC target vector). So we think now that a first quick win could be to create a DC network save/restore with as less as possible fields.

- [ ] Yes
- [ ] No

**Other information**:


src/main/java/com/powsybl/openloadflow/dc/fastdc/WoodburyEngine.java
```diff
@@ -83,6 +81,7 @@ public static double[] runDcLoadFlowWithModifiedTargetVector(DcLoadFlowContext l
 
         DcLoadFlowParameters parameters = loadFlowContext.getParameters();
         if (parameters.isDistributedSlack()) {
+            // FIXME, distribution keys has changed...
```
annetill, line 84: @p-arvy here is why we stop, as slack distribution keys have changed, the trajet vector has to be updated. Doing it by hand is really complex and in a way a code duplication, so it could be nice if you test a DC network save/restore with the less possible number of fields (only buses, some branches, all branches, which attributes inside).

p-arvy: Okay thank you for the feedback @annetill. I will take a look.

p-arvy: Hello @annetill, @geofjamg, my understanding is that we want to return to applying the `LfContingency`, but this time saving/modifying/restoring the least possible number of fields. 

In doing so, I have the impression that it is sufficient to apply the generators/loads/hvdcs changes due to the `LfContingency`, and save/restore the `BusDcState` of the `LfNetwork`, as the impact of changes on other elements (lost buses/PSTs) is already accounted for in the RHS override by directly modifying the target vector array (in method `WoodburyEngine.runDcLoadFlowWithModifiedTargetVector`). What do you think ? 

I have pushed a commit in this regard (note that the code is not yet clean). Please feel free to comment.

p-arvy: I also rolled back to have the case distinction when the contingency causes generator/load losses or not (to avoid having to save/apply/restore the modifications). This reduces significantly execution time in case of branch contingencies.

vidaldid-rte:
@p-arvy @geofjamg @annetill I think that Pierre has integrated most of it in OLF. Can this PR be closed or is there something left to intergate in OLF to improve the fast DC perf and design ?

PR #1170, opened 2025-01-15
Modified exception message when initial LF fails in AC Sensi

SylvestreSakti:

No

Misleading exception message was thrown when load flow fails in AC sensi analysis. A correction is done to this message.

When running an AC sensitivity analysis, if the first load flow fails because of UNSTABLE outer loop status, a PowsyblException is thrown with message : "Load flow ended with status CONVERGED"

A correct exception message is thrown depending on Load flow result.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
@@ -23,6 +23,7 @@ import com.powsybl.openloadflow.ac.equations.AcVariableType;
 import com.powsybl.openloadflow.ac.solver.AcSolverStatus;
 import com.powsybl.openloadflow.ac.solver.AcSolverUtil;
 import com.powsybl.openloadflow.graph.GraphConnectivityFactory;
+import com.powsybl.openloadflow.lf.outerloop.OuterLoopStatus;
 import com.powsybl.openloadflow.network.*;
 import com.powsybl.openloadflow.network.impl.LfNetworkList;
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -152,9 +153,13 @@ private static boolean runLoadFlow(AcLoadFlowContext context, boolean throwsExce
             return true;
         } else {
             if (throwsExceptionIfNoConvergence) {
-                throw new PowsyblException("Load flow ended with status " + result.getSolverStatus());
+                if (result.getOuterLoopResult().status() != OuterLoopStatus.STABLE) {
+                    throw new PowsyblException("Load flow ended with outer loop status " + result.getOuterLoopResult().statusText());
+                } else {
+                    throw new PowsyblException("Load flow ended with solver status " + result.getSolverStatus());
+                }
```
vidaldid-rte, line 160: The message can still be improved by specifying that this is the load flow of the N case
(failures for contingencies don't interrupt the analysis)

For this you can also rename the parameter from throwsExceptionIfNoConvergence to something like runBaseSituation (te clarify its meaning since now the message would depend on this param) , and make the message more precise.

SylvestreSakti: Thanks, done !

vidaldid-rte (approved):

Hadrien-Godard (approved):

PR #1156, opened 2024-12-17
ContingencyLoadFlowParameters extension support in Security Analysis

vmouradian:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

Final part of  https://github.com/powsybl/powsybl-open-loadflow/issues/1091

Since AIC can be activated only for a contingency, LfNetWorkLoader now needs to always converts the areas.

Adds outerLoopNames parameter to ContingencyLoadFlowParameters for consistency, since outerLoopConfig is used for creating outerloops per-contingency.

**What is the new behavior (if this is a feature change)?**
Uses the class ContingencyLoadFlowParameters created in #1139 in SA calculations

Final part of the work on https://github.com/powsybl/powsybl-open-loadflow/issues/1091

**Does this PR introduce a breaking change or deprecate an API?**
- [x] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

OuterLoopConfig Interface has a new method     
List<O> configure(LoadFlowParameters parameters, OpenLoadFlowParameters parametersExt, LoadFlowParametersOverride loadFlowParametersOverride);

It needs to be added for OuterloopConfig implementations. 
If it is not an issue that the implementation does not support custom per-contingency parameters, the new method can just call old configure().

**Other information**:

PR #1172, opened 2025-01-16
LfAction refactoring and area interchange target action support

obrix:

No issue

Add the support of AreaInterchangeTarget action and change the implementation of LfAction.
LfAction existing design was confusing as everything was handled in the unique LfAction object.
A proper class hierarchy has been setup so as the make feature extension and maintenance easier.

All actions implemented in the LfAction object. Created through the create method and applied through the global update connectivity method (for switch and terminal connection actions) and apply method.
Each instance of LfAction was capable of describing all kind of actions through all the available attributes but a single instance was created per iidm action.

All action creation and application processes are implemented in dedicated class, with common code in AbstractLfAction for all actions and additional AbstractLfBranchAction for switch and terminal connection action (they both impact topology).

Applying the action can be done through a call to the apply method or through a call to LfActionUtils.applyListOfActions. This last method, in a similar way that was done before the refactoring, apply a list of action by first applying the actions modifying topology and then the rest of the actions.

- [x] Yes
- [ ] No

**Other information**:


obrix:
Note for the reviewer : There was not specific rule about crashing or outputing a warning when an action cannot be applied so I kept more or less the same behavior for each action. Some of the warning / exception raising have been moved from construction to action application though.

obrix:
To sum up a bit, there is 4 cases of action setup that causes an exception raise : 

•	For TapChangerAction, if the PiModel is a SimpleModel (no tap changer on the transformer) -> throw UnsupportedOperationException
•	For GeneratorAction, if we want to do something else than simply set an active power target (with absolute or relative value) : throw UnsupportedOperationException
•	For HvdcAction, if ac emulation is already enabled -> throw UnsupportedOperationException 
•	For TerminalsConnectionAction, is a side is provided -> throw UnsupportedOperationException (only both side can be opened / closed)

Those exception are the same as before the refactoring. All other potential issue are only outputting a warning, for example a branch not found in the network for a TerminalsConnectionAction.

I also added a global warning for failing to apply an action in LfActionUtils, which might lead to a lot of warning when doing a security analysis on multiple component. So not sure we should keep that.

src/test/java/com/powsybl/openloadflow/sa/LfActionTest.java
```diff
@@ -172,12 +189,69 @@ void testHvdcAction2() {
...
+        }
+
+        // With area interchange target control disabled
+        acParameters.getNetworkParameters().setAreaInterchangeControl(false);
+        try (LfNetworkList lfNetworks = Networks.load(network, acParameters.getNetworkParameters(), new LfTopoConfig(), ReportNode.NO_OP)) {
+            LfNetwork lfNetwork = lfNetworks.getLargest().orElseThrow();
+            acParameters.getNetworkParameters().setAreaInterchangeControl(false);
+
+            LfAction lfAreaTargetAction = LfActionUtils.createLfAction(targetAction, network, acParameters.getNetworkParameters().isBreakers(), lfNetwork);
+            assertFalse(lfAreaTargetAction.apply(lfNetwork, null, acParameters.getNetworkParameters()));
```
m-guibert, line 254: The behavior should change very soon, after #1156 's merge.

Because of the ContingencyLoadFlowParameters extension, AIC can be performed on some post-contingency cases even if networkParameters.isAreaInterchangeControl() is false. So Area-s will now always be created, and this assertion will fail

src/main/java/com/powsybl/openloadflow/network/action/LfActionUtils.java
```diff
@@ -0,0 +1,100 @@
...
+ * @author Jean-Luc Bouchot {@literal <jlbouchot at gmail.com>}
+ */
+public final class LfActionUtils {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LfActionUtils.class);
+
+    private LfActionUtils() {
+    }
+
+    public static LfAction createLfAction(Action action, Network network, boolean breakers, LfNetwork lfNetwork) {
```
m-guibert, line 34: I added the Breaking change tag, as your refactor changed the API.

Add a note about this change and how to migrate in the PR description

src/test/java/com/powsybl/openloadflow/sa/LfActionTest.java
obrix: Thanks I will update on if 1156 is merged before this one.

src/main/java/com/powsybl/openloadflow/network/action/LfActionUtils.java
obrix: Not sure about considering it a breaking change, I think it is adapted when changing a public API or the interpretation of an existing input. 
This PR is only an internal change of implementation within OLF, iidm Action is the public API and it stays the same.

vidaldid-rte, line 34: Why don(t you return an optional ? The previous code could help detect NOOP actions for the network by returning an empty value if the element did not exist.
You now loose the ability to detect actions that do nothing on a given LfNetwork.

obrix: The old code was returning an optional but actually did not do anything with the "empty" information and just silently ignored the invalid action.
Same checks are done in the apply method with a boolean return and a warning is logged (to be discussed maybe).
Ultimately I think the operator strategy result API should hold the info of what action was applied successfully, this is also why I designed it like this, but this a change of API to be seen in the futur..

obrix: In the current code, even if the list of lfAction converted from iidm actions is empty there is always a load flow done and an operator strategy result is produced (see AbstractSecurityAnalysis.runSimulations usage of lfActionById).
Whether that is a good thing is probably a good question but there is no functionnal change here as far as I can see.

Furthermore we still have the checkActions method that do some additional checks and raise an exception if there is some inconsistency. This one is still there.

Ok for putting the missing action in the report, good idea!

vidaldid-rte: Oh... You a right ! So let-s consider the fact that the LF is run even when there is no action something not ideal, but independant from this work....

I'll close that thread when the report is there.

src/main/java/com/powsybl/openloadflow/network/action/LfHvdcAction.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LfHvdcAction.class);
+
+    public LfHvdcAction(String id, HvdcAction action) {
+        super(id, action);
+    }
+
+    @Override
+    public boolean apply(LfNetwork network, LfContingency contingency, LfNetworkParameters networkParameters) {
+        LfHvdc lfHvdc = network.getHvdcById(action.getHvdcId());
```
vidaldid-rte, line 33: I'm adding here a question about the architecture change (not only specific on HVDC).
This change is not purely about architecture.
It changes also at what time static verifications (independant from the contingence) are done.
Before:
   - Where done at action creation time (once per SA run). In case of issue the action was silently ignored (now they are reproted and it is better)
With this PR:
   - Verifications are performed for each contingence that needs them (innefficient - or can the action validity depend on the contingence ?)
   - But the result in case of inconsitency is displayed in context of the contingence via a report (good)

Can't we get the best of both world by checking the action validity at creation time, and generating the report at apply time ?

src/main/java/com/powsybl/openloadflow/network/action/LfGeneratorAction.java
```diff
@@ -0,0 +1,50 @@
...
+                    double change = activePowerValue.getAsDouble() / PerUnit.SB;
+                    double newTargetP = Boolean.TRUE.equals(relativeValue.get()) ? generator.getTargetP() + change : change;
+                    generator.setTargetP(newTargetP);
+                    generator.setInitialTargetP(newTargetP);
+                    generator.reApplyActivePowerControlChecks(networkParameters, null);
+                    return true;
+                }
+            } else {
+                throw new UnsupportedOperationException("Generator action on " + action.getGeneratorId() + " : configuration not supported yet.");
+            }
```
vidaldid-rte, line 46: By checking this at apply time and not at creation time, you loose the fail fast property that existed before.

Before:
   If there was an issue it was raised at launch time. User could see it immediately.
Witht his PR:
   FOr a long computation, the issue may happen hours after launch time/
    => User looses time
    => The user may not see the issue if lost in a large log.

obrix: Indeed this can probably be moved at construction time. I will do it.

src/main/java/com/powsybl/openloadflow/network/action/LfHvdcAction.java
obrix: Good questions !

I can think of some scenario where the action application validity could depend on the contingency :

* With the work currently being done on load flow parameters specific to some contingencies : might impact the succesful application of an action.
* With some work planned on Artelys side to take into account the equivalent of "auto" action / SPS from a RAO that might be dependent on a type of contingency and then impact differently a same action done in different N-1 scenario. Maybe that's a bit early to care about that though but thats a possibility. 
* More simply an action trying to impact an equipment that was indirectly impacted by a specific contingency

I think most checks / actions application are very fast (a lookup + eventually a boolean check + a call to one or more setter). I am not too worried about that at the moment.
Some of the checks like null branch, null area could probably be removed from the apply method because they are redondant with the checkActions function though.

Maye be some checks can definitely be moved at construction time like you suggested for the generator. I'll look at it again.

src/main/java/com/powsybl/openloadflow/network/action/LfHvdcAction.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    @Override
+    public boolean apply(LfNetwork network, LfContingency contingency, LfNetworkParameters networkParameters) {
+        LfHvdc lfHvdc = network.getHvdcById(action.getHvdcId());
+        if (lfHvdc != null) {
+            if (action.isAcEmulationEnabled().isEmpty()) {
+                LOGGER.warn("Hvdc action {}: only explicitly disabling ac emulation is supported.", action.getId());
+                return false;
```
vidaldid-rte, line 42: Wouldn't a report be better ? (It is the kind of information a user needs to know)

src/main/java/com/powsybl/openloadflow/network/action/AbstractLfBranchAction.java
```diff
@@ -0,0 +1,116 @@
...
+        // re-update connectivity according to post contingency state (revert after LfContingency apply)
+        connectivity.startTemporaryChanges();
+        if (contingency != null) {
+            contingency.getDisabledNetwork().getBranches().forEach(connectivity::removeEdge);
+        }
+
+        // update connectivity according to post action state
+        connectivity.startTemporaryChanges();
+
+        applyOnConnectivity(network, connectivity);
```
vidaldid-rte, line 68: Why not calling directly updateConnectivity ? the apply seems to just add redundant calls to findEnabledDisabledBranches

src/main/java/com/powsybl/openloadflow/network/action/AbstractLfBranchAction.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    abstract boolean findEnabledDisabledBranches(LfNetwork lfNetwork);
+
+    /**
+     * Standalone apply
+     */
+    @Override
+    public boolean apply(LfNetwork network, LfContingency contingency, LfNetworkParameters networkParameters) {
+        boolean found = findEnabledDisabledBranches(network);
```
vidaldid-rte, line 56: if found is false why not returning immediatly ?

src/main/java/com/powsybl/openloadflow/network/action/LfActionUtils.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public static void applyListOfActions(List<LfAction> actions, LfNetwork network, LfContingency contingency, LfNetworkParameters networkParameters, ReportNode node) {
+        Objects.requireNonNull(actions);
+        Objects.requireNonNull(network);
+
+        // first apply action modifying connectivity
+        List<LfAction> branchActions = actions.stream()
+            .filter(action -> action instanceof AbstractLfBranchAction<?>)
+            .toList();
+        updateConnectivity(branchActions, network, contingency, node);
```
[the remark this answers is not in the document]
vidaldid-rte: Can we return to the previous solution where:
    - connectivity is touched only once
    - the precontingency state is restored only once
    - all connectivity changes are surrounded by a  connectivity.startTemporaryChanges()/  connectivity.undoTemporaryChanges()

src/main/java/com/powsybl/openloadflow/network/action/LfGeneratorAction.java
```diff
@@ -0,0 +1,56 @@
...
+        super(id, action);
+        LfGenerator generator = lfNetwork.getGeneratorById(action.getGeneratorId());
+        if (generator != null) {
+            OptionalDouble activePowerValue = action.getActivePowerValue();
+            Optional<Boolean> relativeValue = action.isActivePowerRelativeValue();
+            if (relativeValue.isEmpty() || activePowerValue.isEmpty()) {
+                throw new UnsupportedOperationException("Generator action on " + action.getGeneratorId() + " : configuration not supported yet.");
+            } else {
+                generatorChange = new GeneratorChange(activePowerValue.getAsDouble(), relativeValue.get());
+            }
```
vidaldid-rte, line 39: Why do you put in the GeneratorChange a quantity that is not in per unit as it was done previously  ?
LfNetwork objects normally handle only per unit values.

src/main/java/com/powsybl/openloadflow/network/action/LfGeneratorAction.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    @Override
+    public boolean apply(LfNetwork network, LfContingency contingency, LfNetworkParameters networkParameters) {
+        LfGenerator generator = network.getGeneratorById(action.getGeneratorId());
+        if (generator != null && !generator.isDisabled()) {
+            double change = generatorChange.targetP() / PerUnit.SB;
```
vidaldid-rte, line 47: The per unit conversion should be done when the GeneratorChange object is created.

src/main/java/com/powsybl/openloadflow/network/action/AbstractLfBranchAction.java
obrix: Indeed we can get out early

src/main/java/com/powsybl/openloadflow/network/action/AbstractLfBranchAction.java
obrix: No reason indeed, probably forgot to update it while iterating

src/main/java/com/powsybl/openloadflow/network/action/LfHvdcAction.java
vidaldid-rte: Also, this log is an improvement over the previous code that kept this error silent. Thanks !

src/main/java/com/powsybl/openloadflow/network/action/LfGeneratorAction.java
obrix: Because I reintroduced generator change only as an internal storage of value to do all the checks in the constructor like we discussed. Yes the per unit conversion can be done where we want, that's purely internal.

src/main/java/com/powsybl/openloadflow/network/action/LfGeneratorAction.java
obrix: See other comment

src/main/java/com/powsybl/openloadflow/network/action/LfGeneratorAction.java
obrix: Move it at construction

src/main/java/com/powsybl/openloadflow/network/action/LfShuntCompensatorPositionAction.java
```diff
@@ -0,0 +1,44 @@
...
+    public boolean apply(LfNetwork network, LfContingency contingency, LfNetworkParameters networkParameters) {
+        LfShunt shunt = network.getShuntById(action.getShuntCompensatorId());
+        if (shunt instanceof LfShuntImpl) { // no svc here
+            if (shunt.getVoltageControl().isPresent()) {
+                LOGGER.warn("Shunt compensator position action: voltage control is present on the shunt, section could be overridden.");
+            }
+            shunt.getControllers().stream().filter(controller -> controller.getId().equals(action.getShuntCompensatorId())).findAny()
+                .ifPresentOrElse(controller -> controller.updateSectionB(action.getSectionCount()),
+                    () -> LOGGER.warn("No section change: shunt {} not present", action.getShuntCompensatorId()));
+            return true;
```
vidaldid-rte, line 40: You return true even if the shunt compensator is not found ?

src/main/java/com/powsybl/openloadflow/network/action/LfActionUtils.java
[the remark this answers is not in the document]
vidaldid-rte: I see now. Thanks !

src/main/java/com/powsybl/openloadflow/network/action/LfHvdcAction.java
obrix: I am not sure the original behaviod for this was really thought out but yes at least there is a warning now.
For the report that may be a good idea indeed, but I would have to update all the action to be coherent and report everything probably. I would prefer to do it in another PR if that's okay for you.

PR #1176, opened 2025-01-30
Adapt to change of LoadFlowProvider API + fix update to not write default values multiple times

HugoKulesza:

No

Feature : 
- adapting to powsybl-core PR : https://github.com/powsybl/powsybl-core/pull/3288
- changing params load from PlatformConfig to only change value present and not override everything with defaults 

- [x] No

- [ ] Yes
- [x] No


src/test/java/com/powsybl/openloadflow/OpenLoadFlowParametersTest.java
```diff
@@ -285,6 +285,22 @@ void testAlwaysUpdateNetwork() {
...
+        assertEquals(SlackDistributionFailureBehavior.LEAVE_ON_SLACK_BUS, olfParameters.getSlackDistributionFailureBehavior());
+
+        MapModuleConfig olfModuleConfig = platformConfig.createModuleConfig(MODULE_SPECIFIC_PARAMETERS);
+        olfModuleConfig.setStringProperty(SLACK_BUS_SELECTION_MODE_PARAM_NAME, SlackBusSelectionMode.FIRST.toString());
+        olfParameters.update(platformConfig);
+
+        assertEquals(SlackBusSelectionMode.FIRST, olfParameters.getSlackBusSelectionMode());
+        assertEquals(SlackDistributionFailureBehavior.LEAVE_ON_SLACK_BUS, olfParameters.getSlackDistributionFailureBehavior());
+    }
+
```
vidaldid-rte, line 303: @HugoKulesza  can you add an end to end test with LoadFlowDefaultParametersLoaderMock that would load a resource file that contains both loadFlow parameter defaults and OpenLoadFLowParameter defaults, and verify that the parameters are loaded as expected ?
If it is easy to add a InMemoryPlatformConfig it would be good to test also the priority order as done in core.

vidaldid-rte (approved):

PR #1178, opened 2025-02-03
Slack distribution participation factors: use maxP instead of maxTargetP

SylvestreSakti:

No

Fix of the formula used to compute participation factor in the specific case of using maxTargetP in ActivePowerControl extension (that was introduced in #1043). It is prefered that this participation factor does not depend on maxTargetP (it is rather prefered to use maxP)

- [x] No

When using maxTargetP in ActivePowerControl :
- In BalanceType.PROPORTIONAL_TO_GENERATION_PMAX : $participation factor = maxTargetP/droop$
- In BalanceType.PROPORTIONAL_TO_GENERATION_REMAINING_MARGIN : $participation factor = max(0, maxTargetP - targetP)$

When using maxTargetP in ActivePowerControl, maxTargetP is not taken into account to compute the participation factor :
- In BalanceType.PROPORTIONAL_TO_GENERATION_PMAX : $participation factor = maxP/droop$
- In BalanceType.PROPORTIONAL_TO_GENERATION_REMAINING_MARGIN : $participation factor = max(0, maxP - targetP)$

- [ ] Yes
- [x] No


src/test/java/com/powsybl/openloadflow/ac/DistributedSlackOnGenerationTest.java
```diff
@@ -140,6 +140,19 @@ void testProportionalToPWithTargetLimit() {
         assertActivePowerEquals(-74.605, g4.getTerminal());
     }
 
+    @Test
+    void testProportionalToPMaxWithTargetLimit() {
```
jeandemanged, line 144: the test demonstrates that the active power limit from ActivePowerControl extension is correctly taken into account, but it does not demonstrate the change in participation factors computation. Checking the limit is already done in `testProportionalToPWithTargetLimit`, so this test seems redundant to me compared to existing.
A test where the generator is not hitting the limit should be done instead.

jeandemanged (changes requested):
I am ok with the change, but the added test does not demonstrate it.

src/test/java/com/powsybl/openloadflow/ac/DistributedSlackOnGenerationTest.java
SylvestreSakti: I agree, the goal of the test was unclear. It seems redundant but there was no test case with the balance type PROPORTIONAL_TO_GENERATION_PMAX. 

So I extended the test with one assert where the limit is not hitted and one assert where the limit is hit.

So that we can check that :
- maxTargetP does limit the generator participation
- maxTargetP does not influence the participation factor

jeandemanged (approved):

vidaldid-rte (approved):

PR #1179, opened 2025-02-04
Improve log and report when max outerloop iterations reached

vidaldid-rte:

- [X ] The commit message follows our guidelines

No

Better logging and reporting of failure condition if max outerloop iterations is reached

When a LF fails because the maximum number of outerloop is reached the log message is unclear.
The logs contains an output like this. Although there is more detailed in the componentResults status text,
users looking at the logs or reports do not know the cause of the problem and how to fix it and may ask for
support.

INFO  c.p.openloadflow.ac.AcloadFlowEngine - Ac loadflow complete on network {CC0 SC0} (result=AcLoadFlowResult(outerLoopIterations=20, newtonRaphsonIterations=22, solverStatus=CONVERGED, outerLoopStatus=UNSTABLE, slackBusActivePowerMismatch=-0.026383757113385142, distributedActivePower=-72.65332551565376))

An error log is added that says that the maximum number of outerloop is reached.
The information is also added to the report.

- [ ] Yes
- [X] No


SylvestreSakti (commented):
This is indeed better with this report message ! Can the same message be sent for DcLoadFlowEngine ?

vidaldid-rte:
> This is indeed better with this report message ! Can the same message be sent for DcLoadFlowEngine ?

Done.

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -319,6 +323,17 @@ public static void reportUnsuccessfulOuterLoop(ReportNode reportNode, String out
...
     }
 
+    public static void reportMaxOuterLoopIterations(ReportNode reportNode, int iterationCount, boolean withLog) {
+        ReportNode added = reportNode.newReportNode()
+                .withMessageTemplate("reportMaxOuterLoopIterations", "Maximum number of outerloop iterations reached: ${outerLoopIterationCount}")
+                .withUntypedValue("outerLoopIterationCount", iterationCount)
+                .withSeverity(TypedValue.ERROR_SEVERITY)
+                .add();
+        if (withLog) {
+            LOGGER.error(added.getMessage());
```
SylvestreSakti, line 333: Isn'it better to get the LOGGER as a parameter ? Otherwise the log message will be related to the class `Reports`instead of `AcLoadFlowEngine` or `DcLoadFlowEngine`

vidaldid-rte: Indeed. Done.

SylvestreSakti (approved):

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -319,6 +320,17 @@ public static void reportUnsuccessfulOuterLoop(ReportNode reportNode, String out
                 .add();
     }
 
+    public static void reportMaxOuterLoopIterations(ReportNode reportNode, int iterationCount, boolean withLog, Logger logger) {
+        ReportNode added = reportNode.newReportNode()
+                .withMessageTemplate("reportMaxOuterLoopIterations", "Maximum number of outerloop iterations reached: ${outerLoopIterationCount}")
```
jeandemanged, line 325: minor: change message key to just `"maxOuterLoopIterations"` ?

vidaldid-rte: Done. Thanks. It had not realized this was the convention to name report keys.

jeandemanged (approved):

PR #1188, opened 2025-02-10
Move up unrealistic voltage check from NR to AC loadflow engine

vidaldid-rte:

- [X ] The commit message follows our guidelines

No

PR #1135 highlighted the fact that unrealistic voltage check is node in the Newton Raphson only.
It was decided during review to move the check up to the AC loadflow engine so that solvers remain agnostic to the physics of equations to be solved, and also that the check benefits to all AC solvers.

Unrealistic voltage are detecred only by the Newton Raphson Solver.

Unrealistic voltage are detecred for any solver.

- [ ] Yes
- [X] No

Unrealistic voltages may be detected for other AC solvers than Newton Raphson.


SylvestreSakti:
I'am ok with the changes (I don't know it is is really impacting for other AcSolvers but at least it will be indeed mentioned in the migration guide). Just a little detail in the doc may be changed, as these parameters currently mention the Newton-Raphson solver but now concern every solvers.

SylvestreSakti (approved):

geofjamg (approved):
Ok, good design improvement. Jus to keep in mind that a perfect clean design would remove the UNREALISTIC_STATE from AcSolverStatus which is part of the AcSolver interface so son now not related to any unrealistic state check and use a more specific status for ACLoadFlowEngine adding this  UNREALISTIC_STATE value.

vidaldid-rte:
> Ok, good design improvement. Jus to keep in mind that a perfect clean design would remove the UNREALISTIC_STATE from AcSolverStatus which is part of the AcSolver interface so son now not related to any unrealistic state check and use a more specific status for ACLoadFlowEngine adding this UNREALISTIC_STATE value.

I looked into this and did not find an easy way since the status is used to determine if the resolution of the AC problems succeded.
This being said, since the Solver is an ACSOlver, not a mere NonLinearSolver, and knows a bit about the physics (it takes a
 VoltageIInitializer, I'm not really choked about this UNREALISTIC_STATE status in the interface. I take this PR more as reuse of a shared check that a speration of conepts.

PR #1191, opened 2025-02-14
for limit reduction, duration criteria should depend on next limit's duration

vidaldid-rte:

No

Limit reduction based on acceptable duration criteria should be applied based on the duration of the next limit.

Usually, the acceptable duration displayed in security analysis is the duration of the next limit.
When a reduction is applied to all limits with an acceptable duration of 5 minutes or less,
the expectation is that he reduction is applied also to the previous limit (with a duration just higher than 5 minutes) 
so that the operator is warned in advance.

Reduction is applied to the limit with duration in range of the criteria. But this has no effect in term of detected the violation of the previous limit earlier.

- Reduction is applied to the limit with duration in range of the criteria, but also to the previous limit.

- In case two reductions appli to the same limit, the stronger reduction is used
  (previously, the reduction declared last was applied)

- [ ] Yes
- [X] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
@@ -1,5 +1,5 @@
-/**
- * Copyright (c) 2019, RTE (http://www.rte-france.com)
+/*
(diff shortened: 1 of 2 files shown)


vidaldid-rte:
I put it in WIP time to verify a few things.

PR #1192, opened 2025-02-18
Refactor connectivity break analysis to add connectivity loss impact of the contingencies

p-arvy:

No, but the development is based on the collaborative work of #1169 , aiming to improve the performance of the fast DC SA.
Note that this PR does not directly improve the performance of the fast DC SA but lays the groundwork for it while also improving code quality.

The connectivity loss calculation is directly integrated in the class `ConnectivityBreakAnalysis`, allowing to:
- Improve code quality;
- Refactor the case distinction between contingencies that break connectivity or not in the class `WoodburyDcSecurityAnalysis`;
- Align the results of the fast DC SA with those of an equivalent load flow (in which the contingencies were applied manually) in the case of a HVDC line without power after the application of a contingency.

- [x] No

Connectivity results are redundantly obtained in the `WoodburyDcSecurityAnalysis` class. When the `LfContingency` object is computed for each contingency (as the SA progresses), the `ContingencyConnectivityLossImpact` is calculated through a connectivity analysis.

The connectivity results (`ContingencyConnectivityLossImpact`) are integrated in the `ConnectivityBreakAnalysis`.

This approach allows the use of post-contingency connectivity to determine HVDC lines that end up without power, thus aligning the results of the fast DC SA with those of an equivalent load flow (where contingencies were applied manually).

Additionally, since `ConnectivityAnalysisResult` objects are used for both cases where a contingency either breaks connectivity or not, there is no longer a need to differentiate between these cases using separate method calls to compute the post-contingency state vector.

- [ ] Yes
- [x] No

**Other information**:


Hadrien-Godard (dismissed):

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -4095,6 +4095,65 @@ void testDcSaHvdcLineContingency(boolean dcFastMode) {
         assertEquals(network.getLine("l56").getTerminal1().getP(), postContingencyResult.getNetworkResult().getBranchResult("l56").getP1(), LoadFlowAssert.DELTA_POWER);
     }
 
+    @Test
+    void testFastDcSaHvdcLineWithPowerAfterContingency() {
+        Network network = HvdcNetworkFactory.createNetworkWithGenerators2();
```
vidaldid-rte, line 4100: This test  fails if not in fast DC mode. Why is the result different ?

vidaldid-rte: FYI, it looks like a bug in DC. It's not a regression in your PR and it exists at least since 1.12.2. I'll investigate a bit.

Hadrien-Godard: In my understanding, the standard behavior in this case was changed in AC but not in DC. This PR aligns the fastDC behavior with the AC behavior but does not fix the wrong DC behavior.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -4095,6 +4095,65 @@ void testDcSaHvdcLineContingency(boolean dcFastMode) {
...
+        assertEquals(1, postContingencyResult.getConnectivityResult().getCreatedSynchronousComponentCount());
+        assertEquals(3, postContingencyResult.getNetworkResult().getBranchResults().size()); // only branches in main synchronous component
+        assertEquals(network.getLine("l12").getTerminal1().getP(), postContingencyResult.getNetworkResult().getBranchResult("l12").getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getLine("l13").getTerminal1().getP(), postContingencyResult.getNetworkResult().getBranchResult("l13").getP1(), LoadFlowAssert.DELTA_POWER);
+        assertEquals(network.getLine("l23").getTerminal1().getP(), postContingencyResult.getNetworkResult().getBranchResult("l23").getP1(), LoadFlowAssert.DELTA_POWER);
+    }
+
+    @Test
+    void testFastDcSaHvdcLineWithoutPowerAfterContingency() {
+        Network network = HvdcNetworkFactory.createWithHvdcInAcEmulation();
```
vidaldid-rte, line 4128: For the log, this test also highlight a bug in "normal AS DC". Probably the same root cause in the initial compensation.

src/main/java/com/powsybl/openloadflow/dc/fastdc/ConnectivityBreakAnalysis.java
```diff
@@ -36,61 +33,75 @@ public final class ConnectivityBreakAnalysis {
...
-        private ConnectivityAnalysisResult(Set<String> elementsToReconnect,
-                                           GraphConnectivity<LfBus, LfBranch> connectivity,
-                                           LfNetwork lfNetwork) {
-            this.elementsToReconnect = elementsToReconnect;
-            slackConnectedComponent = connectivity.getConnectedComponent(lfNetwork.getSlackBus());
-            disabledBuses = connectivity.getVerticesRemovedFromMainComponent();
-            partialDisabledBranches = connectivity.getEdgesRemovedFromMainComponent();
+        public ConnectivityAnalysisResult(PropagatedContingency propagatedContingency, LfNetwork network) {
+            this(propagatedContingency, network, Collections.emptySet(), new DisabledElements(Collections.emptySet(), Collections.emptySet(), Collections.emptySet()), Collections.emptySet(), 0);
         }
```
vidaldid-rte, line 54: naming suggestions for code clarity:
   rename the argument from propagatedContingency to nonBreakingConnectivityContingency
   create a contant NO_DISABLED_ELEMENTS = new DisabledElements(Collections.emptySet(), Collections.emptySet(), Collections.emptySet()), Collections.emptySet(), 0)  and use it in the constructor.

The would help to more easily understand the purpose of this constructor.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -380,10 +380,21 @@ private Map<LfBranch, DisabledBranchStatus> findBranchToOpenDirectlyImpactedByCo
         return branchesToOpen;
     }
 
-    record ContingencyConnectivityLossImpact(boolean ok, int createdSynchronousComponents, Set<LfBus> busesToLost, Set<LfHvdc> hvdcsWithoutPower) {
+    public record ContingencyConnectivityLossImpact(boolean ok, int createdSynchronousComponents, Set<LfBus> busesToLost, Set<LfHvdc> hvdcsWithoutPower) {
     }
```
vidaldid-rte, line 384: Why do you need the "ok" field ? It seems to always be set to true.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
```diff
@@ -440,17 +445,21 @@ private static boolean isConnectedAfterContingencySide2(Map<LfBranch, DisabledBr
...
+        ContingencyConnectivityLossImpact run(LfNetwork network, String contingencyId, Map<LfBranch, DisabledBranchStatus> branchesToOpen, boolean relocateSlackBus);
+    }
+
     public Optional<LfContingency> toLfContingency(LfNetwork network) {
-        return toLfContingency(network, true);
+        return toLfContingency(network, true, PropagatedContingency::findBusesAndBranchesImpactedBecauseOfConnectivityLoss);
     }
 
-    public Optional<LfContingency> toLfContingency(LfNetwork network, boolean relocateSlackBus) {
+    public Optional<LfContingency> toLfContingency(LfNetwork network, boolean relocateSlackBus, ContingencyConnectivityLossImpactAnalysis analysis) {
```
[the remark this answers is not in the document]
vidaldid-rte: You probably need to distinguish separated buses from lost buses.

vidaldid-rte (changes requested):
Promising refactoring.
I think their is however a bug with HVDC loss detection (see coments).
Some other comments.

src/main/java/com/powsybl/openloadflow/dc/fastdc/ConnectivityBreakAnalysis.java
p-arvy: Good point. Thank you for the suggestion. It's modified.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
Hadrien-Godard: Removed. Thank you.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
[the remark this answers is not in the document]
Hadrien-Godard: From what I understand, the LFContingency object contains two attributes related to "lost" HVDCs. The first one is "disabledNetwork" that contains a set of "lost" HVDCs. The other is "hvdcsWithoutPower" which also contains a set of "lost" HVDCs.

The HVDCs in disabledNetwork are disabled. Disabling an HVDC disables the AC emulation mode, but does not modify the flow. That is why it is triggered when one bus around the line is in a separated component (because the AC emulation needs both buses to be in the same component).
The HVDCs in hvdcsWithoutPower see their flow set to 0. It is triggered when one bus is totally lost. So this is a more restrictive set of HVDCs. That is why the two sets are computed differently.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
```diff
@@ -291,65 +244,63 @@ private OperatorStrategyResult computeOperatorStrategyResultFromPostContingencyA
...
-        return operatorStrategyResult;
-    }
-
     /**
      * Add the post contingency and operator strategy results, associated to given contingency, in the given list of results.
      * The post contingency and post operator strategy states are computed with given supplier and function.
      */
-    private void addPostContingencyAndOperatorStrategyResults(DcLoadFlowContext context, PropagatedContingency contingency, Map<String, List<OperatorStrategy>> operatorStrategiesByContingencyId,
-                                                              Map<String, LfAction> lfActionById, Supplier<double[]> toPostContingencyStates, Function<List<LfAction>, double[]> toPostContingencyAndOperatorStrategyStates,
+    private void addPostContingencyAndOperatorStrategyResults(DcLoadFlowContext context, ConnectivityAnalysisResult connectivityAnalysisResult, Map<String, List<OperatorStrategy>> operatorStrategiesByContingencyId,
```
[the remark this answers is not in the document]
Hadrien-Godard: I have suggested something in the last commit encapsulated some arguments in different records.

src/main/java/com/powsybl/openloadflow/network/impl/PropagatedContingency.java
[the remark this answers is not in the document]
vidaldid-rte: @Hadrien-Godard  For regular HVDC (not in AC emulation) whose ends get separated by a contingency the behaviour is correct. I added a test to verify this.

For AC emulation, the behaviour is inconsistent in OLF. And I'm not sure it is realistic in either case.
    - If you simulate a loadflow (with intializaiton previous values) the HVDC jumps from previous equilibrium to the active set point set in the IIDM file (and that can be very far from the flow before contingency)
    - If you run an AC security analysis the HVDC is brutally cut off.

I'll track that issue separately.

If you agree with my test, we can close this comment.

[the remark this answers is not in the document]
p-arvy: @vidaldid-rte I agree with your test. Let's close this comment.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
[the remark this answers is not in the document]
p-arvy: Thanks for the remark @vidaldid-rte. I made a proposal mixing both versions:
- A `WoodburyContext` record that contains the shared objects for the component security analysis (including `DcLoadFlowContext`), except `ToStates` which remains a separate record.
- A `SecurityAnalysisSimulationResults` record that contains all security analysis results (both pre-contingency and post-contingency). I removed the `ConnectivityAnalysisResult` object, which seems independent to me.

Feel free to comment. If this works for you, we can close this comment.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
vidaldid-rte, line 4100: This test should be run of fastDC and normal DC once https://github.com/powsybl/powsybl-open-loadflow/pull/1209 is merged.

src/main/java/com/powsybl/openloadflow/sa/WoodburyDcSecurityAnalysis.java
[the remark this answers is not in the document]
vidaldid-rte: Thanks !

Hadrien-Godard (approved):

PR #1200, opened 2025-03-07
Add missing log message for generators discarded from active power control because targetP < minP

SylvestreSakti:

No

Add a missing log message

Generators discarded from active power control for targetP < minP are counted but the count is never logged

The count is logged

- [ ] Yes
- [x] No


src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -1018,6 +1018,10 @@ private LfNetwork create(int numCC, int numSC, Network network, List<Bus> buses,
             LOGGER.warn("Network {}: {} generators have been discarded from active power control because of a targetP > maxP",
                     lfNetwork, report.generatorsDiscardedFromActivePowerControlBecauseTargetPGreaterThanMaxP);
         }
+        if (report.generatorsDiscardedFromActivePowerControlBecauseTargetPLowerThanMinP > 0) {
+            LOGGER.warn("Network {}: {} generators have been discarded from active power control because of a targetP < minP",
+                    lfNetwork, report.generatorsDiscardedFromActivePowerControlBecauseTargetPLowerThanMinP);
+        }
```
vidaldid-rte, line 1024: Now that this log is displayed it show an issue.
Generators with targetP=0 (or close to 0) appear in the category of stopped generators, but also in this category (if minP > 0).
The criteria that increments generatorsDiscardedFromActivePowerControlBecauseTargetPLowerThanMinP should probably exclude stopped generators.

vidaldid-rte (changes requested):
Before displaying the log, we should exclude the stop generators from this list of starting generators.

SylvestreSakti:
It was also the case with all the other checks, e.g. generators stopped but having targetP > maxTargetP were counted twice, generators stopped but having maxTargetP=minTargetP were counted twice, etc.

It's now fixed

vidaldid-rte (approved):

jeandemanged (approved):

PR #1153, opened 2024-12-13
Add switch actions in Fast DC Security Analysis

p-arvy:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

**What kind of change does this PR introduce?**

**What is the current behavior?**

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

PR #1209, opened 2025-03-24
fix double slack distribution in SA DC

vidaldid-rte:

No

In DC the active power mismatch is computer on all bus whether they are enabled or disabled.
This leads to wrong computations in Security Analysis if the lost network contains unbalanced loads and generators.

**Note** that in general the slack distribution is a loss of time in DC security analysis because it was initially made extactly based on the contingency propagation. So the DC engine performs a useless iteration on all buses in SA.

Slack distribution in DC security analysis (default mode) is long each time a unbalanced part of the network is lost.

Slack distribution in DC security analysis is correct.

- [ ] Yes
- [X] No


Hadrien-Godard (approved):

p-arvy (approved):
Thanks for the bugfix @vidaldid-rte. 
I confirm this corrects the TUs in PR #1192. 
It also fixes some TUs added in PR #1153, adding the switching actions in fast DC SA. For some of them, there were differences between a classic DC SA and an equivalent DC LF.

PR #1213, opened 2025-03-27
Remove network results and violation detection on disabled branches on one side in fast DC Security Analysis

p-arvy:

No.

This PR aims to remove network results and limit violation detection in the fast DC SA, for branches that have been disabled on one side due to a contingency.

Currently, if branches are disabled only on one side (for example, in the case of a contingency on a busbar section), the fast DC SA will compute network results and may detect limit violations on these branches.

This behavior differs from the slow DC SA, as the latter uses a listener to replace `ClosedBranchSide1/2DcFlowEquationTerm` with `EvaluableConstants.ZERO` when one side of the line is disabled. As a result, branch results are set to 0 on both sides, and no limits are violated (**although they are still analyzed**).

In fast DC SA, a customized filter is now applied before computing the network results or checking for limit violations. This filter is inspired by the work done in PR #1169.

A branch where one side is disabled by a contingency is considered disabled, meaning no computation or verification is performed on these branches. This differs from the slow DC SA (see the previous paragraph), where the filter simply checks whether the branches have been disabled on both sides by the contingency.

- [ ] Yes
- [x] No

**Other information**:


src/main/java/com/powsybl/openloadflow/dc/fastdc/ConnectivityBreakAnalysis.java
```diff
@@ -142,10 +142,10 @@ private static List<ConnectivityAnalysisResult> computeConnectivityData(LfNetwor
 
         GraphConnectivity<LfBus, LfBranch> connectivity = lfNetwork.getConnectivity();
         for (PropagatedContingency contingency : potentiallyBreakingConnectivityContingencies) {
-            List<ComputedContingencyElement> breakingConnectivityCandidates = contingency.getBranchIdsToOpen().keySet().stream().map(contingencyElementByBranch::get).collect(Collectors.toList());
+            List<ComputedContingencyElement> breakingConnectivityCandidates = contingency.getBranchIdsToOpen().keySet().stream().sorted().map(contingencyElementByBranch::get).collect(Collectors.toList());
```
p-arvy, line 145: When multiple elements can restore connectivity, the traversal order of the `Set`, which may vary between runs, causes the `elementsToReconnect` to change.

Depending on the selected elements, the post-contingent angles of nodes disconnected from the main connected component (which should therefore not be used—this is what this PR aims to address) may vary. As a result, they are not deterministic. Applying a sort and changing the `Set` to a `LinkedHashSet` resolves this "issue".

Hadrien-Godard (dismissed):

src/main/java/com/powsybl/openloadflow/dc/fastdc/ConnectivityBreakAnalysis.java
vidaldid-rte: This info is valuable as a comment in the code.

vidaldid-rte (approved):
Just one minor remark about a comment (about sorting) that should be in the code to explain the purpose and importance of sorting.

src/main/java/com/powsybl/openloadflow/dc/fastdc/ConnectivityBreakAnalysis.java
p-arvy: Done. Thanks.

vidaldid-rte (approved):

PR #1215, opened 2025-04-03
Active Power Distribution: better reports severities when residual mismatch remains

SylvestreSakti:

- [ ] A PR or issue has been opened in all impacted repositories (if any) <- No impacted repositories

No

Reporting slack bus mismatch is slightly changed, to avoid confusion for users :

Whenever there is a slack mismatch after distribution, even if it is residual, a report is sent with ERROR severity

- When slack mismatch remains, the report showing this slack distribution failure is sent only if the mismatch exceeds `slackBusPMaxMismatch`.
- The severity of this report is now set to WARN, because there are cases where remaining mismatch happen without a failing load flow.
- A new report with DEBUG severity is sent when slack distribution fails but the mismatch is residual (lower than `slackBusPMaxMismatch`)

- Severity of the corresponding reports for AreaInterchangeControl have also been modified from ERROR to WARN

- [ ] Yes
- [x] No


vidaldid-rte (dismissed):

vmouradian:
Hello, we also have these report methods for AIC : https://github.com/powsybl/powsybl-open-loadflow/blob/c8b3a560c0bece9d1ce917c2bf026f912f10ccd1/src/main/java/com/powsybl/openloadflow/util/Reports.java#L195-L209  

But they are used in case of a mismatch higher than the threshold, not the epsilon. We could maybe also add a report for residual mismatch here but I am not sure it is very necessary, do you have an opinion on that ?

vmouradian:
> @vmouradian The main driver for this PR was a report with Error severity inside a successulf loadflow computation.
Error severity should only be used for something that causes a failure. I see that the reports you show have an ERROR severity. Do they really correspond to unrecovarable fatal errors ? If not, you should at least change the severity to warning.

You are rigth, it is very similar to the situation when `reportMismatchDistributionFailure` is used. So it would need to be a warning too.

SylvestreSakti:
Hello @vmouradian , yes as @vidaldid-rte says, the main aim of this PR was to have the adequate level of severity for the report. We also found it counfising to compare with `P_RESIDUE_EPS`, and that's why the report is changed to be compared with the `slackBusPMaxMismatch`. To keep this previously existing information though, I added the debug report in case.

I don't know if it is necessary to have the same information for the usecase of AIC. If you are not sure, maybe we can leave it as it is. I just changed the severity level of the report.

vidaldid-rte (approved):

vidaldid-rte (dismissed):

jeandemanged (dismissed):

SylvestreSakti:
I suggest that we wait for #1216 (report internationalization) to be merged before merging this PR.

SylvestreSakti:
Done ! This PR can be reviewed and merged

jeandemanged (approved):

Issue #1206, opened 2025-03-20
[ReportNode] Enable internationalization


alicecaron:
### Describe the current behavior

`ReportNode` use the deprecated `withMessageTemplate` method that do not allow internationalization of the messages.

### Describe the expected behavior

`ReportNode` use the new `withMessageTemplate` method and owns its reports dictionaries to allow internationalization of the messages in en_US and fr_FR locales.

### Describe the motivation

Add internationalization feature of reports + avoid use of powsybl-core deprecated methods.

PR #1216, opened 2025-04-03
Reports i18n

alicecaron:

#1206 

Feature: allow internationalization of reports.
Refacto: remove usages of deprecated `ReportNode::withMessageTemplate(String key, String messageTemplate)` methode and use the new one instead, loading a new ResourceBundle for open loadflow translations.

It is not possible to have report in another language.

It is possible to have a report in a given locale language.

- [X] Yes
- [ ] No

The third version of the report node API from powsybl-core introduces a mechanism for internationalizing messages (i18n).
For all users, when creating a new ReportNode root, the i18n bundles should be loaded. You can either:
- Use the new ReportNode.newRootReportNode().withResourceBundles(String... bundleBaseNames) method, if you want to use specific bundles. To use this open load flow resource bundle, the bundle base name to use is `PowsyblOpenLoadFlowReportResourceBundle.BASE_NAME`
- Or call ReportNode.newRootReportNode().withAllResourceBundlesFromClasspath(), if you want to use all the bundles present in your classpath.


src/test/resources/saMtReport.txt
```diff
@@ -1,4 +1,4 @@
-+ Test Report Node
++ TEST Report Node
```
SylvestreSakti, line 1: Hello, i am ok with this PR ! Just a small detail, I don't understand why this report had to be changed

alicecaron: Hi! It is because in this test class there were 2 ReportNode with the same key "TEST" but not the same template:
`.withMessageTemplate("TEST", "TEST Report Node")` and `.withMessageTemplate("TEST", "Test Report Node")` and when I created the test reports.properties I had to chose one of those translations. Duplicate keys are not recommended from the beginning but it was possible before, not now with the ResourceBundle

SylvestreSakti: Ok thanks, I understand !

SylvestreSakti (approved):

vidaldid-rte:
@alicecaron why is this PR a breaking change ? What does it break for OpenLoadFlow users if we merge this PR  ? Can you detail ? The explenations you provided are about the already done change in Core.

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -616,14 +613,15 @@ public static void reportAcLfComplete(ReportNode reportNode, boolean success, St
...
-        return reportNode.newReportNode().withMessageTemplate("loadFlow", "Load flow on network '${networkId}'")
+        return reportNode.newReportNode().withMessageTemplate("olf.loadFlow")
                 .withUntypedValue(NETWORK_ID, networkId)
                 .add();
     }
 
     public static ReportNode createRootLfNetworkReportNode(int networkNumCc, int networkNumSc) {
         return ReportNode.newRootReportNode()
-                .withMessageTemplate(LF_NETWORK_KEY, "Network CC${networkNumCc} SC${networkNumSc}")
+                .withResourceBundles(PowsyblOpenLoadFlowReportResourceBundle.BASE_NAME)
```
vidaldid-rte, line 623: The root report of LFNetwork is used in prcesses:
  that may call methods from core. 
  that may call methods from openLoadFlow extensions (custom solver, custom outerloops etc...)

This bundle is not sufficient.

src/main/java/com/powsybl/openloadflow/util/Reports.java
```diff
@@ -874,9 +873,16 @@ public static void reportContingencyActivePowerLossDistribution(ReportNode repor
...
-                .withMessageTemplate("LfActionUtils", "Action '${actionId}': may not have been applied successfully on contingency '${contingencyId}'")
+                .withMessageTemplate("olf.LfActionUtils")
                 .withUntypedValue(ACTION_ID, actionId)
                 .withUntypedValue(CONTINGENCY_ID, contingencyId)
                 .add();
     }
+
+    public static ReportNode createThreadRootReport() {
+        return ReportNode.newRootReportNode()
+                .withResourceBundles(PowsyblOpenLoadFlowReportResourceBundle.BASE_NAME)
```
vidaldid-rte, line 884: This bundle is not sufficient.
   As the function is called you don't know if a thread will be limited to OLF calls
   And in the current usage of the method it is for security analysis that may call OLF extensions or core.

src/main/resources/com/powsybl/openloadflow/reports_en_US.properties
```diff
@@ -0,0 +1 @@
+
```
vidaldid-rte, line 1: Do we really need this file ? Is the reports.properties not enough ?

src/test/java/com/powsybl/openloadflow/dc/DcLoadFlowTest.java
```diff
@@ -495,7 +497,10 @@ void outerLoopMaxTotalIterationTest() throws IOException {
         parametersExt.setAreaInterchangePMaxMismatch(1)
                 .setMaxOuterLoopIterations(1);
 
-        ReportNode reportNodeWithLimit1 = ReportNode.newRootReportNode().withMessageTemplate("test", "test").build();
+        ReportNode reportNodeWithLimit1 = ReportNode.newRootReportNode()
+                .withResourceBundles(PowsyblOpenLoadFlowReportResourceBundle.BASE_NAME, PowsyblCoreTestReportResourceBundle.TEST_BASE_NAME)
+                .withMessageTemplate("test")
+                .build();
```
vidaldid-rte, line 504: General question for all the tests that check reports.
How do you ensure that the build is stable and not dependant on the Locale settings of the computer running the tests ?

alicecaron:
> @alicecaron why is this PR a breaking change ? What does it break for OpenLoadFlow users if we merge this PR ? Can you detail ? The explenations you provided are about the already done change in Core.

For users using OpenLoadFlow and creating rootReportNode on their side, they now need to specified the OLF report bundle if they list them one by one to be able to have the messages of the OLF reports in their rootReportNode children generated within OLF.

src/test/java/com/powsybl/openloadflow/dc/DcLoadFlowTest.java
alicecaron: When no locale is given when creating a root ReportNode via the `ReportNode::newRootReportNode` method, the builder `ReportNodeRootBuilderImpl` uses the `ReportConstants.DEFAULT_LOCALE` which is `Locale.US` (en_US). So the tests by default will be translated in en_US locale if no locale is specified, it does not depend on the configuration of the computer running the tests.

vidaldid-rte: > When no locale is given when creating a root ReportNode via the `ReportNode::newRootReportNode` method, the builder `ReportNodeRootBuilderImpl` uses the `ReportConstants.DEFAULT_LOCALE` which is `Locale.US` (en_US). So the tests by default will be translated in en_US locale if no locale is specified, it does not depend on the configuration of the computer running the tests.

Really ? Then it triggers another question. If the root Reports created by OpenLoadFlow are then added to a report with Lcoal french, will the messages be displayed in French ? Can you please add a Unit Test that checks this ?

src/main/resources/com/powsybl/openloadflow/reports_fr_FR.properties
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
alicecaron: It is to separate French from France french that wil be used for France translations. In any case we can create the files only when we add translation keys I can delete them as long as there is no translation done

src/main/resources/com/powsybl/openloadflow/reports_en_US.properties
alicecaron: to be re discussed as we may change the default locale mecanism in core, but for now, in powsybl-core, en_US is the default locale. We need the file to avoid switching on the JVM locale fallback if the file is not present.

src/main/resources/com/powsybl/openloadflow/reports_fr_FR.properties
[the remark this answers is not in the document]
vidaldid-rte: Yes. What would be good is the minimum number of files + one test translation in french and a unit test to verify the mechanic works.

src/test/java/com/powsybl/openloadflow/dc/DcLoadFlowTest.java
vidaldid-rte: Worked around now that root the Locale is added to the created report nodes.

vidaldid-rte (approved):
Thanks for this migration ! Let's start with that as decided and improve as we go.

jeandemanged (approved):

SylvestreSakti (approved):

PR #1220, opened 2025-04-08
Filter disabled generators when computing bus Q limits

vidaldid-rte:

No

When a generator attached to a bus is disabled, exclude it from Q limit computations.

Disabled generators are counted in Q limits.

Disabled generators are excluded from Q limits.

- [ ] Yes
- [X] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
@@ -451,6 +451,7 @@ public abstract class AbstractLfBus extends AbstractElement implements LfBus {
 
     private double getLimitQ(ToDoubleFunction<LfGenerator> limitQ) {
         return generators.stream()
+                .filter(g -> !g.isDisabled())
                 .mapToDouble(generator -> (generator.getGeneratorControlType() == LfGenerator.GeneratorControlType.VOLTAGE ||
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -451,6 +451,7 @@ public double getMaxP() {
 
     private double getLimitQ(ToDoubleFunction<LfGenerator> limitQ) {
         return generators.stream()
+                .filter(g -> !g.isDisabled())
```
SylvestreSakti, line 454: Should we apply the same filter for `generationTargetP` and `generationTargetQ` ?

vidaldid-rte: Good question, although if these quantities were wrong we would have noticed :+1: 

Here is what happens (and it's not as homogeneous as it could be...) 

Here the protection is necessary because the disconnected groupe is removed from tension control and its targetQ might be added to the limit.

Fot generationTargetP, the mecanism is different. The contingency triggers a setTargetP to the group that invalidates the bus generationTargetP. So no need.

Slightly different for generationTargetQ - le quantity is directly updated by the contingence.

SylvestreSakti (approved):

jeandemanged (approved):

SylvestreSakti (approved):

PR #1222, opened 2025-04-11
Change log levels for post-contingency and action messages in Fast DC Security Analysis

p-arvy:

No, but the development is based on the collaborative work of #1169, aiming to improve the performance of the fast DC SA.

In default and fast DC security analyses, the logs indicating the start and end of a post-contingency or post-operator strategy simulation are at the `INFO` level. For the fast DC SA, these logs are changed to the `DEBUG` level.

Currently, the logs are displayed via method calls defined in the `AbstractSecurityAnalysis` class, which log at the `INFO` level. This represents about 5% of the CPU time for a fast DC SA run, and slows down the total execution time.
2-per-cent-log
3-per-cent-log

Since the computation cost is roughly the same for all contingencies in fast DC SA, these logs should be displayed at a higher level of detail.

The logging methods in the `AbstractSecurityAnalysis` class have been modified so that the log level is configurable. In the fast DC SA, the log level is now set to `DEBUG`, which makes the logs cleaner and improves performances.

The `DEBUG` level was chosen to keep the logs consistent with those related to active power distribution, when overriding pre-contingency states in the fast DC SA (the following logs are from the unit test `WoodburyDcSecurityAnalysisWithActionsTest.testSaDcPhaseTapChangerTapPositionChangeWithConnectivityBreak`).

image

- [ ] Yes
- [x] No

To enable the display of detailed logs for post-contingency and post-operator strategy simulations in fast DC Security Analysis, you can set the `com.powsybl.openloadflow.sa.AbstractSecurityAnalysis` logger to DEBUG in your logging configuration file. For better performance, it is recommended to avoid enabling these logs. It should be used only when needed for debugging or detailed analysis.

**Other information**:


Hadrien-Godard (dismissed):

Hadrien-Godard (approved):

docs/security/parameters.md
```diff
@@ -52,7 +52,11 @@ The default value is 1.
 
 **dcFastMode**  
 The `dcFastMode` property allows to use fast DC security analysis, based on Woodbury's formula for calculating post-contingency states, 
-when DC mode is activated.
+when DC mode is activated. 
+
+Setting the `com.powsybl.openloadflow.sa.AbstractSecurityAnalysis` logger to DEBUG in your logging configuration file enables the display of detailed logs
```
vidaldid-rte, line 57: After more thought @p-arvy I wonder if this comment on log should be here. It is really low level and implementation dependent. What about moving that in the PR description (at the end where breaking changes are described) and adding that in the Release notes as well ?

vidaldid-rte (approved):

Hadrien-Godard (approved):

Issue #1219, opened 2025-04-04
Remaining slack mismatch in AIC


vmouradian:
### Describe the current behavior

When using AIC outerloop, a LF can be succesful, but slack mismatch at the end can be more than the selected threshold if this one is low.

### Describe the expected behavior

Slack bus mismatch should be below the threshold if the loadflow is sucessful.

### Describe the steps

Example : 
add `network.getLine("l23_A1").setX(0);`
after creating the network in` AreaInterchangeControlTest#twoAreasWithUnpairedDanglingLine`

PR #1223, opened 2025-04-11
Fix slack mismatch remaining after AIC

vmouradian:

[#1219](https://github.com/powsybl/powsybl-open-loadflow/issues/1219)

Bug Fix

AIC : after balancing all areas, some slack mismatch can remain
It is split and distributed equally among areas.
If the mismatch is very small then we can have mismatch / areas_count < epsilon_P and then it is not distributed at all

remaining mismatch is distributed among all buses in one run of active power distribution


vidaldid-rte (approved):
I am not familiar with area interchange.
I approve on the basis that the modification is local to this algorithm and should not create any regression, and the modification follows the project coding standard.

SylvestreSakti (approved):

PR #1225, opened 2025-04-15
Remove useless ReportNode locale report and set tests locale default to stabilize report node tests

alicecaron:

No

Clean tests related to reports, set the US locale for the tests to stabilize them regardless of the machine environment running the tests. Remove useless test reports_en.properties (according to the core MR that will now set the default locale as the report default  locale (see [core#3406](https://github.com/powsybl/powsybl-core/pull/3406))).

The tests about reports are sensible to the default locale and need an empty reports_en.properties to be successfull with the message template translation.

The tests on reports always set the locale US to ensure the tests will run successfully regardless the default locale of the machine running the tests.

- [ ] Yes
- [x] No

**Other information**:


src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
```diff
@@ -187,21 +187,25 @@ private static boolean switchPqPv(List<PqToPvBus> pqToPvBuses, ContextData conte
 
             int pvPqSwitchCount = contextData.getPvPqSwitchCount(controllerBus.getId());
             if (pvPqSwitchCount >= maxPqPvSwitch) {
-                pqPvNodes.add(Reports.reportPvPqSwitchLimit(controllerBus, pvPqSwitchCount, log, LOGGER));
+                pqPvNodes.add(Reports.reportPvPqSwitchLimit(reportNode, controllerBus, pvPqSwitchCount, log, LOGGER));
             } else {
```
vidaldid-rte, line 191: Now that the methode takes a report as argument it is not clear from its name whether it
    - Creates a single root report node
    - Adds a node to an existing report


Can you plase rename it - something like createRootReportPvPqSwitchLimit

Same for all similar cases....

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowEurostagTutorialExample1Test.java
```diff
@@ -55,6 +58,7 @@ class AcLoadFlowEurostagTutorialExample1Test {
 
     @BeforeEach
     void setUp() {
+        Locale.setDefault(Locale.US);
```
vidaldid-rte, line 61: What is the rule for adding this in a test ?
Is there a reason to run tests, by default, with the machine locale ?
If not, why not adding this statement in the maven configuration for running tests so that all tests run with the US locale ?

src/main/java/com/powsybl/openloadflow/ac/outerloop/ReactiveLimitsOuterLoop.java
alicecaron: you're right, done

src/test/java/com/powsybl/openloadflow/ac/AcLoadFlowEurostagTutorialExample1Test.java
vidaldid-rte: I would not do that personnaly in the code. I prefer the settings in the POM.

   - If you modify the static variable in the code, like the Locale, you have to clean it at test end, or you may have side effects for other tests using the same VM.
   - If you do this for all tests taht compare report strings how do you ensure on the long run that each time a test is modified or created, the rule is still enforced ?
   
   I prefer the follwing:
      - tests run by default with a deterministic local (the modif you did in the pom)
      - tests that explicitely need to test something different form the defaut Locale modify the Locale during the test and restore the defaut Local when they are done.

geofjamg: Agree, this is a very bad practice, creating a side effect on other tests. We must never change a global variable in a test.

vidaldid-rte (approved):

jeandemanged (approved):

PR #1228, opened 2025-04-25
Sensitivity Analysis : Select tie line factor with its dangling line id

SylvestreSakti:

No

When computing sensitivity factor of a tie line, we want to be able to select it by its dangling line id

When a dangling line id is specified, the corresponding LfDanglingLineBranch is selected (only if the dangling line is not paired)

When a dangling line id is specified, if the dangling line is paired (and its studied side is the side 1 corresponding to the network side), the corresponding LfTieLineBranch is selected

- [ ] Yes
- [x] No


vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -775,6 +775,24 @@ protected void checkLoadFlowParameters(LoadFlowParameters lfParameters) {
         }
     }
 
+    private Pair<SensitivityFunctionType, String> checkAndUpdateFunctionTypeDanglingLine(Network network, SensitivityFunctionType functionType, String functionId) {
+        if (isFlowFunction(functionType) || isReactivePowerFunctionType(functionType)) {
```
Hadrien-Godard, line 779: If I am right, for this test to be false, you should request for a sensitivity on a bus voltage, with the same Id as a dangling line.
However this test is significant because it prevents the unwanted side effect that you can have in this "weird" case.
So, I suggest to add a unit test.
This will also cover the last line of this method.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -775,6 +775,24 @@ protected void checkLoadFlowParameters(LoadFlowParameters lfParameters) {
...
+            DanglingLine danglingLine = network.getDanglingLine(functionId);
+            if (danglingLine != null && danglingLine.isPaired()) {
```
Hadrien-Godard, line 781: This one cannot be false, as it is tested before calling this method.

vidaldid-rte: @Hadrien-Godard  this is true. But only in the context of the current call graph. Robust and maintainable  API design should be indeependant of the call graph at a moment in time, and it is good practice to check assumptions in the function.
So although the Sonar robot complains this is not tested, I am for keeping this. It improves code quality.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
SylvestreSakti: I did not think about this case, and indeed, it could have been a weird one ! But when trying to implement the unit test, actually it is impossible (because when trying to add a sensitivity with a bus voltage function, the reader automatically selects its bus as the functionId, and it is impossible to have a network that has a dangling line and a bus with the same id).

I wanted to keep these conditions in case other function types are implemented in the future, and that would be unsupported in this case. Again, it is for maintanibility but it has the drawback of reducing sonar's coverage.

Hadrien-Godard (approved):

PR #1239, opened 2025-06-06
Adapt to the removal of PST FIXED_TAP regulation mode and the new PST…

### LINKED ISSUE #1202: Adapt to the new possible Powsybl PhaseTapChanger regulation modes

A `PhaseTapChanger` can have a `FIXED_TAP` `RegulationMode` regardless of its `regulating` value.

A `PhaseTapChanger` `FIXED_TAP` `RegulationMode` no longer exists. Instead the `regulating` boolean should be set to false to represent the old fixed tap mode.  
So `regulating` attribute must be taken into account to determine if phase tap changer can change position during a loadflow regardless its `RegulationMode` which must be set in any case as one of the following ones: `CURRENT_LIMITER`, `ACTIVE_POWER_CONTROL`.

The simulation of phase shifter control should not modify the tap position of phase tap changers if the new phase tap changer field `loadTapChangingCapabilities` is set to false (meaning the phase tap changer does not have load tap changing capabilities).

### Describe the motivation

Adapt to the core issues:
- [powsybl-core#3344](https://github.com/powsybl/powsybl-core/issues/3344) issue that removes the `FIXED_TAP` regulation mode.
- [powsybl-core#3438](https://github.com/powsybl/powsybl-core/pull/3438) add the new `loadTapChangingCapabilities` attribute to phase tap changer.

### Extra Information

_No response_

alicecaron:
… LoadTapChangingCapabilities flag to decide whether to modify tap position during a loadflow

Fixes #1202 

- Phase Tap changers do not have a `loadTapChangingCapabilities` attribute. Only Ratio Tap changers have a `loadTapChangingCapabilities` attribute.
- The `RatioTapChanger.regulating` status and `RatioTapChanger.loadTapChangingCapabilities` statuses are independent and need to be both `true` to be allowed to regulate in OLF.

- Both phase and ratio tap changers have a `loadTapChangingCapabilities` attribute, indicating if tap position can be changed.
- The `TapChanger.regulating` status only is considered to check if the tap changer is allowed to regulate in OLF (if regulating=true).

- [x] Yes

For ratio tap changers if `loadTapChangingCapabilities` is set to false then it no longer impacts the behaviour of the ratio tap changer capacity of regulation. Only the `regulating` attribute valu is taken into account to allow the tap changer to change tap position in OLF.


src/main/java/com/powsybl/openloadflow/network/impl/LfBranchImpl.java
```diff
@@ -125,9 +125,7 @@ private static LfBranchImpl createTransformer(TwoWindingsTransformer twt, LfNetw
         double zb = PerUnit.zb(nominalV2);
 
         PhaseTapChanger ptc = twt.getPhaseTapChanger();
-        if (ptc != null
-                && (ptc.isRegulating()
-                && ptc.getRegulationMode() != PhaseTapChanger.RegulationMode.FIXED_TAP || retainPtc)) {
+        if (ptc != null && (ptc.isRegulating() && ptc.hasLoadTapChangingCapabilities() || retainPtc)) {
             // we have a phase control, whatever we also have a voltage control or not, we create a pi model array
```
vidaldid-rte, line 129: Is it authorized to have regulating = true and hasLoadTapChangingCapabilities = false ?
  What does it mean then ?
  Shouldn''t it be an unauthorized state ? And then it simplifies code that just needs to check the regulating flag ?

alicecaron: You're right, we will forbid this state in powsybl-core and fix (regulating = true && hasLoadTapChangingCapabilities = false) by setting hasLoadTapChangingCapabilities to true for PST as for CGMES we must not change the regulating value. I kept the check here to be sure both are set to true, but it should be by construction, you think we should remove this double check anyway?

alicecaron: yes done

PR #1240, opened 2025-06-06
Update target Q of generators when target P changes (in case of forceTargetQInReactiveLimits)

SylvestreSakti:

No

#1154 introduced a new parameter to force generators to have a targetQ within reactive limits. When activated, this introduced a corner case to handle, if : 
- A generator has a targetQ outside reactive limits, so its targetQ is updated to fit in the limits
- Then slack distribution is applied and the targetP of this generator changes, then the reactive limits changes
- This change is not taken into account and targetQ is again outside reactive limits
- Then the loadflow is computed
- Then at the end, when updating generators state, targetQ is computed again and updated to fit in the new limits

This produces a wrong result with reactive power inbalances in the network. A caching system and refactoring of generationTargetQ has been done to handle this case.

targetQ is computed only once in the loadflow (at the beginning) and once at the end, after the loadflow when updating states.

A system of caching generationTargetQ is introduced. When generationTargetP of any bus is updated (during active power distribution), then this generationTargetQ value is invalidated. It'll be recomputed again when needed (with an update if it is outside reactive limits)

Here are the use cases where generationTargetQ is modified :

| Use case | Before this PR | After this PR
|--------|-----------|-------------|
|Bus is PQ at LfNetwork loading| Computed to sum of the targetQ of generators | Invalidated and computed|
|Bus is PV at LfNetwork loading| Computed to sum of the targetQ of generators that are not in voltage control | Invalidated and computed |
| Bus is PV after PQ during loadflow | Set to 0 (not very relevant but no impact) | Invalidated and computed |
| Bus is PQ after PV during loadflow (ReactiveLimitsOuterloop) | Set to QLimit | Frozen to QLimit |
| Bus is PQ after PV during loadflow (TransformerVoltageControlOuterloop) | Set to Q.eval() | Frozen to Q.eval() |
| Bus is either PQ or PV after LfContingency | Set to generationTargetQ without disabled generators targetQ | Invalidated and computed |
| Bus is restored in Security or Sensitivity analysis | Set to previous generationTargetQ | Frozen to previous generationTargetQ if previously frozen, otherwise invalidated and computed|

With this new refactoring, there are only two semantics that cover all the mentioned cases : 
- Either the bus generationTargetQ is frozen (it cannot change until invalidation)
- Either the bus generationTargetQ is invalidated (and its value will always be the sum of the targetQ of generators that are not in voltage control and that are not disabled) 

- [ ] Yes
- [x] No


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
```diff
@@ -49,7 +49,7 @@ public abstract class AbstractLfBus extends AbstractElement implements LfBus {
 
     protected Double generationTargetP;
 
-    protected double generationTargetQ = 0;
+    protected Double generationTargetQ = 0.;
```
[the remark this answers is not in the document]
vidaldid-rte: By the way it is accessed directly in setGenerationTargetQ and this may cause an NPE if invalidated.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
```diff
@@ -47,6 +47,8 @@ public abstract class AbstractLfGenerator extends AbstractLfInjection implements
 
     protected TwoSides controlledBranchSide;
 
+    protected double targetQ = Double.NaN;
+
```
vidaldid-rte, line 51: This field is never used (and overriden in LfGeneratorImpl)

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -31,6 +32,8 @@ public final class LfGeneratorImpl extends AbstractLfGenerator {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LfGeneratorImpl.class);
 
+    private static final double TARGET_Q_EPSILON = 1e-2 / PerUnit.SB;
+
```
vidaldid-rte, line 36: Maybe you should use the min between this value and param maxReactivePowerMismatch. Or you may not respect the max mismatch.

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -170,15 +175,23 @@ public OptionalDouble getRemoteControlReactiveKey() {
...
-                targetQ = maxQ;
+            if (computedTargetQ < minQ) {
+                computedTargetQ = minQ;
+            } else if (computedTargetQ > maxQ) {
+                computedTargetQ = maxQ;
+            }
+            if (Math.abs(targetQ - computedTargetQ) > TARGET_Q_EPSILON) { // Log message only if updated value
+                LOGGER.info("TargetQ of generator {} has been updated because outside reactive power limits (from {} MVar to {} MVar)", getId(), targetQ * PerUnit.SB, maxQ * PerUnit.SB);
+                Reports.reportGeneratorWithUpdatedTargetQ(network.getReportNode(), this, targetQ * PerUnit.SB, maxQ * PerUnit.SB);
             }
```
vidaldid-rte, line 194: Is there a reason to duplicate logging code and reporting code ?

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -170,15 +175,23 @@ public OptionalDouble getRemoteControlReactiveKey() {
 
     @Override
     public double getTargetQ() {
-        double targetQ = Networks.zeroIfNan(getGenerator().getTargetQ()) / PerUnit.SB;
+        if (targetQ == null) {
+            targetQ = Networks.zeroIfNan(getGenerator().getTargetQ()) / PerUnit.SB;
```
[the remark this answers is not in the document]
SylvestreSakti: The only goal of this mechanism is to remember the previous value of targetQ before updating it for the logging purpose.
Maybe I could change the implementation (using a variable oldTargetQ) to make the purpose more understandable.

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
vidaldid-rte: The report should be emitted only if the Q defined at IIDM level is modified. So basically the first time you call this function (or assuming you get notification from IIDM model change for Network cache support, it the targetQ has been modified in IIDM and needs to be readjusted).
When done during computation it is business as usual.

src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
```diff
@@ -25,7 +25,7 @@ public class LfStarBus extends AbstractLfBus {
     private final double nominalV;
 
     public LfStarBus(LfNetwork network, ThreeWindingsTransformer t3wt, LfNetworkParameters parameters) {
-        super(network, Networks.getPropertyV(t3wt), Math.toRadians(Networks.getPropertyAngle(t3wt)), false);
+        super(network, Networks.getPropertyV(t3wt), Math.toRadians(Networks.getPropertyAngle(t3wt)), parameters);
         this.t3wtRef = Ref.create(t3wt, parameters.isCacheEnabled());
```
vidaldid-rte, line 29: You did not change the conformLoad flag here ?

src/main/resources/com/powsybl/openloadflow/reports.properties
```diff
@@ -24,6 +24,7 @@ olf.generatorsDiscardedFromVoltageControlBecauseNotStarted = ${impactedGenerator
 olf.generatorsDiscardedFromVoltageControlBecauseReactiveRangeIsTooSmall = ${impactedGeneratorCount} generators have been discarded from voltage control because of a too small reactive range
 olf.generatorsDiscardedFromVoltageControlBecauseTargetPIsOutsideActiveLimits = ${impactedGeneratorCount} generators have been discarded from voltage control because targetP is outside active power limits
 olf.generatorsDiscardedFromVoltageControlBecauseTargetVIsImplausible = ${impactedGeneratorCount} generators have been discarded from voltage control because targetV is implausible
+olf.generatorWithUpdatedTargetQ = TargetQ of generator ${generatorId} has been updated because outside reactive power limits (from ${oldTargetQ} MVar to ${newTargetQ} MVar)
```
vidaldid-rte: THis report is not tested.

src/main/java/com/powsybl/openloadflow/network/impl/LfStarBus.java
SylvestreSakti: From my understanding, it is not used for LfStarBus

src/main/java/com/powsybl/openloadflow/network/impl/LfDanglingLineBus.java
```diff
@@ -25,9 +25,10 @@ public class LfDanglingLineBus extends AbstractLfBus {
     private final double nominalV;
 
     public LfDanglingLineBus(LfNetwork network, DanglingLine danglingLine, LfNetworkParameters parameters, LfNetworkLoadingReport report) {
-        super(network, Networks.getPropertyV(danglingLine), Math.toRadians(Networks.getPropertyAngle(danglingLine)), false);
+        super(network, Networks.getPropertyV(danglingLine), Math.toRadians(Networks.getPropertyAngle(danglingLine)), parameters);
         this.danglingLineRef = Ref.create(danglingLine, parameters.isCacheEnabled());
         nominalV = danglingLine.getTerminal().getVoltageLevel().getNominalV();
+        this.distributedOnConformLoad = false;
```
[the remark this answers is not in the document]
SylvestreSakti: Done

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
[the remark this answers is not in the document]
SylvestreSakti: Finally, I removed the caching of targetQ of LfGenerator (we still keep the caching of generationTargetQ in LfBus). This was only for logging purpose, so I prefered to introduce a boolean isTargetQForcedInReactiveLimits instead which is simpler. (And it avoids potential interference with Network Cache)

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
SylvestreSakti: The report is now only emitted once indeed. 

A remark on the Logger : I realized that even if the Logger object is passed as a parameter to the reporting method in the class Reports, then we still have a logged message displayed as "c.p.o.u.Reports" message. We want it to be displayed as "c.p.o.n.i.LfGeneratorImpl " message. So I kept the logging info outside (we may have to fix that with other log messages)

src/main/resources/com/powsybl/openloadflow/reports.properties
SylvestreSakti: Now tested

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
vidaldid-rte: I made the same observation. This is because the logback pattern in test/resources/logback.xml has  %-20C{1} to describe the log source. C means caller class. In the user decides instead to have something like %logger{36} , then the logger name is displaed.
The problem with the current code is that there is code duplication between reports and logs for something that is essentially the same thing. And code duplication means possible inconsistencies.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfBus.java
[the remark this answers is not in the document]
SylvestreSakti: Some refactoring has been done to highlight these semantics (especially the concept of "freezing" generationTargetQ) : 

| Use case | Before this PR | After this PR
|--------|-----------|-------------|
|Bus is PQ at LfNetwork loading| Computed to sum of the targetQ of generators | Invalidated and computed|
|Bus is PV at LfNetwork loading| Computed to sum of the targetQ of generators that are not in voltage control | Invalidated and computed |
| Bus is PV after PQ during loadflow | Set to 0 (not very relevant but no impact) | Invalidated and computed |
| Bus is PQ after PV during loadflow (ReactiveLimitsOuterloop) | Set to QLimit | Frozen to QLimit |
| Bus is PQ after PV during loadflow (TransformerVoltageControlOuterloop) | Set to Q.eval() | Frozen to Q.eval() |
| Bus is either PQ or PV after LfContingency | Set to generationTargetQ without disabled generators targetQ | Invalidated and computed |


With this new refactoring, there are only two semantics that cover all the mentioned cases : 
- Either the bus generationTargetQ is frozen (it cannot change until invalidation)
- Either the bus generationTargetQ is invalidated (and its value will always be the sum of the targetQ of generators that are not in voltage control and that are not disabled)

vidaldid-rte (approved):

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
```diff
@@ -29,7 +27,7 @@
  */
 public final class LfGeneratorImpl extends AbstractLfGenerator {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(LfGeneratorImpl.class);
+    public static final Logger LOGGER = LoggerFactory.getLogger(LfGeneratorImpl.class);
```
Hadrien-Godard, line 30: I don't understand this change. It does not seem necessary for me to put this Logger public. When I revert it as private, compile and tests are ok.

Hadrien-Godard (approved):

src/main/java/com/powsybl/openloadflow/network/impl/LfGeneratorImpl.java
SylvestreSakti: Thank you, it was done first to send the Logger to the Report class (so that it would write the same message in the log and in the report). (See this comment https://github.com/powsybl/powsybl-open-loadflow/pull/1240#discussion_r2145204002).
Finally, I did it in another manner, so indeed, it doesn't need to be public.

PR #1248, opened 2025-07-01
Make generator started MW threshold configurable

jeandemanged:

<del>- [ ] A PR or issue has been opened in all impacted repositories (if any)</del>

No

Feature

Use of hardcoded 1e-4 MW threshold to identify not started generators and exclude them for slack distribution and voltage control.
Threshold cannot be modified, and these checks cannot be disabled by the user.
Use case for disabling the check: allow generators having a targetP at zero and with Pmin<=TargetP<=Pmax to participate in slack distribution.

The threshold is configurable via new `generatorStartedMwThreshold` parameter.
The check can be disabled by providing a zero or negative value for the parameter - which was not possible before. By doing this generators having a targetP at zero and with Pmin<=TargetP<=Pmax are able to participate in slack distribution.

- [x] No


jeandemanged:
Hi @vidaldid-rte, @SylvestreSakti ,

Closing this one in favor of #1265 which would make things easier from user perspective.

PR #1255, opened 2025-07-16
Handle case of not found synchronous component of regulating terminal of voltage control

SylvestreSakti:
Note: the problem is addessed by #1261 that provides a test to reproduce it.
Thi fix in this PR is however the same as in #1261

No

Some rare situations leads to Null Pointer Exception when running security analysis, due to synchronous component of regulating terminal being `null`.

Null Pointer exception is thrown

A check is added to avoid the exception. A warning is sent in case of missing synchronous component. This should lead to investigate further to get the cause of such an exception.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
@@ -242,6 +242,10 @@ public abstract class AbstractLfGenerator extends AbstractLfInjection implements
             LOGGER.warn("Regulating terminal of LfGenerator {} is out of voltage: voltage control discarded", getId());
             return;
         }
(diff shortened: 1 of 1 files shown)


HugoKulesza (changes requested):
I think it should be `parameters.isBreakers && synchronousComponent==null` (I don't know what behaviour the getBusBreakerView.getBus can have if it is not supposed to be a BusBreakerView)

SylvestreSakti:
> I think it should be `parameters.isBreakers && synchronousComponent==null` (I don't know what behaviour the getBusBreakerView.getBus can have if it is not supposed to be a BusBreakerView)

Indeed ! Thanks

SylvestreSakti:
I close this PR since this bug is addressed by #1261

PR #1261, opened 2025-07-29
NPE in Security Analysis in breaker mode if a group regulates voltage on unconnected remote terminal

vidaldid-rte:

No

BUg fix

Security analysis breaks with a NPE if run in breaker mode and with a generator crontrolling tension on an unconnected bus of an equipment thatis connected on the other side to the current synchronous component. 

[TO BE DISCUSSED]
Proposal: Voltage control of the generator is moved to local voltage control.

- [ ] Yes
- [X] No

### DIFF
--- a/src/test/java/com/powsybl/openloadflow/network/DoubleBusNetworkFactory.java
+++ b/src/test/java/com/powsybl/openloadflow/network/DoubleBusNetworkFactory.java
@@ -1,0 +1,271 @@
+package com.powsybl.openloadflow.network;
+
+import com.powsybl.iidm.network.*;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class DoubleBusNetworkFactory {
+
+    /**
+     * vl1 : 400 kv Double bus Voltage Level
+     * vlgens : 20 kv Two separated bus with a generator each and a tranformer connected to VL1
+     *          One of the generators is connected also to a dangling transformer and
+     *          monnitors voltage on the dangling terminal
+     * vlload : 63 kV. A load connected by a transforer to vl1
+     * @return
+     */
+    public static Network create() {
+        Network network = NetworkFactory.findDefault().createNetwork("double-bus", "source");
+        AtomicInteger node =  new AtomicInteger();
+        Substation substation = network.newSubstation()
+                .setId("station")
+                .add();
+        VoltageLevel vl = substation.newVoltageLevel()
+                .setNominalV(400)
+                .setTopologyKind(TopologyKind.NODE_BREAKER)
+                .setId("vl1")
+                .add();
+
+        BusbarSection bbs1 = vl.getNodeBreakerView().newBusbarSection()
+                .setNode(node.incrementAndGet())
+                .setId("bbs1")
+                .add();
+
+        int bbs1Node = node.get();
+
+        BusbarSection bbs2 = vl.getNodeBreakerView().newBusbarSection()
+                .setNode(node.incrementAndGet())
+                .setId("bbs2")
+                .add();
+
+        int bbs2Node = node.get();
+
+        vl.getNodeBreakerView().newSwitch()
+                .setRetained(true)
+                .setId("discbbs1")
+                .setKind(SwitchKind.DISCONNECTOR)
+                .setNode1(bbs1Node)
+                .setNode2(node.incrementAndGet())
+                .add();
+
+        vl.getNodeBreakerView().newSwitch()
+                .setRetained(true)
+                .setId("bbs1-2")
+                .setKind(SwitchKind.BREAKER)
+                .setNode1(node.get())
+                .setNode2(node.incrementAndGet())
+                .add();
+
+        vl.getNodeBreakerView().newSwitch()
+                .setRetained(true)
+                .setId("discbbs2")
+                .setKind(SwitchKind.DISCONNECTOR)
+                .setNode1(node.get())
+                .setNode2(bbs2Node)
+                .add();
+
+        VoltageLevel vlGens = substation.newVoltageLevel()
+                .setNominalV(20)
+                .setTopologyKind(TopologyKind.NODE_BREAKER)
+                .setId("vlgen")
+                .add();
+
+        createGroup(node, "1", substation, vlGens, vl, bbs1, bbs2, false);
+        // Invert bbs to exchange remoe regulated terminal
+        createGroup(node, "2", substation, vlGens, vl, bbs2, bbs1, true);
+
+        VoltageLevel vlLoads = substation.newVoltageLevel()
+                .setNominalV(63)
+                .setTopologyKind(TopologyKind.NODE_BREAKER)
+                .setId("vlload")
+                .add();
+        createLoad(node, "1", substation, vlLoads, vl, bbs2, bbs1);
+
+
+
+        return network;
+    }
+
+    static private void createGroup(AtomicInteger node, String suffix, Substation substation, VoltageLevel vlGens,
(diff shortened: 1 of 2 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfGenerator.java
```diff
@@ -238,7 +238,7 @@ protected void setVoltageControl(double targetV, Terminal terminal, Terminal reg
             return;
         }
         Bus controlledBus = parameters.isBreakers() ? regulatingTerminal.getBusBreakerView().getBus() : regulatingTerminal.getBusView().getBus();
-        if (controlledBus == null) {
+        if (controlledBus == null || controlledBus.getSynchronousComponent() == null) {
             LOGGER.warn("Regulating terminal of LfGenerator {} is out of voltage: voltage control discarded", getId());
```
SylvestreSakti, line 242: Can there be a better warn message than the already existing one ? (Not sure of what means "out of voltage")

Maybe this instead : "Regulating terminal of LfGenerator {} is not found in the component: voltage control discarded"

vidaldid-rte: > Can there be a better warn message than the already existing one ? (Not sure of what means "out of voltage")
> 
> Maybe this instead : "Regulating terminal of LfGenerator {} is not found in the component: voltage control discarded"
This case is really about the fact that the terminal is disconnected (no bus).
There is a different case when the terminal belongs to another component.

The expression "out of voltage" (which looks strange because even if a terminal say of a transformer, is disconnected there can be a significant, measurable and dangerous voltage )
 is used in many other places of OpenLoadFlow to describe the fact that a terminal is unconnected.
 If we decide to change it, I suggest this is done in a separate PR.

src/test/java/com/powsybl/openloadflow/network/DoubleBusNetworkFactory.java
jeandemanged, line 1: missing copyright and author

src/test/java/com/powsybl/openloadflow/network/DoubleBusNetworkFactory.java
```diff
@@ -0,0 +1,268 @@
...
+                .setNominalV(63)
+                .setTopologyKind(TopologyKind.NODE_BREAKER)
+                .setId("vlload")
+                .add();
+        createLoad(node, "1", substation, vlLoads, vl, bbs2, bbs1);
+
+        return network;
+    }
+
+    private static void createGroup(AtomicInteger node, String suffix, Substation substation, VoltageLevel vlGens,
```
jeandemanged, line 90: rename to createGenerator

src/test/java/com/powsybl/openloadflow/network/DoubleBusNetworkFactory.java
```diff
@@ -0,0 +1,268 @@
...
+                .add();
+
+        VoltageLevel vlGens = substation.newVoltageLevel()
+                .setNominalV(20)
+                .setTopologyKind(TopologyKind.NODE_BREAKER)
+                .setId("vlgen")
+                .add();
+
+        createGroup(node, "1", substation, vlGens, vl, bbs1, bbs2, false);
+        // Invert bbs to exchange remoe regulated terminal
```
jeandemanged, line 77: typo

```suggestion
        // Invert bbs to exchange remote regulated terminal
```

src/test/java/com/powsybl/openloadflow/network/DoubleBusNetworkFactory.java
```diff
@@ -0,0 +1,268 @@
...
+                .setTopologyKind(TopologyKind.NODE_BREAKER)
+                .setId("vlload")
+                .add();
+        createLoad(node, "1", substation, vlLoads, vl, bbs2, bbs1);
+
+        return network;
+    }
+
+    private static void createGroup(AtomicInteger node, String suffix, Substation substation, VoltageLevel vlGens,
+                                    VoltageLevel vl, BusbarSection bbs1, BusbarSection bbs2, boolean withDandlingControlTerminal) {
```
jeandemanged, line 91: typo

```suggestion
                                    VoltageLevel vl, BusbarSection bbs1, BusbarSection bbs2, boolean withDanglingControlTerminal) {
```

src/test/java/com/powsybl/openloadflow/network/DoubleBusNetworkFactory.java
```diff
@@ -0,0 +1,268 @@
...
+
+    private DoubleBusNetworkFactory() {
+    }
+
+    /**
+     * vl1 : 400 kv Double bus Voltage Level
+     * vlgens : 20 kv Two separated bus with a generator each and a tranformer connected to VL1
+     *          One of the generators is connected also to a dangling transformer and
+     *          monnitors voltage on the dangling terminal
+     * vlload : 63 kV. A load connected by a transforer to vl1
```
jeandemanged, line 17: typos

```suggestion
     * vl1 : 400 kv Double bus Voltage Level
     * vlgens : 20 kv Two separated bus with a generator each and a tranformer connected to VL1
     *          One of the generators is connected also to a dangling transformer and
     *          monitors voltage on the dangling terminal
     * vlload : 63 kV. A load connected by a transformer to vl1
```

src/test/java/com/powsybl/openloadflow/network/DoubleBusNetworkFactory.java
```diff
@@ -0,0 +1,268 @@
...
+                .setNode(node.incrementAndGet())
+                .add();
+
+        Generator g = vlGens.newGenerator()
+                .setId("g" + suffix)
+                .setTargetP(50)
+                .setMaxP(100)
+                .setMinP(10)
+                .setVoltageRegulatorOn(true)
+                .setRegulatingTerminal(bbs1.getTerminal()) // Will be changed after tranformer is created
```
jeandemanged, line 105: typo

```suggestion
                .setRegulatingTerminal(bbs1.getTerminal()) // Will be changed after transformer is created
```

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -4598,4 +4598,16 @@ void testTwoGeneratorsAtSameBus() {
         assertReactivePowerEquals(-6.5, network.getGenerator("g2").getTerminal());
 
     }
+
+    @Test
+    void testDoubleBus() {
```
jeandemanged, line 4603: assertions to be added, no assertion causes sonar issue with "blocker" severity.

Also the issue does not lie in SA but in LfNetworkLoaderImpl, test could be limited to this class in LfNetworkLoaderImplTest and removed from OpenSecurityAnalysisTest

```java
    @Test
    void issue() {
        network = DoubleBusNetworkFactory.create();
        LfNetworkParameters networkParameters = new LfNetworkParameters()
                .setBreakers(true);
        List<LfNetwork> lfNetworks = Networks.load(network, networkParameters);
        // TODO assertions
    }
```

jeandemanged (changes requested):
wondering if we also don't have an implementation issue in AbstractLfGenerator, today in there we have a controlledBusId String and public LfBus getControlledBus() {
        return network.getBusById(controlledBusId);
    }

would be better to just have an LfBus controlledBus instead of String. But I guess it is done this way because not all LfBus-es are created when loading an LfNetwork ? or not

looking at RTC there what is used is e.g. 
`LfBus controlledBus = getLfBus(rtc.getRegulationTerminal(), lfNetwork, parameters.isBreakers());`

maybe we should just do the same

PR #1273, opened 2025-08-28
Update default parameter values (breaking change)

vidaldid-rte:

No

Update default parameter values to provide better defaults for simulations

Currently, the value for slackDistributionFailureBehavior is LEAVE_ON_SLACK_BUS. With this parameter OpenLoadFlow returns a success simulation status even if it is impossible to distribute the slack. In the returned solution, production is not equal to the sum of loads and losses. And a fictive injection is added to the slack bus. However, in most cases, the fact that the slack cannot be distributed should be considered as an error.

The current default value of transformerVoltageControlMode is WITH_GENERATOR_VOLTAGE_CONTROL, a very simple outerloop that runs only once and then freezes the transfomer taps even if the voltage at tranformer controlled buses changes after other outerloops. OpenLoadFlow provides two more realistic algorithm: AFTER_GENERATOR_VOLTAGE_CONTROL which is a more realistic version of WITH_GENERATOR_VOLTAGE_CONTROL, and can be configured to minimize the tap difference with the initial position and avoid extreme positions when sveral transformers regulated the same bus, and  INCREMENTAL_VOLTAGE_CONTROL which is based on sensitivity analysis and tolerates impossible voltage targets.
The new default is  INCREMENTAL_VOLTAGE_CONTROL.

The default value of slackDistributionFailureBehavior is set to FAIL.
To allow sample models to run successfully, the default value of plausibleActivePowerLimit has been increased from 5000MW to 10000MX.

The new default for transfomerVoltageContolMode is INCREMENTAL_VOLTAGE_CONTROL.

Some unit tests had to be modifed to run with the default parameters;
Ieee14Test.java
   The reference results have been modified because the slack distribution is performed because plausibleActivePowerLimit is increased.

AcLoadFlowBatteryTest
The minP of battery bat2 has been udated so that targetP >=minP so that it can participate to slack distribution.
 The reference results have been modified.

ACLoadFlowReportTest.busesOutOfRealisticVoltageRangeTest
   The reference results have been modified because the slack distribution is performed because plausibleActivePowerLimit is increased.

WoodburyDcSecurityAnalysisWithActionsTest
   slackDistributionFailureBehavior explicitely set to LEAVE_ON_SLACK_BUS

GenerationActivePowerDistributionStep.securityAnalysisTest
  Group capacity (maxP) has been increased to allow slack distribution in some contingencies.

OpenSecurityAnalysisTest
  testLineOpenOneSideContingencyBusBreaker
   The reference results have been modified because the slack distribution is performed because plausibleActivePowerLimit is increased.

  testWithShuntAndGeneratorVoltageControls2
    The second group has been started so that it can participate to slack distribution in some contingencies.
    
  testNoRemainingLoad
    The exected cpnvergence status of the contingency has been updated to FAIL since there is no load and the balance mode is proportiional to load. 

  testWithFictitiousLoad
    An extenions LoadDetail has been added to allow slack distribution in mode proportional to conform load.

AcSensitivityAnalysisContingenciesTest
   testPredefinedResults3
      Capacity increased for second generator to allow slack distribution

SecondaryVoltageControlTest

        testWithGeneratorSharedRemoteVoltage()
   The reference results have been modified because the slack distribution is performed because plausibleActivePowerLimit is increased.

DcSensitivityAnalysisTest
	testHvdcSensiVsc2
	   keep HVDC in the right direction so that energy can flow from all generators to the load
        testVscInjection
	   Small generator added on the SC of the load to allow slack distribution

AcLoadFlowVscTest
       testVscVoltageControlWithOneSideDisconnected
           Accept that SC1 fails if fictive load added when AC emulation is removed

ReferenceBusPrioritiesTest
       balanceMode set to proportional to Load to allow convergence on second component

DCLoadFlowTest
       multiCcTest add a generator on the isolated part
   

Unit tests about tranformer voltage control that test the WITH_TRANFORMER_VOLTAGE control and its specific resutls (such as computed continuous ration) set explicitely the parameter and continue to run as before.

- [X] Yes
- [ ] No

Users using OpenLoadFLow with default parameters will see different results:
    - If they have groups with pMax > 5000MW. The groups will now be able to participate to slack distribution.
    - If they have networks impossible to balance. OpenLoadFlow will no return an error. 

    - if they use tranformerVoltageControl. In this case, the computed tap positions might be different. OpenLoadFlow will also need more outerloops (since the INCREMENTAL_VOLTAGE_CONTROL can re-run if another outerloop has modified the voltage of buses controlled by transformers).

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
@@ -52,7 +52,7 @@ import static com.powsybl.openloadflow.network.util.ParticipatingElement.normali
  * @author Gaël Macherel {@literal <gael.macherel@artelys.com>}
  */
 public class DcSensitivityAnalysis extends AbstractSensitivityAnalysis<DcVariableType, DcEquationType> {
-    private static Logger LOGGER = LoggerFactory.getLogger(DcSensitivityAnalysis.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(DcSensitivityAnalysis.class);
 
     private static final double FUNCTION_REFERENCE_ZER0_THRESHOLD = 1e-13;
 


SylvestreSakti:
Hi, thank you for these modifications ! I think `DcValueVoltageInitializer` should be modified to keep the `LEAVE_ON_SLACK_BUS` behavior. Some rare cases have been met where the DC slack distribution fails (because losses are ignored) at the initialization, and AC simulation solves that inbalance. (A unit test may be added for that)

vidaldid-rte:
> Instead of breaking all this unit tests, I would have set parameters for this test to old values (only the one that need to be modified)

It is important to check that the default models work with the default parameters, and with a result that makes sense. For instance that helped identify the need to increase the max plausible targetP.
Other tests had issues that needed to be fixed (targetP lower than pMin for a battery charging / component on an HVDC side without any way to balance the line losses...)

But lets's review the tests to make sure the test purpose has been kept, and that apropriate coverage of the LEAVE_ON_SLACK_BUS is kepts.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -69,6 +69,7 @@ private static DcLoadFlowParameters createDcLoadFlowParameters(LfNetworkParamete
                 .setEquationSystemCreationParameters(equationSystemCreationParameters)
                 .setMatrixFactory(matrixFactory)
                 .setDistributedSlack(lfParameters.isDistributedSlack())
+                .setSlackDistributionFailureBehavior(OpenLoadFlowParameters.SlackDistributionFailureBehavior.LEAVE_ON_SLACK_BUS)
```
SylvestreSakti, line 72: As discussed, this is not well defined why SlackDistributionFailureBehavior was not taken into account for DcSensitivityAnalysis. It may be better to take into account the parameter for consistency, and mention that in the release note (if any user gets a problem with FAIL, the parameter can be manually changed)

vidaldid-rte:
> Instead of breaking all this unit tests, I would have set parameters for this test to old values (only the one that need to be modified)
It is important to run the tests using sample models with the defautl parameters to ensure they work well.
And for example this showed the need for increasing the default value of max plausible target P.

Also some tests had issues that are visible with the new default (targetP smaller than minP, components on the other side of an HVDC that have no way to distribute slack and absorb losses, etc...

However it is good to verify that I did not by inadvertance change the purpose of a test. As discussed, lets spend some time together on this review.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
vidaldid-rte: Test finally added (in a slightly modified version) to get required coverage.

src/main/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysis.java
```diff
@@ -64,11 +68,18 @@ private static DcLoadFlowParameters createDcLoadFlowParameters(LfNetworkParamete
...
+        if (parametersExt.getSlackDistributionFailureBehavior() != OpenLoadFlowParameters.SlackDistributionFailureBehavior.LEAVE_ON_SLACK_BUS) {
+            LOGGER.warn("Slack distribution failure mode {} ignored. Using LEAVE_ON_SLACK_BUS for DC sensitivity analysis", parametersExt.getSlackDistributionFailureBehavior());
+        }
+
         return new DcLoadFlowParameters()
                 .setNetworkParameters(networkParameters)
                 .setEquationSystemCreationParameters(equationSystemCreationParameters)
                 .setMatrixFactory(matrixFactory)
                 .setDistributedSlack(lfParameters.isDistributedSlack())
+                // Currently the DC sensitiviy analysis dpes not check slack distribution success or failure and runs always in
```
jeandemanged, line 80: typos

```suggestion
                // Currently the DC sensitivity analysis does not check slack distribution success or failure and runs always in
```

src/test/java/com/powsybl/openloadflow/ac/SecondaryVoltageControlTest.java
```diff
@@ -191,6 +191,10 @@ void testUnblockGeneratorFromLimit() throws IOException {
         g6.setTargetV(11.8);
         g8.setTargetV(19.5);
 
+        // This scenario works if the slack distribution fails and an injection is added at the slack bus
+        parametersExt.setPlausibleActivePowerLimit(5000); // Remove large groups from slack distribution
```
jeandemanged, line 195: "group" is understandable only to french people, use "generator" instead

```suggestion
        parametersExt.setPlausibleActivePowerLimit(5000); // Remove large generators from slack distribution
```

src/test/java/com/powsybl/openloadflow/dc/DcLoadFlowTest.java
```diff
@@ -563,6 +565,9 @@ void testDcApproxIgnoreG() {
     @Test
     void testDcResidualMismatchRemaining() {
         Network network = IeeeCdfNetworkFactory.create9();
+        // The purpose of this test is to verify the report when slack distribution mismatch is left on lack bus.
+        parametersExt.setSlackDistributionFailureBehavior(OpenLoadFlowParameters.SlackDistributionFailureBehavior.LEAVE_ON_SLACK_BUS);
+        parametersExt.setPlausibleActivePowerLimit(5000); // remove large groups from slack distribution to cause slack distribution failure
```
jeandemanged, line 570: groups -> generators

```suggestion
        parametersExt.setPlausibleActivePowerLimit(5000); // remove large generators from slack distribution to cause slack distribution failure
```

src/test/java/com/powsybl/openloadflow/dc/DcLoadFlowTest.java
```diff
@@ -577,6 +582,9 @@ void testDcSlackDistributionFailureBehavior() {
         parameters.setBalanceType(LoadFlowParameters.BalanceType.PROPORTIONAL_TO_GENERATION_P);
         Generator referenceGenerator = network.getGenerator("B1-G");
 
+        // Remove large groups from slack distribution to cause a slack distribution failure
```
jeandemanged, line 585:
```suggestion
        // Remove large generators from slack distribution to cause a slack distribution failure
```

src/test/java/com/powsybl/openloadflow/network/HvdcNetworkFactory.java
```diff
@@ -20,13 +20,28 @@ public class HvdcNetworkFactory extends AbstractLoadFlowNetworkFactory {
      * <pre>
      * g1       ld2               ld3
      * |         |                 |
-     * b1 ------- b2-cs2--------cs3-b3
+     * b1 ------- b2-cs2--------cs3-b3--g3 (small group for slack distribution)
```
jeandemanged, line 23:
```suggestion
     * b1 ------- b2-cs2--------cs3-b3--g3 (small generator for slack distribution)
```

src/test/java/com/powsybl/openloadflow/network/HvdcNetworkFactory.java
```diff
@@ -97,6 +112,19 @@ public static Network createVsc() {
            .setP0(50)
            .setQ0(10)
             .add();
+        if (withGeneratorOnB3) {
+            // A small geerator to comensate the loss in the line and HVDC
```
jeandemanged, line 116: typos

```suggestion
            // A small generator to compensate the losses in the line and HVDC
```

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -3659,6 +3672,15 @@ void testWithFictitiousLoad2() {
 
     private void testWithFictitiousLoad2(LoadFlowParameters.BalanceType balanceType) {
         Network network = DistributedSlackNetworkFactory.createNetworkWithLoads();
+
+        // Add a variable capacity to each load
```
jeandemanged, line 3676: capacity -> part

```suggestion
        // Add a variable part to each load
```

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -3595,13 +3599,22 @@ void testWithFictitiousLoad() {
 
     private void testWithFictitiousLoad(LoadFlowParameters.BalanceType balanceType) {
         Network network = DistributedSlackNetworkFactory.createNetworkWithLoads();
+
+        // Add a variable capacity to each load
```
jeandemanged, line 3603: capacity -> part

```suggestion
        // Add a variable part to each load
```

jeandemanged (changes requested):

PR #1275, opened 2025-09-02
Equation model vectorization

geofjamg:

No

Feature

**What is the current behavior ?**

Each equation is stored separately (in `Equation` objects) with its specific `EquationTerm`. Data (V, phi, Pi model characteristics...) is stored directly in the `EquationTerm`objects and the term is evaluated through `eval()` and `der()` implementations.

*Figure 1 : Simplified view of an equation system implemented with the old behaviour*
Equation(old)

In addition to existing scalar equation model, we added an array equation model that allow a more compact model and improve evaluation of equations and its derivatives. For closed branch terms, for bus active target and bus reactive target equations, the vectorized view is implemented. All equations of the same type are stored in one unique `EquationArray` object that contains one unique `EquationTermArray` for each type of terms. Data (V, phi, Pi model characteristics...) is stored in an `AcNetworkVector` object, and all terms are evaluated once with the `eval()` and `der()` methods of the `EquationTermArray`.
Previous single equations still remain for other types (`Equation` are renamed `SingleEquation` and `EquationTerm` are renamed `SingleEquationTerm`). `EquationArray` can also include specif `SingleEquationTerm` objects for specific equations (e.g. open branch terms or AC emulation)

*Figure 2 : Simplified view of the same equation system implemented with the new behaviour*
Equation(new) 2

- [ ] Yes
- [x] No

**Other information**:


src/test/java/com/powsybl/openloadflow/equations/LUDecompositionTest.java
```diff
@@ -112,13 +112,13 @@ void testIncrementalLURobustification() {
 
         LfBus bus = network.getBus(0);
 
-        EquationSystem<AcVariableType, AcEquationType> equationSystem = new EquationSystem<>();
+        EquationSystem<AcVariableType, AcEquationType> equationSystem = new EquationSystem<>(AcEquationType.class, network);
```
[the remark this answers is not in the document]
geofjamg: It does not break any existing design, equation system already depends on network package (look at imports, there is not diff with main branch). OLF equation system is not a general purpose equation model, but something specific to LF problem. We reference everywhere elements (which are components of LF network) including in equations and terms.
Anyway, yes as we only need the network to get elements cardinal and and status, we can pass a reduced abstraction of LfNetwork like "ElementsVector" to get the needed data.

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -1310,11 +1310,7 @@ void testViolationOnThreeWindingsTransformersLeg() {
 
         SecurityAnalysisResult result = runSecurityAnalysis(network, List.of(), new LoadFlowParameters());
         assertEquals(1, result.getPreContingencyResult().getLimitViolationsResult().getLimitViolations().size());
-        LimitViolation expected = new LimitViolation("3wt", null, LimitViolationType.CURRENT, "permanent",
-                60, 400., 1.0F, 435.0831773201809, TwoSides.TWO);
-        int compare = LimitViolations.comparator().compare(expected,
-                result.getPreContingencyResult().getLimitViolationsResult().getLimitViolations().get(0));
-        assertEquals(0, compare);
+        assertEquals(435.083, result.getPreContingencyResult().getLimitViolationsResult().getLimitViolations().get(0).getValue(), LoadFlowAssert.DELTA_I);
```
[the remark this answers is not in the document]
geofjamg: This is just a merge issue.
Reverted.

src/main/java/com/powsybl/openloadflow/equations/EquationTermArray.java
```diff
@@ -0,0 +1,300 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.function.DoubleSupplier;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public class EquationTermArray<V extends Enum<V> & Quantity, E extends Enum<E> & Quantity> {
+
+    public void compress() {
```
vidaldid-rte, line 32: Can this method be under field declaration and constructor ?

src/main/java/com/powsybl/openloadflow/ac/equations/vector/AcNetworkVector.java
```diff
@@ -0,0 +1,787 @@
...
+
+            // dummy Q
+            branchVector.dummyQ[branchNum] = branchVector.dummyQRow[branchNum] != -1 ? state[branchVector.dummyQRow[branchNum]] : 0;
+            branchVector.negDummyQ[branchNum] = -branchVector.dummyQ[branchNum];
+            branchVector.derDummyQ[branchNum] = 1;
+            branchVector.derNegDummyQ[branchNum] = -1;
+
+            if (!branchVector.disabled[branchNum]) {
+
+                branchVector.r1State[branchNum] = branchVector.r1Row[branchNum] != -1 ? state[branchVector.r1Row[branchNum]]
```
vidaldid-rte, line 355: How do you debug ? How can you set a breakpoint when the values of a specific branch Id are computed .

src/main/java/com/powsybl/openloadflow/ac/AcLoadFlowParameters.java
```diff
@@ -160,6 +162,15 @@ public AcLoadFlowParameters setMinNominalVoltageRealisticVoltageCheck(double min
         return this;
     }
 
+    public boolean isVectorized() {
+        return vectorized;
+    }
+
+    public AcLoadFlowParameters setVectorized(boolean vectorized) {
```
vidaldid-rte, line 169: Doesn't seem to be called.
Why should this not always be on ? Any reason ?
And if it is not always on, how do we ensure proper automatic testing of all tested loadflow features ?

src/main/java/com/powsybl/openloadflow/equations/EquationSystemIndex.java
```diff
@@ -188,13 +219,107 @@ public void onEquationTermChange(EquationTerm<V, E> term, EquationTermEventType
...
+                    notifyEquationTermArrayChange(equationTermArray, termNum, EquationSystemIndexListener.ChangeType.REMOVED);
+                    break;
+
+                default:
+                    throw new IllegalStateException("Event type not supported: " + eventType);
+            }
+        }
+    }
+
+    public List<ScalarEquation<V, E>> getSortedEquationsToSolve() {
```
vidaldid-rte, line 290: That name is now misleading. It only returns the scalar equations. Should be renamed.
But this API seems low level. Users need to know that there are scalar equations and vector equations.
Can't we propose a higher level API for the operations that need to perform actions on all equations ?

src/main/java/com/powsybl/openloadflow/network/impl/LfLoadImpl.java
```diff
@@ -62,13 +64,53 @@ public String getId() {
         return bus.getId() + "_load";
     }
 
+    @Override
```
vidaldid-rte, line 67: So you made the Load an LfElement.
But why AbstractInjection is not an AbstractffElement ?
What is the definition of an LfElement ?

src/main/java/com/powsybl/openloadflow/lf/AbstractEquationSystemUpdater.java
```diff
@@ -61,6 +61,19 @@ protected void updateElementEquations(LfElement element, boolean enable) {
                     equationTerm.setActive(enable);
                 }
             }
+
```
vidaldid-rte, line 64: If we go to the direction of distinguishing scalar equations and equation array, can't we provide better abstractions for such operations. This code that performs a different operation for scalars and arrays seems low level dissemeniated in several places and a source of futur maintenance issues.

src/main/java/com/powsybl/openloadflow/equations/Variable.java
```diff
@@ -36,11 +36,15 @@ public V getType() {
     }
 
     public int getRow() {
+        return row[0];
+    }
+
+    public int[] getRowRef() {
```
vidaldid-rte, line 42: The only caller, EquationDerivativeVector could do the same thing with getRow. Why a reference ?

The caller hardcodes access to the 0 index of the array....

src/main/java/com/powsybl/openloadflow/ac/VoltageMagnitudeInitializer.java
```diff
@@ -212,7 +212,17 @@ public void prepare(LfNetwork network) {
...
+            double[] targets = TargetVector.createArray(network, equationSystem, new TargetVector.Initializer<InitVmVariableType, InitVmEquationType>() {
+                @Override
+                public void initialize(ScalarEquation<InitVmVariableType, InitVmEquationType> equation, LfNetwork network, double[] targets) {
+                    VoltageMagnitudeInitializer.initTarget(equation, network, targets);
+                }
+
+                @Override
+                public void initialize(EquationArray<InitVmVariableType, InitVmEquationType> equationArray, LfNetwork network, double[] targets) {
+                    throw new UnsupportedOperationException("TODO");
+                }
```
vidaldid-rte, line 224: Instead of TODO, it would be better to add a comment that the EquationSystem, creted above only has Scalar Equation Terms.
The real toto would be to convert the above equations (InitVmBusEquationTerm) to an arrayEquationTerms

src/main/java/com/powsybl/openloadflow/equations/ScalarEquationTerm.java
```diff
@@ -0,0 +1,206 @@
...
+public interface ScalarEquationTerm<V extends Enum<V> & Quantity, E extends Enum<E> & Quantity> extends EquationTerm<V, E> {
+
+    class MultiplyByScalarEquationTerm<V extends Enum<V> & Quantity, E extends Enum<E> & Quantity> implements ScalarEquationTerm<V, E> {
+
+        private final ScalarEquationTerm<V, E> term;
+
+        private final DoubleSupplier scalarSupplier;
+
+        MultiplyByScalarEquationTerm(ScalarEquationTerm<V, E> term, double scalar) {
+            this(term, () -> scalar);
```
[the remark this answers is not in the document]
vidaldid-rte: The fact that multiplies does not exist for ArrayEquationTerms forces the cohexistence of two versions of equations (Array and Scalar) for anything involved with multiply (for example ReativeFlowEquation).

src/main/java/com/powsybl/openloadflow/equations/Variable.java
```diff
@@ -20,7 +20,7 @@ public class Variable<V extends Enum<V> & Quantity> implements Comparable<Variab
 
     private final V type;
 
-    private int row = -1;
+    private final int[] row = new int[] {-1};
```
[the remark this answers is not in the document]
geofjamg: It looks like to be still needed to quickly get acces to row even after been updated. Maybe we should use a `MutableInt` instead

src/main/java/com/powsybl/openloadflow/equations/Variable.java
geofjamg: Yes to replace by a `MutableInt`

vidaldid-rte: Ah. Would be clearer if it does not affect performance.

src/main/java/com/powsybl/openloadflow/equations/ScalarEquation.java
```diff
@@ -0,0 +1,300 @@
...
+    @Override
+    public void setActive(boolean active) {
+        checkNotRemoved();
+        if (active != this.active) {
+            this.active = active;
+            equationSystem.notifyEquationChange(this, active ? EquationEventType.EQUATION_ACTIVATED : EquationEventType.EQUATION_DEACTIVATED);
+        }
+    }
+
+    public Equation<V, E> addTerm(EquationTerm<V, E> term) {
```
vidaldid-rte, line 104: The signature of the method appears to be incorrect.
The term cannot be an EquationTerm but must be a ScalarEquationTerm.
Same for EquationTermArrayElements. Must be of the same kind.

src/main/java/com/powsybl/openloadflow/equations/EquationSystemListener.java
```diff
@@ -12,7 +12,11 @@
...
 
-    void onEquationChange(Equation<V, E> equation, EquationEventType eventType);
+    void onEquationChange(ScalarEquation<V, E> equation, EquationEventType eventType);
 
-    void onEquationTermChange(EquationTerm<V, E> term, EquationTermEventType eventType);
+    void onEquationTermChange(ScalarEquationTerm<V, E> term, EquationTermEventType eventType);
+
+    void onEquationArrayChange(EquationArray<V, E> equationArray, int elementNum, EquationEventType eventType);
+
+    void onEquationTermArrayChange(EquationTermArray<V, E> equationTermArray, int termNum, EquationTermEventType eventType);
```
vidaldid-rte, line 21: The two new notifications are essentially the same as the existing one. Ther are about a change for an individual network element. Are the new methods really needed ? Can't we use the existing ones with an ArrayEquationTermElement and and EquationArray element ?

src/test/java/com/powsybl/openloadflow/equations/LUDecompositionTest.java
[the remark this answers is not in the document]
geofjamg: LfElementContainer Added

src/main/java/com/powsybl/openloadflow/equations/SingleEquation.java
```diff
@@ -0,0 +1,298 @@
...
+            c = type.ordinal() - o.type.ordinal();
+        }
+        return c;
+    }
+
+    public void write(Writer writer, boolean writeInactiveTerms) throws IOException {
+        writer.append(type.getSymbol())
+                .append(Integer.toString(elementNum))
+                .append(" = ");
+        List<SingleEquationTerm<V, E>> activeTerms = writeInactiveTerms ? terms : terms.stream().filter(SingleEquationTerm::isActive).collect(Collectors.toList());
```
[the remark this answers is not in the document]
geofjamg: Renamed

src/main/java/com/powsybl/openloadflow/lf/AbstractEquationSystemUpdater.java
vidaldid-rte: Closing. The PR surface has been reduced. Let's make eventual abstraction that encapsulate the usage of single vs: array equations out of scope of this PR.

src/main/java/com/powsybl/openloadflow/equations/EquationVector.java
```diff
@@ -37,29 +36,28 @@ public void onStateUpdate() {
...
             throw new IllegalArgumentException("Bad equation vector length: " + array.length);
         }
 
-        evalLhs(array, equations);
+        Arrays.fill(array, 0); // necessary?
+        for (SingleEquation<V, E> equation : equations) {
+            array[equation.getColumn()] = equation.evalLhs();
+        }
+        for (EquationArray<V, E> equationArray : equationSystem.getEquationArrays()) {
+            equationArray.eval(array);
```
vidaldid-rte: maybe a simple fix to split in two iterations as discussed
One on the array terms
One on the single terms of equations array that add (their sum) to the array cell (in a way that minimizes the number of small objetc to load - ideally only the active single terms)
So that the flow on array terms remains linear and fast.

src/main/java/com/powsybl/openloadflow/equations/EquationTermArray.java
```diff
@@ -0,0 +1,326 @@
+/**
+ * Copyright (c) 2023, RTE (http://www.rte-france.com)
```
vidaldid-rte: Is 2023 really the latest modification date for this class ? shouldn't if be 2023-2025

src/main/java/com/powsybl/openloadflow/equations/EquationSystemIndex.java
```diff
@@ -97,20 +124,24 @@ private void addTerm(EquationTerm<V, E> term) {
...
     }
 
-    private void removeTerm(EquationTerm<V, E> term) {
+    private void removeTerm(ScalarEquationTerm<V, E> term) {
         notifyEquationTermChange(term);
-        for (Variable<V> variable : term.getVariables()) {
+        removeVariables(term.getVariables());
+    }
+
+    private void removeVariables(List<Variable<V>> variables) {
```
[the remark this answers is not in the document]
vidaldid-rte: Again it is not a remove operation. It's more like a smartPointer dereference that removes only when there is no more a reference.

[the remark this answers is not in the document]
vidaldid-rte: OK. I thought it was new code. I was really surprised the firs time I read the code that removed variables...

src/main/java/com/powsybl/openloadflow/ac/equations/ClosedBranchAcFlowEquationTerm.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package com.powsybl.openloadflow.ac.equations;
+
+import com.powsybl.openloadflow.equations.EquationTerm;
+import com.powsybl.openloadflow.equations.Variable;
+
+/**
+ * @author Sylvestre Prabakaran {@literal <sylvestre.prabakaran at rte-france.com>}
+ */
+public interface ClosedBranchAcFlowEquationTerm extends EquationTerm<AcVariableType, AcEquationType> {
```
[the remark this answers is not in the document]
SylvestreSakti: Is this compulsory ? We could imagine an EquationTermArrayElement (that is not a SingleEquationTerm) that implements this interface

src/main/java/com/powsybl/openloadflow/ac/equations/fastdecoupled/AbstractClosedBranchAcFlowFastDecoupledEquationTerm.java
```diff
@@ -0,0 +1,63 @@
...
+    protected final double b2;
+    protected final double r1;
+    protected final double a1;
+    protected final Variable<AcVariableType> phi1Var;
+    protected final Variable<AcVariableType> phi2Var;
+    protected final Variable<AcVariableType> v1Var;
+    protected final Variable<AcVariableType> v2Var;
+    protected final Variable<AcVariableType> a1Var;
+    protected final Variable<AcVariableType> r1Var;
+
```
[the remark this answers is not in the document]
SylvestreSakti: This is the case since this term is built only during one step of derivative calculation. Comment added to clarify that.

src/main/java/com/powsybl/openloadflow/equations/EquationTerm.java
```diff
@@ -1,9 +1,8 @@
 /**
- * Copyright (c) 2019, RTE (http://www.rte-france.com)
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
- * SPDX-License-Identifier: MPL-2.0
```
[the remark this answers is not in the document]
SylvestreSakti: SPDX added

src/main/java/com/powsybl/openloadflow/equations/EquationVector.java
```diff
@@ -37,29 +36,28 @@ public void onStateUpdate() {
...
-            array[equation.getColumn()] = equation.evalLhs();
-        }
-    }
-
     @Override
     protected void updateArray(double[] array) {
         Stopwatch stopwatch = Stopwatch.createStarted();
 
-        var equations = equationSystem.getIndex().getSortedEquationsToSolve();
+        var equations = equationSystem.getIndex().getSortedSingleEquationsToSolve();
```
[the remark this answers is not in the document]
vidaldid-rte: Closing. To reconsider after the release.

src/main/java/com/powsybl/openloadflow/ac/equations/fastdecoupled/AbstractClosedBranchAcFlowFastDecoupledEquationTerm.java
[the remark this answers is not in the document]
vidaldid-rte: Actually, the variables are fine. The problem is with the double for a1 and r1 :)
(actually all values move when a tap changes but the classic branch equation term only update a1 and r1 from the pi model)

src/main/java/com/powsybl/openloadflow/ac/equations/ClosedBranchAcFlowEquationTerm.java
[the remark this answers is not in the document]
SylvestreSakti: Finally removed these interfaces :+1:

vidaldid-rte (dismissed):
This is not the approach I would have recommended. But it is in line with what was decided and delivers some performance. So no objection to merge if this is what the majority wants.

SylvestreSakti (dismissed):

SylvestreSakti (approved):

vidaldid-rte (approved):
Ah... ce robot.

PR #1283, opened 2025-09-29
Split AIC remaining slack mismatch into several distributions, one per area

vmouradian:

Bug fix

**Context**
During AIC outerloop active power is distributed over each area until the mismatch (interchange mismatch + slack of the area) is below 'areaInterchangePMaxMismatch'.
When all areas match this condition, it can happen that the slack remaining on the slack bus is above 'slackBusPMaxMismatch'.

This slack mismatch is then distributed over all areas -> the way that this distribution is done is the bugfix here.

The mismatch is distributed in one activePowerDistribution run on all buses.

This causes a bug. Due to the reset at the begining of the distribution.
Generators are reset to their initial state at the begining, then the mismatch is distributed equally over all generators.
This means all the interchange mismatch distribution is undone and just the slack lislatch is distributed. 
This adds a lot of iterations utill the AIC is complete and even leads to reach the iteration limit making the LF fail.

The distributed individually on each area with a dedicated ActivePowerDistribution run, so the previous distributions are kept.

To attribute the participation of each area this logic has been added :  

This distribution will affect each area's interchange and will not necessarily make it closer to its target.
The distribution factor of each area will be computed in a way that minimises chances of having areas increase their interchange mismatch up to more than areaInterchangePMaxMismatch in absolute value.
So the factor is proportional to the "margin" that the area can distribute while keeping  
 -areaInterchangePMaxMismatch < area_total_mismatch < areaInterchangePMaxMismatch.  

It is computed like this:  
factor = sign(slack_bus_mismatch) * area_total_mismatch + areaInterchangePMaxMismatch
Then factors are normalized to have sum of factors equal to 1.

The distribution is done starting by the areas with the higher factor to the lowest.  
In case the power to distribute for an area is under P_RESIDUE_EPS, then P_RESIDUE_EPS is distributed.  
The distribution stops when the remaining mismatch to distribute is under P_RESIDUE_EPS.

- [ ] Yes
- [x] No


src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
```diff
@@ -114,35 +116,57 @@ public OuterLoopResult check(O context, ReportNode reportNode) {
...
         for (Map.Entry<String, Pair<Set<LfBus>, Double>> e : areas.entrySet()) {
             double areaActivePowerMismatch = e.getValue().getRight();
             ActivePowerDistribution.Result result = activePowerDistribution.run(null, e.getValue().getLeft(), areaActivePowerMismatch);
-            resultByArea.put(e.getKey(), result);
+            resultByArea.put(e.getKey(), Pair.of(areaActivePowerMismatch, result));
+        }
+        return resultByArea;
+    }
+
+    protected Map<String, Pair<Double, ActivePowerDistribution.Result>> distributeRemainingSlackMismatch(double mismatch, LfNetwork network) {
```
m-guibert: Add doc to explain quickly how mismatch is distributed

src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
```diff
@@ -210,9 +233,9 @@ protected OuterLoopResult buildOuterLoopResult(Map<String, Pair<Set<LfBus>, Doub
             }
         } else {
             if (movedBuses) {
-                areas.entrySet().stream().sorted(Map.Entry.comparingByKey()).forEach(entry -> {
-                    logger.info("Area {} interchange mismatch ({} MW) distributed in {} distribution iteration(s)", entry.getKey(), entry.getValue().getValue() * PerUnit.SB, iterationsByArea.get(entry.getKey()));
-                    Reports.reportAreaInterchangeControlAreaDistributionSuccess(iterationReportNode, entry.getKey(), entry.getValue().getValue() * PerUnit.SB, iterationsByArea.get(entry.getKey()));
+                resultByArea.entrySet().stream().sorted(Map.Entry.comparingByKey()).forEach(entry -> {
+                    logger.info("Area {} interchange mismatch ({} MW) distributed in {} distribution iteration(s)", entry.getKey(), entry.getValue().getLeft() * PerUnit.SB, entry.getValue().getRight().iteration());
+                    Reports.reportAreaInterchangeControlAreaDistributionSuccess(iterationReportNode, entry.getKey(), entry.getValue().getLeft() * PerUnit.SB, entry.getValue().getRight().iteration());
```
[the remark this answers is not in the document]
vmouradian: clarified code in [7789445](https://github.com/powsybl/powsybl-open-loadflow/pull/1283/commits/7789445e45f11f3b78cb2bb43337385f452db269) with a record, does this seem better ?

src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
```diff
@@ -204,23 +253,48 @@ protected OuterLoopResult buildOuterLoopResult(Map<String, Pair<Set<LfBus>, Doub
...
+        remainingMismatches.stream()
+                .sorted(Comparator.comparing(areaResult -> areaResult.areaId))
+                .forEach(areaResult -> {
+                    if (ActivePowerDistributionType.AREA_INTERCHANGE.equals(areaResult.type)) {
+                        logger.error("Remaining interchange mismatch for Area {}: {} MW", areaResult.areaId, areaResult.distributionResult.remainingMismatch() * PerUnit.SB);
+                        Reports.reportAicAreaInterchangeDistributionMismatch(failureReportNode, areaResult.areaId, areaResult.distributionResult.remainingMismatch() * PerUnit.SB);
+                    } else {
+                        logger.error("Remaining slack distribution mismatch for Area {}: {} MW", areaResult.areaId, areaResult.distributionResult.remainingMismatch() * PerUnit.SB);
+                        Reports.reportAicAreaSlackDistributionMismatch(failureReportNode, areaResult.areaId, areaResult.distributionResult.remainingMismatch() * PerUnit.SB);
+                    }
```
m-guibert: Couldn't  you simplify to something like that?
```
String mismatchType = ActivePowerDistributionType.AREA_INTERCHANGE.equals(areaResult.type) ? "interchange" : "slack distribution";
logger.error("Remaining {} mismatch for Area {}: {} MW", mismatchType , areaResult.areaId, areaResult.distributionResult.remainingMismatch() * PerUnit.SB);
Reports.reportAicAreaInterchangeDistributionMismatch(failureReportNode, mismatchType, areaResult.areaId, areaResult.distributionResult.remainingMismatch() * PerUnit.SB);
```
Plus, same remark for reportAndLogAreaActivePowerDistributionSuccess

vmouradian: I find it nor really doable because the reports are in english and french. It would need a translation in the method, potentially with different fields, and I did nor really find a good way to do this

src/main/java/com/powsybl/openloadflow/lf/outerloop/AbstractAreaInterchangeControlOuterLoop.java
```diff
@@ -167,35 +219,34 @@ protected double getSlackInjection(String areaId, double slackBusActivePowerMism
...
+        List<AreaActivePowerDistributionResult> remainingMismatches = new ArrayList<>();
         double totalDistributedActivePower = 0.0;
         boolean movedBuses = false;
-        for (Map.Entry<String, ActivePowerDistribution.Result> e : resultByArea.entrySet()) {
-            String area = e.getKey();
-            ActivePowerDistribution.Result result = e.getValue();
-            if (!lessThanInterchangeMaxMismatch(result.remainingMismatch())) {
-                remainingMismatchByArea.put(area, result.remainingMismatch());
+        for (AreaActivePowerDistributionResult areaResult : areaResults) {
+            ActivePowerDistribution.Result distributionResult = areaResult.distributionResult;
```
m-guibert: Move this line below the if, since it is only needed after it

PR #1298, opened 2025-11-14
Security Analysis state monitored results for zero impedance branches in zero impedance mode

Hadrien-Godard:

Bug fix

Currently in zero impedance mode, zero impedance monitored branches have NaN flow results.

State monitored zero impedance branches in zero impedance mode SA have flow results. Note that limit violation detection is not affected, that is violations on non impedant branches are not detected.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameterJsonSerializer.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameterJsonSerializer.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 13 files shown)


src/main/java/com/powsybl/openloadflow/network/util/ZeroImpedanceFlows.java
```diff
@@ -48,15 +50,36 @@ public void compute() {
                 .forEach(branch -> branch.updateFlows(0.0, 0.0, 0.0, 0.0));
     }
 
+    public void computeAndProvideResults(Map<String, LfBranch.LfBranchResults> resultsMap) {
```
vidaldid-rte: This is very similar to computeWIthFlowsUpdate. Is it possible to share the code in a method that would take an  argument (data or interface) for the different operations (buildFLowResults vs  updateFlows)

Hadrien-Godard: Done

PR #1302, opened 2025-11-25
Implement loadflow for AC DC networks

2nisb:

No

Feature

Open loadflow does not support DC detailed components from iidm

Open loadflow is able to modelize DC components (DcBus, DcLine, VoltageSourceConverter) and run loadflows on AC DC networks, with one unique jacobian matrix for the whole connected network.

- [x] Yes
- [ ] No

If you defined your own powsybl openloadflow implementation, you should:
Write your own implementations of `LfAcDcConverter`, `LfVoltageSourceConverter`, `LfDcBus` and `LfDcLine`

Implement the following methods in your :
  - `VoltageInitializer` implementations:
    `double getMagnitude(LfDcBus dcBus)`
- `LfBus` implementations:
   `List<LfVoltageSourceConverter> getConverters()`
   `void addConverter(LfVoltageSourceConverter converter)`
  `Optional<VoltageSourceConverterVoltageControl> getVoltageSourceConverterVoltageControl()`
  `void setVoltageSourceConverterVoltageControl(VoltageSourceConverterVoltageControl voltageSourceConverterVoltageControl)`
  `boolean isVoltageSourceConverterVoltageControlled()`

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/LfAcDcNetwork.java
+++ b/src/main/java/com/powsybl/openloadflow/network/LfAcDcNetwork.java
@@ -45,6 +45,8 @@ public class LfAcDcNetwork extends LfNetwork {
 
     @Override
     protected void invalidateSlackAndReference() {
+        slackBuses.clear();
+        acDcReferenceBuses.clear();
         acNetworks.forEach(LfNetwork::invalidateSlackAndReference);
         dcNetworks.forEach(LfNetwork::invalidateSlackAndReference);
     }
@@ -52,18 +54,16 @@ public class LfAcDcNetwork extends LfNetwork {
     @Override
     public void updateSlackBusesAndReferenceBus() {
         if (!acNetworks.isEmpty()) {
-            for (LfNetwork acSubNetwork : acNetworks) {
-                acSubNetwork.updateSlackBusesAndReferenceBus();
-                for (LfBus bus : acSubNetwork.slackBuses) {
+            for (LfNetwork acNetwork : acNetworks) {
+                acNetwork.updateSlackBusesAndReferenceBus();
+                for (LfBus bus : acNetwork.slackBuses) {
                     LfBus slackBus = this.getBusById(bus.getId());
                     if (!this.slackBuses.contains(slackBus)) {
-                        slackBus.setSlack(true);
                         this.slackBuses.add(slackBus);
                     }
                 }
-                LfBus referenceBus = this.getBusById(acSubNetwork.referenceBus.getId());
+                LfBus referenceBus = this.getBusById(acNetwork.referenceBus.getId());
                 if (!acDcReferenceBuses.contains(referenceBus)) {
-                    referenceBus.setReference(true);
                     this.acDcReferenceBuses.add(referenceBus);
                 }
             }


bperr:
Hi there,
I have continued @2nisb work on this pull request and I think it is ready for review. I will gladly appreciate your feedbacks 😊 
For now I added the AC/DC loadflow documentation at the end of the file, please tell me if you think it should be moved to a specific position in the documentation !

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -1305,6 +1404,16 @@ public List<LfNetwork> load(Network network, LfTopoConfig topoConfig, LfNetworkP
             }
         }
 
+        Map<Pair<Integer, Integer>, List<DcBus>> dcBusesByCc = new TreeMap<>();
+        Iterable<DcBus> dcBuses = Networks.getDcBuses(network);
+        for (DcBus dcBus : dcBuses) {
+            Component cc = dcBus.getConnectedComponent();
+            Component sc = dcBus.getDcComponent();
```
SylvestreSakti: This is not a synchronous component (should not be named sc). There is no notion of synchronous component for DC components. I would suggest `dcBusesByCc` should be a Map<Integer, List<DcBus>> that is only indexed by its connected component number.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
SylvestreSakti: Since AcEquationSystemCreator is already a big class, I would suggest creating a class AcEquationSystemForAcDcNetworkCreator that inherits from it , that implements all the additional equations specific to DC network.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -772,6 +790,42 @@ protected void createImpedantBranch(LfBranch branch, LfBus bus1, LfBus bus2,
...
+        Evaluable p1 = null;
+        Evaluable p2 = null;
+        Evaluable i1 = null;
+        Evaluable i2 = null;
+
+        // We keep the same construction as impedant branch equations, we will implement equation updating later
+        SingleEquationTerm<AcVariableType, AcEquationType> closedP1 = null;
+        EquationTerm<AcVariableType, AcEquationType> closedI1 = null;
+        SingleEquationTerm<AcVariableType, AcEquationType> closedP2 = null;
+        EquationTerm<AcVariableType, AcEquationType> closedI2 = null;
```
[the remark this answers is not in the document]
bperr: No other reason than the SingleEquationTerm class did not exist when we firstly wrote this.
I updated it !

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
bperr: But actually I think we need to separate DC buses by their DC component number, because they belong to different DC networks and might have different DC voltages.

So should I not rather just rename `sc` to `dcc` (DC component) ?

SylvestreSakti: If my understanding is good, the dc component number would be useful in case the solver does a specific load flow for each DC network. Since your approach solves the whole connected component with all DC networks at once, there seems to be no need to dissociate each DC component (or each synchronous component for AC part). That's why I suggested not to store it.

If I'm wrong, for which part of the code this dcc variable would be used after ?

src/main/java/com/powsybl/openloadflow/network/util/VoltageInitializer.java
```diff
@@ -19,5 +22,11 @@ public interface VoltageInitializer {
 
     double getMagnitude(LfBus bus);
 
+    double getReactivePower(LfVoltageSourceConverter converter);
+
+    double getActivePower(LfAcDcConverter converter);
+
+    double getMagnitude(LfDcNode dcNode);
```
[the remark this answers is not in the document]
bperr: I moved these three methods to a new interface and created two implementation as suggested.
The `VoltageInitializer` interface has an additional method `getAcDcNetworkIntializer()`

In `UniformValueVoltageInitializer` and `PreviousValueVoltageInitializer`, the corresponding new class is used is returned by this method. In others an Exception is thrown.

bperr:
Hello @SylvestreSakti 
Thanks for your feedback ! 
I pushed the necessary modifications to ensure there is only one AC component.
As you suggested, I also moved the Equation terms relative to DC network in their own package. However, for the moment I haven't moved the DC part of `AcEquationSystemCreator` to another class, what is your final opinion on this topic ?

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -1327,28 +1436,75 @@ public List<LfNetwork> load(Network network, LfTopoConfig topoConfig, LfNetworkP
...
-                    return create(numCc, numSc, network, lfBuses, switchesByCc.get(networkKey), topoConfig,
+                    return createAc(numCc, numSc, network, lfBuses, switchesByCc.get(networkKey), topoConfig,
                             parameters, Reports.createRootLfNetworkReportNode(reportNode, numCc, numSc));
                 })
-                .collect(Collectors.toList());
+                .toList();
+
+        Stream<Map.Entry<Pair<Integer, Integer>, List<DcBus>>> filteredDcBusesByComponentStream = switch (parameters.getComponentMode()) {
+            case MAIN_CONNECTED -> dcBusesByCc.entrySet().stream().filter(e -> e.getKey().getLeft() == ComponentConstants.MAIN_NUM);
+            case MAIN_SYNCHRONOUS -> dcBusesByCc.entrySet().stream().filter(e -> e.getKey().getRight() == ComponentConstants.MAIN_NUM);
```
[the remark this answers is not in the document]
SylvestreSakti: I think I was wrong, since there can be DC lines inside main synchronous component.

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -1059,7 +1224,10 @@ private List<EquationTerm<AcVariableType, AcEquationType>> createActiveInjection
 
     private void createMultipleSlackBusesEquations(EquationSystem<AcVariableType, AcEquationType> equationSystem) {
         List<LfBus> slackBuses = network.getSlackBuses();
-        if (slackBuses.size() > 1) {
+        if (network instanceof LfAcDcNetwork acDcNetwork && slackBuses.size() > acDcNetwork.getAcSubNetworks().size()) {
```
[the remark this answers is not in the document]
bperr: Indeed, with one AC component, multiple slack is supported !

bperr:
Hi,
I just added a single commit that changes the current sign convention in DC network. The currrents that are added in the DC node current balance equation are now the currents that flowing **out** of the node (to be coherent with the active and reactive power balance of AC buses).
Let me know if you are okay with this convention

docs/loadflow/loadflow.md
```diff
@@ -21,6 +21,12 @@ equal to zero and $1$. In case of a branch with voltage or phase control, the $\
 
 Open Load Flow also supports networks with HVDC lines (High Voltage Direct Current lines). An HVDC line is connected to the rest of the AC network through HVDC converter stations, that can be either LCC (Line-Commutated Converter) or VSC (Voltage-Source Converter).
 
+### DC detailed model
+
+Additionally, Open Load Flow supports AC-DC loadflow formulation with detailed model of DC components.
+However, it is currently restricted to embedded DC networks, meaning that the global network should only contain one AC
+network.
```
jeandemanged: terms/definitions to be clarified:
- "network" vs. "island/component"
- "embedded": my understanding is that it means all AC Terminals of DC converters of a given DC island/component must be in the same AC island/component, correct ?
- "should only contain one AC network": would 2 AC islands, each with their own "embedded" DC island work ?, e.g.:
  - ConnectedComponent 0 with synchronous component 0 and DC component 0
  - ConnectedComponent 1 with synchronous component 1 and DC component 1

docs/loadflow/loadflow.md
```diff
@@ -329,4 +335,139 @@ Users should notice that default parameters are optimized for the Newton-Raphson
 When the Fast-Decoupled algorithm is used, we recommend these values for some convergence parameters:
 - [`maxNewtonRaphsonIterations`](parameters.md): 75,
 - [`lineSearchStateVectorScalingMaxIteration`](parameters.md): 4,
-- [`lineSearchStateVectorScalingStepFold`](parameters.md): `3/2 = 1.5`.
\ No newline at end of file
+- [`lineSearchStateVectorScalingStepFold`](parameters.md): `3/2 = 1.5`.
+
+## AC DC flows computing
+
+AC DC lows computing in OpenLoadFLow is similar to AC flows computing, but with AC and DC equations in the same system.
```
jeandemanged: typo

```suggestion
AC DC flows computing in OpenLoadFLow is similar to AC flows computing, but with AC and DC equations in the same system.
```

docs/loadflow/loadflow.md
```diff
@@ -329,4 +335,139 @@ Users should notice that default parameters are optimized for the Newton-Raphson
...
+
+### DC Line
+
+Each DC line adds one term in both of its two connected DC nodes current balance:
+
+$\sum_{i} I_i + \frac{V_1 - V_2}{R}= 0$ for dcNode1
+
+$\sum_{i} I_i - \frac{V_1 - V_2}{R}= 0$ for dcNode2
+
+### Voltage source converters
```
jeandemanged: See https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/grid_model/network_subnetwork.html#ac-dc-converter


Need to document that LCCs in detailed model are currently not supported by OLF

docs/loadflow/loadflow.md
bperr: I reformulated the documentation to be clearer. Let me know if this seems correct now
To my understanding
- A network object represents set of connected component
- A connected component is a collection of AC and DC island/component that are connected through AC/DC converters.
- An island/component is a set of buses of the same type (meaning either all AC or all DC) connected through equipments like lines, transformers,etc...

So to answer your last points:
- yes a DC island is embedded when it is connected to only one AC island. It is an "internal" connection, meaning it does not connect two different AC island.
- I reformulated this point in the documentation, in a single **Connected component**, there should be only one AC island (and no restriction on the number of DC island). Therefore, as you suggested a network with two connected component with each one AC island and some embeded DC islands can be processed by this load flow

docs/loadflow/loadflow.md
```diff
@@ -329,4 +335,139 @@ Users should notice that default parameters are optimized for the Newton-Raphson
...
+
+$V_{1} - V_{2} = V_{Ref}$
+
+Similarly, if the converter controls reactive power, we add an equation to impose $Q_{AC}$ :
+
+$Q_{AC}$ = $Q_{Ref}$
+
+Else the converter controls the AC voltage, and we add an equation to impose $V_{AC}$:
+
+$V_{AC}= V_{Ref}$
```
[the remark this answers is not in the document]
bperr: In the case of VSC with only one AC terminal, is not the PCC the same thing as the converter terminal ?
> For VSC converters having a single AC Terminal, it is valid for the PCC Terminal to be the converter terminal itself. 

Currently, the AC LfBus controlled by the converter is selected through the converter Terminal1, but it can easilly be changed to the converter PccTerminal

docs/loadflow/parameters.md
```diff
@@ -622,6 +623,27 @@ to fix the network data.
...
+
+If `true`, the network supports DC detailed components, and the loadflow is computed on the whole connected network, 
+AC and DC sides in the same Jacobian matrix. Currently, the network shall contain only one AC island, but the number of 
+embedded DC islands is not restricted.
+
+If `false`, the loadflow is the classic one, without DC detailed components.
+
+The default value is `false`.
+
+**Note:** With AC-DC network load flow, the following parameters are restricted:
```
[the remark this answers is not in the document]
bperr: Indeed, an exception is thrown for all these cases. I clarified this point in the documentation.

src/main/java/com/powsybl/openloadflow/network/LfAcDcNetwork.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public List<LfBus> getSlackBuses() {
+        updateSlackBusesAndReferenceBus();
+        return slackBuses;
+    }
+
+    @Override
+    public LfBus getReferenceBus() {
+        updateSlackBusesAndReferenceBus();
+        //FIXME: which bus do we return ?
```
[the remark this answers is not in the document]
bperr: Yes. It is a topic that I have in mind to discuss in a future PR

PR #1333, opened 2026-01-20
Add Q-threshold for voltage controller buses in reactive limits outerloop

p-arvy:

No. 

In this [PR](https://github.com/powsybl/powsybl-open-loadflow-knitro-solver/pull/21/files), we encountered an issue with the reactive limits check. The solver introduced in this PR directly integrates reactive limits into the formulation of an optimization problem that solves the OLF load-flow system. However, at the end of the solve, the reactive limit violation check detects limit violations, even though these are actually due to the convergence threshold around the limits.

A Q-threshold is added to the reactive limit violation check in the reactive limits outer loop.

In the reactive limits outer loop, the reactive power of voltage-controlling buses is checked against their reactive limits. This check was previously performed without any threshold..

A threshold is now introduced in the reactive limit violation check

- [ ] Yes
- [x] No

If you have automated tests that check the number of executions of the reactive limits outer loop, this number may change.

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameterJsonSerializer.java
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowParameterJsonSerializer.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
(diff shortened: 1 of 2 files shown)


SylvestreSakti:
> > Also do you think we could imagine a unit test to cover this small use case ? (A limit that is violated or not depending of this mismatch ? Maybe apparent power violation ?)
> 
> Specifically testing this small use case seems difficult, as it depends on the reactive convergence threshold and, more specifically, on the path followed by the solver when solving the system of equations. Do you have a specific idea, @SylvestreSakti ? @Hadrien-Godard?

Yes indeed it seems difficult to cover this without the case in open-loadflow-knitro-solver. Ok to keep it as it is

PR #1345, opened 2026-02-05
Support having multiple active operational limits group

### LINKED ISSUE #1356: Support having multiple operational limits groups active at the same time

Within a security analysis only one operational limits groups among the active ones is used to check for limit violations.

Within a security analysis all the active operational limits groups are used to check for limit violations.

### Describe the motivation

Since issue [core#3703](https://github.com/powsybl/powsybl-core/issues/3703) it is possible to have several active operational limits groups at the same time to check for limits violations, we want to be able to check for all the active groups not only one as before when only one "selected" operational limits group was possible.

### Extra Information

- is blocked by issue [core#3703](https://github.com/powsybl/powsybl-core/issues/3703) that model the LimitReduction list of groups on which the reduction should be applied.
- loop on the active operational limits groups when checking for limit violations in pre contingency, post contingencies and post contingencies after actions (in `LimitViolationManager:detectViolations` ?). Limit reductions are given in the 3 usages of the method

SylvestreSakti:

Closes #1356 

powsybl-core PR https://github.com/powsybl/powsybl-core/pull/3735 will introduce multiple operational limits groups. This current PR aims at supporting this new functionnality.

Only one operational limits group can be selected

Multiple operational limits group can be selected. A new class LfLimitsGroup (which only contains a list of LfLimit objects) has been introduced to avoid using nested lists.
The operationalLimitsGroupId is added to the LimitViolations that are in the result

- [x] Yes
- [ ] No

For custom open-loadflow implementation using LfNetwork, the **LfBranch API** has been modified : 
`getLimits1()` and `getLimits2()` return now a `List<LfLimitsGroup>` instead of a `List<LfLimit>`

**Other information**:

### DIFF
--- a/docs/advanced_programming/contingency_active_power_loss.md
+++ b/docs/advanced_programming/contingency_active_power_loss.md
@@ -14,7 +14,7 @@ To add another plugin, you will need to code (in Java) an implementation of the 
 interface and make this implementation available to the Java ServiceLoader (e.g. using Google's AutoService):
 - the `getName()` method should provide the plugin name - which can then be used in the `contingencyActivePowerLossDistribution` security analysis parameter
 - the `run(...)` method will be called by the security analysis engine for each contingency and should provide the logic.
-  This method has access to:
+  This method must return the amount of distributed active power (in per-unit) and has access to:
     - the network
     - the contingency in open-loadflow representation, including among others information about disconnected network elements, and how much active power has been lost.
     - the contingency definition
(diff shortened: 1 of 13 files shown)


src/main/java/com/powsybl/openloadflow/network/LfBranch.java
```diff
@@ -38,6 +39,66 @@ enum BranchType {
     record LfBranchResults(double p1, double p2, double q1, double q2, double i1, double i2) {
     }
 
+    class LfLimitsGroup {
+
+        private List<LfLimit> sortedLimits;
+
```
NathanDissoubray: In core we added a string id in the violations to identify the group of the branch that the violation originated from. This was not needed before since there would only ever be a single group activated at a time, but now that there are multiple, it could be difficult to know which one caused the issue.

I am not sure this would be required for a loadflow, but I'll mention it in case you think it could be useful @SylvestreSakti

PR #1352, opened 2026-02-23
Evolution of the sensitivity analysis ResultWriter and Result to support multi-component computation, improve diagnostic and support cancel

### LINKED ISSUE #1221: Sensitivity analysis: run simulation on all components

Currently, the sensitivity analysis is performed on the main synchronous component. It is not possible (without altering the network) to run the sensitivity analysis on other components.

Sensitivity analysis is run on all components of the network. Results are merged.

### Describe the motivation

It is perfectly legitimate to want to run a sensitivity anlysis on a component that is not the main synchronous one.

### Extra Information

A similar work has been done for the security analysis: [#1089](https://github.com/powsybl/powsybl-open-loadflow/pull/1089). The goal is to put the same effort into the sensitivity analysis.

snoopfab:

https://github.com/powsybl/powsybl-core/issues/3626

Implement features:
- Enable to add multi-component support for sensitivity analysis inOpenLoadFlow
- Enable to return partial results in case of interruption
- Enable a better diagnosis of contingency failure

https://github.com/powsybl/powsybl-core/issues/3626

- Enable to add multi-component support for sensitivity analysis inOpenLoadFlow
- Enable to return partial results in case of interruption
- Enable a better diagnosis of contingency failure

- [x] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
@@ -1327,6 +1327,27 @@ abstract class AbstractSensitivityAnalysis<V extends Enum<V> & Quantity, E exten
         return type.getSide().orElseThrow(() -> new PowsyblException("Cannot convert variable type " + type + " to a leg number"));
     }
 
+    protected LoadFlowParameters getAndCheckLoadFlowParameters(Network network, String workingVariantId, List<Contingency> contingencies, SensitivityFactorReader factorReader, SensitivityResultWriter resultWriter, ReportNode sensiReportNode) {
+        Objects.requireNonNull(network);
+        Objects.requireNonNull(contingencies);
+        Objects.requireNonNull(factorReader);
+        Objects.requireNonNull(resultWriter);
+        Objects.requireNonNull(sensiReportNode);
+        network.getVariantManager().setWorkingVariant(workingVariantId);
+        return parameters.getLoadFlowParameters();
+    }
+
+    static List<LfNetwork> getNetworksToSimulate(LfNetworkList networks, LoadFlowParameters.ComponentMode mode) {
+        return switch (mode) {
+            case MAIN_CONNECTED -> networks.getList().stream()
+                    .filter(n -> n.getNumCC() == ComponentConstants.MAIN_NUM && n.getValidity().equals(LfNetwork.Validity.VALID)).toList();
+            case MAIN_SYNCHRONOUS -> networks.getList().stream()
+                    .filter(n -> n.getNumSC() == ComponentConstants.MAIN_NUM && n.getValidity().equals(LfNetwork.Validity.VALID)).toList();
+            case ALL_CONNECTED -> networks.getList().stream()
+                    .filter(n -> n.getValidity().equals(LfNetwork.Validity.VALID)).toList();
+        };
+    }
+
     public abstract void analyse(Network network, String workingVariantId, List<Contingency> contingencies, List<OperatorStrategy> operatorStrategies,
                                  List<Action> actions, PropagatedContingencyCreationParameters creationParameters,
                                  List<SensitivityVariableSet> variableSets, SensitivityFactorReader factorReader,
--- a/src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
+++ b/src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
@@ -14,6 +14,7 @@ import com.powsybl.contingency.Contingency;
 import com.powsybl.contingency.strategy.OperatorStrategy;
 import com.powsybl.iidm.network.Network;
 import com.powsybl.loadflow.LoadFlowParameters;
+import com.powsybl.loadflow.LoadFlowResult;
 import com.powsybl.math.matrix.DenseMatrix;
 import com.powsybl.math.matrix.MatrixFactory;
 import com.powsybl.openloadflow.OpenLoadFlowParameters;
@@ -54,6 +55,8 @@ import java.util.stream.Collectors;
  * @author Gael Macherel {@literal <gael.macherel at artelys.com>}
  */
 public class AcSensitivityAnalysis extends AbstractSensitivityAnalysis<AcVariableType, AcEquationType> {
+
+    public static final String NO_IMPACT_ON_ANY_COMPONENT_S = "No impact on any component(s)";
 
     public AcSensitivityAnalysis(MatrixFactory matrixFactory, GraphConnectivityFactory<LfBus, LfBranch> connectivityFactory, SensitivityAnalysisParameters parameters) {
         super(matrixFactory, connectivityFactory, parameters);
@@ -121,14 +124,17 @@ public class AcSensitivityAnalysis extends AbstractSensitivityAnalysis<AcVariabl
             activePowerDistribution.run(lfNetwork, lfContingency.getActivePowerLoss());
         }
 
-        if (!runLoadFlow(context, false)) {
+        SensitivityAnalysisResult.LoadFlowStatus loadFlowResults = runLoadFlow(context);
+        if (!loadFlowResults.status().equals(LoadFlowResult.ComponentResult.Status.CONVERGED)) {
             // write contingency status
-            resultWriter.writeStateStatus(contingencyIndex, -1, SensitivityAnalysisResult.Status.FAILURE);
+            resultWriter.writeStateStatus(contingencyIndex, -1, SensitivityAnalysisResult.Status.FAILURE,
+                    loadFlowResults, lfNetwork.getNumCC(), lfNetwork.getNumSC());
             return;
         }
 
         // write contingency status
-        resultWriter.writeStateStatus(contingencyIndex, -1, SensitivityAnalysisResult.Status.SUCCESS);
+        resultWriter.writeStateStatus(contingencyIndex, -1, SensitivityAnalysisResult.Status.SUCCESS,
+                loadFlowResults, lfNetwork.getNumCC(), lfNetwork.getNumSC());
 
         // if we have at least one bus target voltage linked to a ratio tap changer, we have to rebuild the AC equation
         // system obtained just before the transformer steps rounding.
@@ -156,23 +162,28 @@ public class AcSensitivityAnalysis extends AbstractSensitivityAnalysis<AcVariabl
         calculateSensitivityValues(lfFactors, factorGroups, factorsStates, contingencyIndex, resultWriter);
     }
 
-    private static boolean runLoadFlow(AcLoadFlowContext context, boolean isRunningBaseSituation) {
+    private static SensitivityAnalysisResult.LoadFlowStatus runLoadFlow(AcLoadFlowContext context) {
         AcLoadFlowResult result = new AcloadFlowEngine(context)
                 .run();
         if (result.isSuccess() || result.getSolverStatus() == AcSolverStatus.NO_CALCULATION) {
-            return true;
+            return new SensitivityAnalysisResult.LoadFlowStatus(getStatusFromSolverStatus(result.getSolverStatus()), result.getSolverStatus().toString());
         } else {
-            if (isRunningBaseSituation) {
-                if (result.getOuterLoopResult().status() != OuterLoopStatus.STABLE) {
-                    throw new PowsyblException("Initial load flow of base situation ended with outer loop status " + result.getOuterLoopResult().statusText());
-                } else {
-                    throw new PowsyblException("Initial load flow of base situation ended with solver status " + result.getSolverStatus());
-                }
+            LOGGER.warn("Load flow failed with result={}", result);
+            if (result.getOuterLoopResult() != null && result.getOuterLoopResult().status() != OuterLoopStatus.STABLE) {
+                return new SensitivityAnalysisResult.LoadFlowStatus(LoadFlowResult.ComponentResult.Status.FAILED, result.getOuterLoopResult().statusText());
             } else {
-                LOGGER.warn("Load flow failed with result={}", result);
-                return false;
+                return new SensitivityAnalysisResult.LoadFlowStatus(getStatusFromSolverStatus(result.getSolverStatus()), result.getSolverStatus().toString());
             }
         }
+    }
+
+    private static LoadFlowResult.ComponentResult.Status getStatusFromSolverStatus(AcSolverStatus solverStatus) {
+        return switch (solverStatus) {
+            case CONVERGED -> LoadFlowResult.ComponentResult.Status.CONVERGED;
+            case NO_CALCULATION -> LoadFlowResult.ComponentResult.Status.NO_CALCULATION;
+            case MAX_ITERATION_REACHED -> LoadFlowResult.ComponentResult.Status.MAX_ITERATION_REACHED;
+            case SOLVER_FAILED, UNREALISTIC_STATE -> LoadFlowResult.ComponentResult.Status.FAILED;
+        };
     }
 
     /**
@@ -221,6 +232,7 @@ public class AcSensitivityAnalysis extends AbstractSensitivityAnalysis<AcVariabl
 
                 analyzeContingencySet(network, lfNetworks, propagatedContingencies, acParameters, lfParameters, lfParametersExt, variableSets, factorReader,
                         topoConfig.isBreaker(), resultWriter, variablesTargetVoltageInfo, sensitivityAnalysisParametersExt);
+                resultWriter.computationComplete();
             }
         } else {
             try (SequentialSensitivityResultWriter sequentialSensitivityResultWriter = new SequentialSensitivityResultWriter(resultWriter)) {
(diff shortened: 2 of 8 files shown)


snoopfab:
I rebased the `sensitivity-result-writer-enhancements` branches of **powsybl-core** and **powsybl-open-loadflow** onto `main`.

The rebases were really not trivial (an automatic merge was impossible).

While my branches were under review, Geoffroy’s PR **#3670** (operator strategies in the sensitivity API) was merged into core, along with its OLF counterpart **#1326**. These two PRs touched the same `SensitivityResultWriter` interface, the same `SensitivityAnalysisResult` class, and the same provider implementations as my work on multi‑component reporting — so we ended up with two parallel refactorings of the same surface area.

Rather than keeping the two APIs side by side, I merged my additions into the interface redesigned by Geoffroy. His design is centered around the new `SensitivityState` (a contingency + an optional operator strategy).

My additions — per‑component load‑flow status and partial result reporting in case of interruption — are a cross‑cutting concern that integrates naturally into this state.

So concretely, for the writer interface:

*   `writeSensitivityValue` gained the operator strategy index via #3670; nothing new on my side.
*   `writeContingencyStatus` and the post‑#3670 3‑argument `writeStateStatus` have both disappeared. They are replaced by:
        writeStateStatus(int contingencyIndex, int operatorStrategyIndex,
                         Status, LoadFlowStatus, int numCC, int numCS)
    This is a merge of his “which state failed?” with my “which load‑flow status on which component?”  
    Pre‑contingency reporting is done with both indices set to `-1`, which replaces my former `writeSynchronousComponentStatus`.
*   `computationComplete()` was added.

The final proposed contract is therefore:

```java
/**
 * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
 * @author Fabrice Buscaylet {@literal <fabrice.buscaylet at artelys.com}
 */
public interface SensitivityResultWriter {

    /**
     * In the case of multi-component computation, a factor has a variable
     * that is linked to an equipment that belongs to a single component.
     * Called once per factor and contingency in the factor context that
     * belongs to the same component as the factor equipment.
     * NaN for invalid factors whose equipment does not belong to a component
     * that has been run.
     *
     * @param factorIndex the factor index
     * @param contingencyIndex the contingency index, -1 for pre-contingency state
     * @param operatorStrategyIndex the operator strategy index, -1 if none
     * @param value the sensitivity value
     * @param functionReference the function reference
     */
    void writeSensitivityValue(int factorIndex, int contingencyIndex,
                               int operatorStrategyIndex, double value,
                               double functionReference);

    /**
     * Reports the status for a given state (contingency + optional operator strategy)
     * and a given component (numCC/numCS).
     * <p>
     * Called for every state and every numCC/numCS where the state has an impact.
     * For pre-contingency reporting, both {@code contingencyIndex} and
     * {@code operatorStrategyIndex} are -1.
     * For contingencies that are never run, called once in the end with NO_IMPACT
     * and numCC and numCS set to -1.
     *
     * @param contingencyIndex the contingency index, -1 for pre-contingency state
     * @param operatorStrategyIndex the operator strategy index, -1 if none
     * @param status the sensitivity analysis status
     * @param loadFlowStatus the load flow status for this component
     * @param numCC index of connected component, -1 if not applicable
     * @param numCS index of synchronous component, -1 if not applicable
     */
    void writeStateStatus(int contingencyIndex, int operatorStrategyIndex,
                          SensitivityAnalysisResult.Status status,
                          SensitivityAnalysisResult.LoadFlowStatus loadFlowStatus,
                          int numCC, int numCS);

    /**
     * Called at the end of the computation if the computation has not been interrupted.
     */
    void computationComplete();
}
```

On the result model side:

*   `SensitivityContingencyStatus` has disappeared and is replaced by `SensitivityStateStatus` from #3670, and I moved the per‑component load‑flow list onto this class.
*   `SensitivityPreContingencyStatus` has also disappeared; the pre‑contingency status is now a normal entry in `stateStatuses`, keyed by `SensitivityState.PRE_CONTINGENCY`.
*   The result JSON is bumped to version **1.2**; readers for v1.0 and v1.1 continue to work via the existing version‑aware deserializer.

On the OLF side:

*   AC analysis received the full multi‑component restructuring (loop per component via `getNetworksToSimulate`, aggregation of `NO_IMPACT` across all components, proper reporting of pre‑contingency load‑flow failure).
*   DC analysis received the same treatment. Operator strategies remain fully supported on both sides.

The tests pass: **61/61** on core, **284/284** on the OLF sensi package.

Two unrelated failures appeared, which look like upstream issues related to the core bump **7.2.0 → 7.3.0‑SNAPSHOT**:

*   `ZeroImpedanceFlowsTest` (the load flow gives 2.9 MW where 1.4 MW is expected, possibly due to the switch to JGraphT 1.5.3),
*   `DcSensitivityAnalysisContingenciesTest.testDebug` (XML IIDM format drift; the reference file just needs to be regenerated).

Practically speaking, since the commits were squashed and the branches had already been pushed to GitHub, the remote history no longer matched what I had locally. I therefore force‑pushed on the same branch.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -1327,6 +1327,27 @@ protected static int getLegNumber(SensitivityVariableType type) {
         return type.getSide().orElseThrow(() -> new PowsyblException("Cannot convert variable type " + type + " to a leg number"));
     }
 
+    protected LoadFlowParameters getAndCheckLoadFlowParameters(Network network, String workingVariantId, List<Contingency> contingencies, SensitivityFactorReader factorReader, SensitivityResultWriter resultWriter, ReportNode sensiReportNode) {
```
vidaldid-rte, line 1330: This method is never used. Should be removed.
(or if it is used, it needs to be renamed because it not a simple getter as it modifies the network variant)

src/main/java/com/powsybl/openloadflow/sensi/mt/SequentialSensitivityResultWriter.java
```diff
@@ -63,11 +63,17 @@ public void flush() {
     }
 
     @Override
-    public void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status) {
+    public void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status,
```
vidaldid-rte, line 66: As discussed in the core PR, the old SensitivityAnalysisResult.Status status param should be removed from the signature since it is redundant (and possibly in contradiction) with the loadflow status per component.

src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
```diff
@@ -1327,6 +1327,27 @@ protected static int getLegNumber(SensitivityVariableType type) {
...
+        Objects.requireNonNull(network);
+        Objects.requireNonNull(contingencies);
+        Objects.requireNonNull(factorReader);
+        Objects.requireNonNull(resultWriter);
+        Objects.requireNonNull(sensiReportNode);
+        network.getVariantManager().setWorkingVariant(workingVariantId);
+        return parameters.getLoadFlowParameters();
+    }
+
+    static List<LfNetwork> getNetworksToSimulate(LfNetworkList networks, LoadFlowParameters.ComponentMode mode) {
```
vidaldid-rte, line 1340: This duplicates the same static method in AbstractSecurityAnalysis.
Should be shared and moved in a place usable by both.

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
vidaldid-rte, line 266: With this line, it is impossible to run on several components. This needs to be updated.
And maybe to avoid reviewing all tests
  - the existing tests that change to much should force the loadFlow param to MAIN_SYNCHRONOUS. But is is good if possible to keep as many sa tests with the new default to ensure right coverage. 
  - Some tests should be added to run the MAIN_CONNECTED and ALL_CONNECTED cases with enough coverage (for example convergences that belong to multiiple components, some with convergence status different on each component etc...). In single threaded, and some in multi-threaded.

The PR should ne noted as breaking change because it will modify this default behaviour to always run on Main Synchronous. And that should be listed in the release notes when PR is merged.

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -307,140 +321,158 @@ private void analyzeContingencySet(Network network, LfNetworkList lfNetworks, Li
...
 
-            contingencies.forEach(contingency -> {
-                if (Thread.currentThread().isInterrupted()) {
-                    throw new PowsyblException("Computation was interrupted");
-                }
-                LOGGER.info("Simulate contingency '{}'", contingency.getContingency().getId());
-                contingency.toLfContingency(lfNetwork)
-                        .ifPresentOrElse(lfContingency -> {
+                // we always restart from base case voltages for contingency simulation
+                context.getParameters().setVoltageInitializer(new PreviousValueVoltageInitializer());
```
vidaldid-rte, line 400: Warning;
  This changes the voltageInitializer to previous value, as should be done to run contingencies.
   But now that the acParameters are reused for multiple LfNetworks, the value should be reinitialized to the original value before the next loop starts.

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -307,140 +321,158 @@ private void analyzeContingencySet(Network network, LfNetworkList lfNetworks, Li
...
         }
 
-        // create networks including all necessary switches
-        LfNetwork lfNetwork = lfNetworks.getLargest().orElseThrow(() -> new PowsyblException("Empty network"));
-
-        ReportNode networkReportNode = lfNetwork.getReportNode();
+        // Multi-component support: the analysis is run on every component selected by the configured ComponentMode.
+        // A contingency that has no impact on any component is reported once at the end with NO_CALCULATION.
+        Set<PropagatedContingency> contingenciesWithNoImpact = new LinkedHashSet<>(contingencies);
+        for (LfNetwork lfNetwork : getNetworksToSimulate(lfNetworks, acParameters.getNetworkParameters().getComponentMode())) {
```
vidaldid-rte, line 327: Warning: The acParameters might be modified in the loop (for example the voltageInitializer is set to previous values to run the contingencies). Something should be put in place to restore the acParameters to there initial state for each lfNetwork.
There is probably an existing pattern in security analysis as the same need exists there.

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -307,140 +321,158 @@ private void analyzeContingencySet(Network network, LfNetworkList lfNetworks, Li
...
-                            ReportNode postContSimReportNode = Reports.createPostContingencySimulation(networkReportNode, lfContingency.getId());
-                            lfNetwork.setReportNode(postContSimReportNode);
+                OpenLoadFlowParameters contingencylfParametersExt = applyGenericContingencyParameters(context, lfParameters, lfParametersExt,
+                        sensitivityAnalysisParametersExt.isStartWithFrozenACEmulation());
 
-                            List<LfSensitivityFactor<AcVariableType, AcEquationType>> contingencyFactors = validFactorHolder.getFactorsForContingency(lfContingency.getId());
-                            contingencyFactors.forEach(lfFactor -> {
-                                lfFactor.setSensitivityValuePredefinedResult(null);
-                                lfFactor.setFunctionPredefinedResult(null);
+                contingencies.forEach(contingency -> {
```
vidaldid-rte, line 405: You should filter the contingencies that 
  - touche an element of the component
  -  have at least one factor in the component.
There is no point in running the contingencies that have no factor in the component.
It is a waste of time. 
Note also that the idea per https://github.com/powsybl/powsybl-core/issues/3626 is to call writeSensitivityValue only once per contingency and per factor (since a factor belongs to only one component)
That should be implemented. And a unit test should verify that this contract is respected.

The contingencies and factors that have not be run on any component then be notified to the writer as invalid factors.

src/test/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysisContingenciesTest.java
```diff
@@ -1248,16 +1249,52 @@ void testNoImpactContingencyAfterNormalContingency() {
     @Test
     void testMaxIterationReachedAfterContingency() {
         Network network = EurostagFactory.fix(EurostagTutorialExample1Factory.create());
-        network.getLine("NHV1_NHV2_1").setX(1000);
+        network.getLine("NHV1_NHV2_1").setX(100000);
```
vidaldid-rte, line 1252: What is the need to change this value ? That should be independant from this PR.

src/test/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysisContingenciesTest.java
```diff
@@ -1248,16 +1249,52 @@ void testNoImpactContingencyAfterNormalContingency() {
...
         List<Contingency> contingencies = List.of(new Contingency("NHV1_NHV2_2", List.of(new BranchContingency("NHV1_NHV2_2"))));
         List<SensitivityFactor> factors = List.of(createBranchFlowPerInjectionIncrease("NHV1_NHV2_1", "LOAD"));
         SensitivityAnalysisParameters sensiParameters = new SensitivityAnalysisParameters();
         sensiParameters.getLoadFlowParameters().setDistributedSlack(false);
         SensitivityAnalysisRunParameters runParameters = new SensitivityAnalysisRunParameters()
                 .setContingencies(contingencies)
                 .setParameters(sensiParameters);
-        SensitivityAnalysisResult result = sensiRunner.run(network, factors, runParameters);
-        assertEquals(SensitivityAnalysisResult.Status.FAILURE, result.getStateStatus(SensitivityState.postContingency("NHV1_NHV2_2")));
+
```
vidaldid-rte, line 1260: Why did you have to change this one liner by a very complex code ?
The multi-component support should not make the API harder to use for simple cases (when a contingency touches only one network component)

src/test/java/com/powsybl/openloadflow/sensi/DcSensitivityAnalysisTest.java
```diff
@@ -1365,4 +1371,70 @@ void testComputationInterrupted() {
...
+        Network network = FourBusNetworkFactory.create();
+        SensitivityAnalysisParameters sensiParameters = createParameters(true, "b1_vl_0", true);
+        List<SensitivityFactor> factors = List.of(createBranchFlowPerInjectionIncrease("l14", "g1"));
+        SensitivityFactorReader factorReader = new SensitivityFactorModelReader(factors, network);
+
+        AtomicInteger valueCallCount = new AtomicInteger(0);
+        AtomicInteger syncStatusCallCount = new AtomicInteger(0);
+        AtomicInteger computationCompleteCallCount = new AtomicInteger(0);
+        List<LoadFlowResult.ComponentResult.Status> syncStatuses = new ArrayList<>();
+        SensitivityResultWriter resultWriter = new SensitivityResultWriter() {
```
vidaldid-rte, line 1386: There should be a simple way to tell if a LF failed. It should be a on liner.

src/test/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysisTest.java
```diff
@@ -130,16 +131,30 @@ void testEsgTutoMT() throws IOException {
...
+        AtomicInteger syncStatusCallCount = new AtomicInteger(0);
+        AtomicInteger computationCompleteCallCount = new AtomicInteger(0);
         SensitivityResultWriter resultWriter = new SensitivityResultWriter() {
 
             public void writeSensitivityValue(int factorIndex, int contingencyIndex, int operatorStrategyIndex, double value, double functionReference) {
                 valueCallCount.incrementAndGet();
             }
 
             @Override
-            public void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status) {
```
vidaldid-rte, line 140: This should not be modified if contingencies are run only when necessary (on a component they modify and when they have a valid factor)

src/test/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysisTest.java
```diff
@@ -2163,19 +2182,61 @@ void testFailingLf() {
...
+            public void computationComplete() {
+                computationCompleteCallCount.incrementAndGet();
+            }
+        };
 
         OpenLoadFlowParameters olfParameters = sensiParameters.getLoadFlowParameters().getExtension(OpenLoadFlowParameters.class);
         olfParameters.setMaxNewtonRaphsonIterations(1);
         SensitivityAnalysisRunParameters runParameters = new SensitivityAnalysisRunParameters()
                 .setParameters(sensiParameters);
-        CompletionException e = assertThrows(CompletionException.class, () -> sensiRunner.run(network, factors, runParameters));
```
vidaldid-rte, line 2171: With the new API, this should remain simple.

vidaldid-rte (changes requested):
This is not ready to be merged.

The run on multiple component is not testable (component mode still forced)
The acParameters need to be restored for each lfCompoent.
The contingencies should be run only when needed.

The API should remain simple for simple cases. For a simple contingency that touchs only one component there is no reason to make it harder to find its status.

The tests that have contingencies touching only the main component should not have to be modified (or marginally to update the API).

More tests are needed to test the behaviour for multi components.

PR #1368, opened 2026-03-12
Add reports for AC Load Flow voltage initialization

jeandemanged:

No

Feature

There are no reports for AC Load Flow bus voltage initialization. This wouldn't be a real problem if the DC_VALUES voltage initializer was not issuing reports entries making the overall report of an AC LF run confusing/misleading, e.g.

```
           + Load flow on network 'ieee9cdf'
              + Network CC0 SC0
                + Network info
                    Network has 9 buses and 9 branches
                    Network balance: active generation=319.64102 MW, active load=315 MW, reactive generation=0 MVar, reactive load=115 MVar
                    Angle reference bus: VL1_0
                    Slack bus: VL1_0
                 Slack bus active power (-4.64102 MW) distributed in 1 distribution iteration(s)
                 DC load flow completed (solverSuccess=true, outerloopStatus=STABLE)
```

- Specific report entries when voltage initialization is done, mentioning the initialization method
- In the case of security analysis, no report is made about previous values voltage initialization for contingency and operator strategy simulations - this would be of no value and would clutter the report unnecessarily. Internally to OLF, use AcLoadFlowParameters.voltageInitReport boolean to control whether the report is to be created or not.
- Implementing above distinction between basecase simulation and contingency/operator strategy simulation revealed a bug in the case of multiple AC islands / synchronous components in security analysis: In the case of slackDistributionFailureBehavior being set to THROW:
  - the first LfNetwork would use THROW for basecase simulation, then switch to FAIL for contingency/operator strategy simulation
  - all subsequent LfNetwork simulations would use FAIL for basecase simulation - which is inconsistent
  - this has been fixed as part of this PR: AcLoadFlowParameters are now copied for each LfNetwork run from the initial values

```
           + Load flow on network 'ieee9cdf'
              + Network CC0 SC0
                 + Network info
                    Network has 9 buses and 9 branches
                    Network balance: active generation=319.64102 MW, active load=315 MW, reactive generation=0 MVar, reactive load=115 MVar
                    Angle reference bus: VL1_0
                    Slack bus: VL1_0
                 + Voltage initialization with method Full Voltage
                    Voltage initialization with method Voltage Magnitude
                    + Voltage initialization with method DC value
                       Slack bus active power (-4.64102 MW) distributed in 1 distribution iteration(s)
                       DC load flow completed (solverSuccess=true, outerloopStatus=STABLE)
```

- [x] No

**Other information**:


src/main/java/com/powsybl/openloadflow/dc/DcValueVoltageInitializer.java
```diff
@@ -24,6 +26,8 @@
  */
 public class DcValueVoltageInitializer implements VoltageInitializer {
 
+    public static final String NAME = "DC value";
```
SylvestreSakti: Same ?
```suggestion
    public static final String NAME = "DC Values";
```

PR #1386, opened 2026-04-10
Add support of network cache for DC load flow

SylvestreSakti:

No

**networkCacheEnabled** parameter was introduced in #635 to support AC load flow fast restart. This PR adds support of DC load flow fast restart.

DC load flow fast restart not supported

DC load flow fast restart supported (leading to more than two times faster execution of multiple load flow)

After introducting DC load flow fast restart, it was observed that resetState (putting all variables to NaN) and updateState (updating all state variables including V) are taking more than the half of time. An optimization is done to resetState only when needed (when new NetworkCache is used), and not updating V state variables (which is useless in DC calculation).

At all, this leads to more than 3 times faster calculation.

- [ ] Yes
- [x] No

**Other information**:

For the moment, network cache updating is equivalent in AC or in DC. This could be further modified in the future to avoid resetting in DC for non impacting modifications such as voltage targets modifications


src/main/java/com/powsybl/openloadflow/ac/AcLoadFlowContext.java
```diff
@@ -30,8 +30,6 @@ public class AcLoadFlowContext extends AbstractLoadFlowContext<AcVariableType, A
 
     private AcLoadFlowResult result;
 
-    private boolean networkUpdated = true;
```
geofjamg: I'm not sure this is a better place in LFNetwork. Maybe in Cache.Entry ?

src/main/java/com/powsybl/openloadflow/fastrestart/NetworkCache.java
```diff
@@ -49,7 +52,8 @@ public static class Entry extends DefaultNetworkListener {
 
         private final LoadFlowParameters parameters;
 
-        private List<AcLoadFlowContext> contexts;
+        private HashMap<LfNetwork, LfNetworkParameters> lfNetworksAndParameters;
```
geofjamg: Why not keeping the AbstractLoadFlowContext which contains the LfNetwork and parameters for both AC and DC ?

src/main/java/com/powsybl/openloadflow/fastrestart/NetworkCache.java
```diff
@@ -436,41 +441,70 @@ public void close() {
...
+
+    private final List<Entry<DcLoadFlowContext>> dcEntries = new ArrayList<>();
 
     private final Lock lock = new ReentrantLock();
 
     private void evictDeadEntries() {
-        Iterator<Entry> it = entries.iterator();
-        while (it.hasNext()) {
-            Entry entry = it.next();
+        Iterator<Entry<AcLoadFlowContext>> itAc = acEntries.iterator();
```
geofjamg: We probably can better mutualize between AC and DC without duplicating

src/main/java/com/powsybl/openloadflow/fastrestart/NetworkCache.java
```diff
@@ -436,41 +441,70 @@ public void close() {
...
             evictDeadEntries();
-            return entries.size();
+            return acEntries.size();
         } finally {
             lock.unlock();
         }
     }
 
-    public Optional<Entry> findEntry(Network network) {
+    public int getDcEntryCount() {
```
geofjamg: Same

src/main/java/com/powsybl/openloadflow/fastrestart/NetworkCache.java
```diff
@@ -436,41 +441,70 @@ public void close() {
...
+        lock.lock();
+        try {
+            evictDeadEntries();
+            return dcEntries.size();
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    public Optional<Entry<AcLoadFlowContext>> findEntryAc(Network network) {
```
geofjamg: Same

src/main/java/com/powsybl/openloadflow/fastrestart/NetworkCache.java
```diff
@@ -505,19 +546,38 @@ public Entry get(Network network, LoadFlowParameters parameters) {
...
             }
         } else {
-            LOGGER.info("Network cache cannot be reused for network '{}' because invalided", network.getId());
+            LOGGER.info("Network cache for AC load flow cannot be reused for network '{}' because invalided", network.getId());
+        }
+
+        return entry;
+    }
+
+    public Entry<DcLoadFlowContext> getDc(Network network, LoadFlowParameters parameters) {
```
geofjamg: Here too

geofjamg (changes requested):

SylvestreSakti:
Finally, this PR has been rebased to follow the network cache refactoring in PR #1421

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
```diff
@@ -207,19 +208,34 @@ private void computeZeroImpedanceFlows(LfNetwork network, LoadFlowModel loadFlow
...
-                .setCacheEnabled(false); // force not caching as not supported in DC LF
 
-        List<DcLoadFlowResult> results = DcLoadFlowEngine.run(network, new LfNetworkLoaderImpl(), dcParameters, reportNode);
-
-        Networks.resetState(network);
+        List<DcLoadFlowResult> results;
+        final boolean updateV;
+        if (parametersExt.isNetworkCacheEnabled()) {
+            if (NetworkCache.DC_LF_INSTANCE.findEntry(network).map(NetworkCache.Entry::getValues).orElse(null) == null) {
+                Networks.resetState(network); // reset state only if new NetworkCache created
```
geofjamg: Could we have side effect by resetting only if cache is invalidated? For example if because if a switch change, some elements are out of main CC, their SV won't be reset.

src/test/java/com/powsybl/openloadflow/LoadFlowWithCachingTest.java
```diff
@@ -223,82 +230,67 @@ void testEntryEviction() {
...
         loadFlowRunner.run(network, parameters);
-        assertNotNull(NetworkCache.AC_LF_INSTANCE.findEntry(network).orElseThrow().getValues());
+        assertNotNull(findEntryFunction.apply(network, isDc).getValues());
 
         network.getVariantManager().removeVariant("v");
-        assertNotNull(NetworkCache.AC_LF_INSTANCE.findEntry(network).orElseThrow().getValues());
-    }
-
-    @Test
-    void testUpdateOnDifferentVariantDoesNotInvalidateCache() {
```
geofjamg, line 284: Why removing this test ?

SylvestreSakti: It was not intended, this is fixed, thanks !

src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
SylvestreSakti: You are right,

At first my thought was that even with switch actions, since the reconnectable elements should be in the LfNetwork, they already have been resetted (even if disabled) at cache creation. But if they have been set later, removing them will indeed not invalidate it. 

Maybe a second boolean should be introduced in the NetworkCache, to tell if the network should be reset or not ? This would be used if they are switch updates, to reset the state without invalidating the cache

SylvestreSakti: The case has been solved, this led to detect a bug in Switch update (when updating mutiple switches) that has been fixed.

geofjamg (dismissed):

geofjamg (approved):

PR #1388, opened 2026-04-13
Transformer continuous voltage control outer loop

geofjamg:

No

Feature

When simulating transformer voltage control, we can only simulate discrete voltage ratio variation based on IIDM ratio tap changer steps.

We can now simulate transformer voltage control in a continuous way. As IIDM model is not able to store continuous ratio output, we use a property attached to the transformer.

- [ ] Yes
- [ ] No

**Other information**:


SylvestreSakti (commented):
Hello, this is a nice feature to get the continuous ratio. I am just wondering if is needed, in the use case, to handle parallel ratio tap changers (the parameter useInitialTapPositoin that is present in AFTER_GENERATOR_VOLTAGE_CONTROL_MODE).

Otherwise, just needs some documentation. 

If this feature gets very useful for lots of users in the future, we could think of an optional solvedRatio in the IIDM but not sure we have to get to that.

PR #1389, opened 2026-04-13
Fix DC Load Flow running slack distribution for epsilon mismatch

jeandemanged:

No

Bug fix

In DC Load Flow slack distribution is always performed.
If the mismatch is epsilonesque (lower than P_RESIDUE_EPS=1e-3 MW) this is an unnecessary slack distribution and generates slack distribution reports + logs.
Especially annoying in DC security analysis where each and every contingency not causing an injection loss issues a report+ log showing zero or near zero mismatch distributions.

DC Load Flow slack distribution is executed only if mismatch is above epsilon (P_RESIDUE_EPS).

- [x] No


src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -1955,7 +1955,28 @@ void reportTest() throws IOException {
...
+        List<Contingency> contingencies = createAllBranchesContingencies(network);
+
+        ReportNode reportNode = ReportNode.newRootReportNode()
+                .withResourceBundles(PowsyblOpenLoadFlowReportResourceBundle.BASE_NAME, PowsyblTestReportResourceBundle.TEST_BASE_NAME)
+                .withMessageTemplate("TestSecurityAnalysis")
+                .build();
+
+        LoadFlowParameters loadFlowParameters = new LoadFlowParameters()
+                .setDc(true)
+                .setHvdcAcEmulation(false);
```
jeandemanged: to not have the outerloop in the report, but probably a fix to do in another PR to not create the outerloop if no HVDC in DC emulation in the network

SylvestreSakti: Hello, indeed there was a bug displaying this outer loop even when not needed. The fix is made in #1397 so these lines can be removed (and this PR will be merged after #1397)

jeandemanged: thank you for the fix @SylvestreSakti, I adjusted accordingly

SylvestreSakti (approved):

PR #1394, opened 2026-05-03
Fix vectorization issue when reconnecting lines

geofjamg:

No

Bug fix

When reconnecting a component (thanks to an actions in security analysis and sensi analysis), we get an error in KLU factorization when using the sparse matrix factory. There is no bug using dense matrix factory. We also only get this error using the vectorized equations.

This is because a bug in equation array that do not fill matrix term in the right order. Some elements (row) in a column are updated several times but not consecutively as required by the sparse matrix implementation. The sparse matrix created is then corrupted leading to an error at first factorization. 

Terms are sorted in a more robust way not based in row num (which could be -1 initialliy if element is disconnected) but using the variable hash allowing a right order of elements even with disconnected elements. 

- [ ] Yes
- [x] No

**Other information**:


SylvestreSakti (dismissed):

src/main/java/com/powsybl/openloadflow/equations/EquationArray.java
```diff
@@ -425,8 +425,8 @@ private void addEquationDerivativeVectorSortedTerms(int elementNum, List<Equatio
                 }
             }
         }
-        // sortByVariableRowAndVectorizedLocalNum
-        terms.sort(Comparator.comparingInt(o -> o.derivative.getVariable().getRow()));
+        // Terms do not have a row initially so they are sorted by variable hash code
+        terms.sort(Comparator.comparingInt(o -> o.derivative.getVariable().hashCode()));
```
SylvestreSakti, line 429: As a more detailed explanation to this fix : 

## How the derivatives in the vectorization are computed : 
During a computation, vectorized equations derivatives are handled this way :

- At initialization (EquationSystem creation), the derivatives are stored in the **EquationDerivativeVector**. The terms are sorted by column and by row. And this EquationDerivativeVector **remains the same until the end of the simulation**.

- At each AC solving computation, the **JacobianMatrix** is constructed through the method initDer. To do so, for the vectorized part, the terms are read from the EquationDerivativeVector in the order, and the matrix is filled with the derivative values. This matrix **is reconstructed whenever equation terms activation has to change** (between some outer loop iterations)
 
- At each Newton-Raphson iteration, the **JacobianMatrix** values are updated through the method updateDer() without any struture change.

## What is the issue of that :
With sparse matrices, at reconstruction (initDer), values have to be filled in the column order (otherwise an exception is thrown). For each column, rows can be given in any order, but **SparseMatrix does not handle filling two times the same row**, without throwing any exception (doing this check would cost too much performance at each value filling). 
Since the terms are sorted by row in the EquationDerivativeVector, this is correctly handled unless in one case : In the security analysis with reconnection action **since the EquationDerivativeVector is only constructed at the beginning, some of the terms that will be activated in the future do not have yet a row number (it is set at -1), and so these are put first. Even after activating them, their row number are updated, but they remain first in the EquationDerivativeVector during each initDer()**. This leads to fill two times the same row when constructing the JacobianMatrix.

## The solution : 
By sorting the variables by hashCode instead of rowNumber at the beginning, we do not care about activation state of the terms. There is no necessity of having rows sorted by order, but we need to have terms that have the same row number to be contiguois in the EquationDerivativeVector. We ensure that using hashCode.

Hadrien-Godard, line 429: I am not sure, but is this risking to break reproducibility?

geofjamg:
```java   
public int hashCode() {
        return elementNum + type.hashCode();
    }
```
This is how variable hashcode is calculated, so this is stable

Hadrien-Godard (dismissed):

SylvestreSakti (approved):

Hadrien-Godard (approved):

PR #1398, opened 2026-05-05
Filtering monitored results in N-1 state result

HugoKulesza:

No

Memory issue can be encountered when state monitored results are heavy. This feature aims at reducing memory use by filtering results that indicate no difference between base case situation and contingency situation.

No filtering possible

During security analysis, monitored elements results can now be filtered when they do not change between N and N-1 states.
This filtering is done using the parameters added in https://github.com/powsybl/powsybl-core/pull/3903

- [ ] Yes
- [x] No


jeandemanged:
Hi @HugoKulesza , could you please resolve the merge conflict ?

src/main/java/com/powsybl/openloadflow/sa/AbstractNetworkResult.java
```diff
@@ -63,7 +56,9 @@ protected AbstractNetworkResult(LfNetwork network, StateMonitorIndexes monitorIn
         this.dcPowerFactor = dcPowerFactor;
     }
 
-    protected void addResults(StateMonitor monitor, Consumer<LfBranch> branchConsumer, Predicate<LfBranch> isBranchDisabled, Map<String, LfBranch.LfBranchResults> zeroImpedanceFlows) {
+    protected void addResults(StateMonitor monitor, Consumer<LfBranch> branchConsumer, Predicate<LfBranch> isBranchDisabled,
+                              Consumer<LfBus> busConsumer, Consumer<String> threeWindingsTransformerResultsConsumer,
+                              Map<String, LfBranch.LfBranchResults> zeroImpedanceFlows) {
```
jeandemanged: the sonar issue seems relevant, zeroImpedanceFlows can be removed

src/main/java/com/powsybl/openloadflow/sa/PostContingencyNetworkResult.java
```diff
@@ -23,43 +28,82 @@
...
 
     private final List<BranchResult> branchResults = new ArrayList<>();
+    private final List<BusResult> busResults = new ArrayList<>();
+    private final List<ThreeWindingsTransformerResult> threeWindingsTransformerResults = new ArrayList<>();
 
     private final PreContingencyNetworkResult preContingencyMonitorInfos;
 
     private final Contingency contingency;
 
+    private final ModifiedMonitoredElementsParameters modifiedMonitoredElementsParameters;
```
jeandemanged, line 37: why do we need to store the parameters in the PostContingencyNetworkResult ? This should rather be a parameter of `public void addResults(...)`

jeandemanged: I see that in AbstractNetworkResult we also store parameters (loadFlowModel, dcPowerFactor) ... forget this

src/main/java/com/powsybl/openloadflow/sa/PostContingencyNetworkResult.java
```diff
@@ -23,43 +28,82 @@
...
+                    preContingencyBranchOfContingencyP1 = preContingencyBranchOfContingencyResult.getP1();
                 }
             }
-            branchResults.addAll(branch.createBranchResult(preContingencyBranchP1, preContingencyBranchOfContingencyP1, createResultExtension, zeroImpedanceFlows, loadFlowModel));
-        }, isBranchDisabled, zeroImpedanceFlows);
+        }
+        if (preContingencyBranchResult != null && Math.abs((preContingencyBranchP1 - branch.getP1().eval() * PerUnit.SB) / preContingencyBranchP1)
+                < modifiedMonitoredElementsParameters.getPowerModificationThreshold()) {
+            return;
+        }
```
jeandemanged: issues I see here:

1. branch.getP1().eval() will not work for zero impedance branches
2. checking only P1, should check also P2
3. https://github.com/powsybl/powsybl-core/pull/3903 mentions just filtering on "power" without saying anything about active or reactive. Adding reactive power in the check would make sense from my point of view. If you want to limit the check on active power, this should be documented somewhere (in core or OLF). In fact it is a pity that https://github.com/powsybl/powsybl-core/pull/3903 named the parameter `power` and not `flow` as done in IncreasedViolationsParameters: One would want also to include current in the filtering check.
4. https://github.com/powsybl/powsybl-core/pull/3903 documents `powerModificationThreshold` as a percentage, and I do not see any division by 100

Issue 1. should be solvable by calling branch.createBranchResult(...) first, and checking values obtained from there to decide whether the BranchResult has to be added in the list.

src/main/java/com/powsybl/openloadflow/sa/PostContingencyNetworkResult.java
```diff
@@ -23,43 +28,82 @@
...
+        branchResults.addAll(branch.createBranchResult(preContingencyBranchP1, preContingencyBranchOfContingencyP1, createResultExtension, zeroImpedanceFlows, loadFlowModel));
+    }
+
+    private void createBusResults(LfBus bus) {
+        List<BusResult> unfilteredBusResults = bus.createBusResults();
+        for (BusResult busResult : unfilteredBusResults) {
+            var preContingencyBusResult = preContingencyMonitorInfos.getBusResult("%s_%s".formatted(busResult.getVoltageLevelId(), busResult.getBusId()));
+            if (preContingencyBusResult != null) {
+                double threshold = modifiedMonitoredElementsParameters.getVoltageModificationThreshold(preContingencyBusResult.getV());
```
jeandemanged, line 104: same as previous comment, [#3903](https://github.com/powsybl/powsybl-core/pull/3903) documents `voltageModificationAbsoluteThreshold` as a percentage, and I do not see any division by 100. But here the issue is on powsybl-core side.

src/main/java/com/powsybl/openloadflow/sa/PostContingencyNetworkResult.java
```diff
@@ -23,43 +28,82 @@
...
+
+    private void create3WTransformerResults(String id, Map<String, LfBranch.LfBranchResults> zeroImpedanceFlows) {
+        LfLegBranch leg1 = (LfLegBranch) network.getBranchById(LfLegBranch.getId(id, 1));
+        var preContingencyResult = preContingencyMonitorInfos.getThreeWindingsTransformerResult(id);
+        double preContingencyLeg1P = preContingencyResult != null ? preContingencyResult.getP1() : Double.NaN;
+
+        if (preContingencyResult != null && Math.abs((preContingencyLeg1P - leg1.getP1().eval() * PerUnit.SB) / preContingencyLeg1P)
+                < modifiedMonitoredElementsParameters.getPowerModificationThreshold()) {
+            return;
+        }
```
jeandemanged: same comment as for the branch case (zero impedance branch issue, should check all 3 legs, ambiguity about checking Q and I and ambiguity about percentage)

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
```diff
@@ -2940,18 +2942,43 @@ void testBusBarSectionBusResults() {
...
-        assertEquals("BBS1", preContingencyResult.getNetworkResult().getBusResults().get(0).getBusId());
-        assertEquals(400.0, preContingencyResult.getNetworkResult().getBusResult("BBS1").getV(), LoadFlowAssert.DELTA_V);
+        assertEquals("BBS3", preContingencyResult.getNetworkResult().getBusResults().get(0).getBusId());
+        assertEquals(400.0, preContingencyResult.getNetworkResult().getBusResult("BBS3").getV(), LoadFlowAssert.DELTA_V);
         assertEquals("BBS2", preContingencyResult.getNetworkResult().getBusResults().get(1).getBusId());
         assertEquals(400.0, preContingencyResult.getNetworkResult().getBusResult("BBS2").getV(), LoadFlowAssert.DELTA_V);
-        assertEquals("BBS3", preContingencyResult.getNetworkResult().getBusResults().get(2).getBusId());
-        assertEquals(400.0, preContingencyResult.getNetworkResult().getBusResult("BBS3").getV(), LoadFlowAssert.DELTA_V);
+        assertEquals("BBS1", preContingencyResult.getNetworkResult().getBusResults().get(2).getBusId());
+        assertEquals(400.0, preContingencyResult.getNetworkResult().getBusResult("BBS1").getV(), LoadFlowAssert.DELTA_V);
```
jeandemanged: looks like we do have random results ordering, probably changing the HashMap-s to LinkedHashMap would solve

src/main/java/com/powsybl/openloadflow/sa/PostContingencyNetworkResult.java
jeandemanged: Hi @HugoKulesza, @SylvestreSakti,
cc @olperr1 
From my point of view https://github.com/powsybl/powsybl-core/pull/3903 should be reworked before final release of powsybl-core to:
- rename `powerModificationThreshold` to `flowModificationThreshold` (really mirroring IncreasedViolationsParameters)
- document that `flowModificationThreshold` applies to P, Q, I. Filtering only on P makes the feature useless to me.
- fix documentation ambiguity about percentage / not percentage

Let me know what you think, I could propose a PR in powsybl-core in that direction.

SylvestreSakti: Hello, not thought about it but it is indeed necessary to keep the same amount of information. Filtering only on P may indeed trigger missing changes if it happens for Q (less probably for I since it is correlated)

I agree with the changes proposed by @jeandemanged

obrix: I agree with the need of a filtering on both sides and also current and reactive power. It would be really useful for us to have these features.

jeandemanged: Hello @SylvestreSakti , @HugoKulesza , @obrix , cc @olperr1 ,

I did tests on a real case:
* ~20.5k buses, ~28.5k branches
* 700 contingencies, all at 400 kV, all on main component, connected both sides
* 7500 branches monitored, all at or above 220 kV, all on main component, connected both sides
* all 3WT (~1300)

I tested filtering of P & Q only (both sides for branches, all 3 sides for 3WT), comparing relative/proportional vs. absolute thresholds. Here results:

image

We can see that absolute filter performs better, we get less reported elements. Also, for real use cases relative filtering is a pain. Even with 1%, for a flow of 1000 MW this means a change of 10 MW not being reported. With absolute filtering, we can just put the convergence criteria of 0.1 or 1.0 MW / MVAr, and besides a better filtering, we have a strong guarantee that "things did not move". Reducing memory footprint is achieved in both cases, but having results that can actually be used is much more important. Having relative/proportional makes sense for worsened violations filtering, because you are near the limit. But for state monitors, absolute thresholds in MW and MVAr is what makes sense.

In the end I would propose:
- in powsybl-core: do not change the name of existing parameter `powerModificationThreshold`, only change its documentation to be an absolute MW/MVAr change.
- I think MW/MVAr filtering is enough, no need to add current change threshold

Let me know what you think about it

jeandemanged: I pushed my branches if you want to try on your network:
- filter-monitored-results-jd-absolute (threshold hardcoded in PostContingencyNetworkResult)
- filter-monitored-results-jd-relative

SylvestreSakti: Thanks for the benchmark ! All these arguments seem relevant, to my mind, it seems better to quantify the change in absolute indeed.

I don't know if there was a reason to want it in percent ? @HugoKulesza ?

Moreover, it seems easier to change in powsybl-core only the documentation :)

SylvestreSakti: After discussion with one user, it seems relevant to have a percentage for him (or a current threshold to filter the active power) because the absolute threshold in 400kV would not be the same as the one in 63 kV. Maybe another parameter to add (in a future release after discussion ?)

For now, it seems better for me to have it the way you suggest @jeandemanged

Vyki:
Dear colleagues,

Thank you for suggesting this feature. We have also observed that the memory requirements of the current CA are unsustainable for larger studies.

I've also been testing ways to solve this problem, but my approach is a little different; theoretically, it could complement your approach.

The main idea is to extend StateMonitor with a resultMode attribute, which specifies that for the ALL option, NetworkResult must always be present, and for VIOLATION_ONLY, it is present only when a limit violation is recorded. Theoretically, this could be extended to include a third option, MODIFIED_ONLY. This would allow for very fine-grained control at the StateMonitor level, thereby providing flexibility for a wider range of use cases.

powsybl-core extension: https://github.com/Vyki/powsybl-core/commit/75f7efa8c31fad4c9f3fdd4e466603f490d7d853
powsybl-open-loadflow extension: https://github.com/Vyki/powsybl-open-loadflow/commit/c351c60a4ecb2bd19ba8909e43930127d2837108

Should I move this discussion to a separate issue ticket?

Thanks,
BR Roman

src/test/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisTest.java
SylvestreSakti: Using a LinkedHashMap for busResults indeed solves this issue

SylvestreSakti:
> Dear colleagues,
> 
> Thank you for suggesting this feature. We have also observed that the memory requirements of the current CA are unsustainable for larger studies.
> 
> I've also been testing ways to solve this problem, but my approach is a little different; theoretically, it could complement your approach.
> 
> The main idea is to extend StateMonitor with a resultMode attribute, which specifies that for the ALL option, NetworkResult must always be present, and for VIOLATION_ONLY, it is present only when a limit violation is recorded. Theoretically, this could be extended to include a third option, MODIFIED_ONLY. This would allow for very fine-grained control at the StateMonitor level, thereby providing flexibility for a wider range of use cases.
> 
> powsybl-core extension: [Vyki/powsybl-core@75f7efa](https://github.com/Vyki/powsybl-core/commit/75f7efa8c31fad4c9f3fdd4e466603f490d7d853) powsybl-open-loadflow extension: [Vyki@c351c60](https://github.com/Vyki/powsybl-open-loadflow/commit/c351c60a4ecb2bd19ba8909e43930127d2837108)
> 
> Should I move this discussion to a separate issue ticket?
> 
> Thanks, BR Roman

Hello Roman,

Thank you for your suggestion, this seems indeed interesting, and the resultMode functionnality may indeed cover the use case of this PR. This discussion could indeed be moved in a separated issue.

More generally, we should probably discuss about the way of storing/displaying the security analys results. This PR and your proposal both try to answer to this main topic. Maybe sending results with a result writer similar to SensitivityAnalysis (that sends the data during the simulation) would be a more generic way to tackle all these issues of memory/clarity of the results. The work started in https://github.com/powsybl/powsybl-core/pull/3625 seems to be a first answer to discuss about

SylvestreSakti (approved):
Ok with this PR ! As discussed with @jeandemanged, he reminded that lost elements do not appear in the state monitored results. Hence a risk of confusion introduced when filtering results: a missing result can mean either a result that did not change, either a lost element. This can be clarified for the user by crossing these results with the connectivity results.

In the future, this should be discussed to have something less confusing (for example, giving NaN results for lost elements ?) But it may be ok to merge the PR as it for now

PR #1405, opened 2026-05-12
DC sensitivity analysis fast restart

geofjamg:

No

Feature

Like for the DC and AC loadflow we have a fast restart model for DC sensi analysis. A cache is kept to avoid when running multiple DC sensi analysis on thye same network and variant to re-create identical intermediate data structure (LF network, matrices, etc).
In addition, has been added:
 - a network variant pool mode: unlike DC and AC loadflow fast restart where when we need to load the LF network from the bus/breaker view (when there is overload management systems), we do not create on the fly an IIDM variant (and remove it), we rely on a network variant pool to pre-create variants and then sensi analyses peek a variant a the pool. This avoid IIDM multithreading variant issues (case when sensi with fast restart are ran from different threads concurrently) and improve performances. With DC sensi, creating the LF network from a bus/breaker view is more common than for the LF as we need it when we calculate sensi from an operator strategy state.
 - a network cache scope has been added to be able to run DC sensi analysese with fast restart from multiple thread on the same network and variants. The caller has to set a unique scope ID to avoid collision in the cache from multiple threads (which might as sensi with different operator strategies) 

- [ ] Yes
- [x] No

**Other information**:


src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -35,65 +37,284 @@
...
+            this.parameters = Objects.requireNonNull(parameters);
+        }
+
+        @Override
+        public LfInput copy() {
+            return new LfInput(OpenLoadFlowParameters.clone(parameters));
+        }
+
+        @Override
+        public Pair<Boolean, String> hasChanged(LfInput other) {
```
SylvestreSakti: Why a pair ? Either the string is null, or not. The boolean seems unnecessary.

src/main/java/com/powsybl/openloadflow/NetworkCache.java
```diff
@@ -35,65 +37,284 @@
...
+                    AcLoadFlowResult result = value.getContext().getResult();
+                    if (result != null && result.getSolverStatus() == AcSolverStatus.CONVERGED) {
+                        value.getContext().getParameters().setVoltageInitializer(new PreviousValueVoltageInitializer(true));
+                    }
+                }
+            }
+        }
+    }
+
+    public static class DcLfEntry extends AbstractEntry<DcSensiInput, DcSensiValue> {
```
SylvestreSakti: I would call this class DcSensiEntry (and use the name DcLfEntry for the DC load flow)

src/main/java/com/powsybl/openloadflow/NetworkCache.java
geofjamg: Indeed, boolean removed

src/main/java/com/powsybl/openloadflow/NetworkCache.java
geofjamg: Renamed

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
```diff
@@ -71,11 +73,63 @@ public void clean() {
...
+    }
+
     @FunctionalInterface
     public interface VariantCleanerFactory {
         VariantCleaner create(Network network, String workingVariantId, String tmpVariantId);
     }
 
+    public interface VariantProvider {
+
+        String getTmpVariantId(String workingVariantId);
```
SylvestreSakti, line 95: This is more than a getter, maybe something like `cloneToTmpVariant` is a better name ?

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
```diff
@@ -71,11 +73,63 @@ public void clean() {
...
+        private final Network network;
+
+        public VariantCloner(Network network) {
+            this.network = Objects.requireNonNull(network);
+        }
+
+        @Override
+        public String getTmpVariantId(String workingVariantId) {
+            String tmpVariantId = "olf-tmp-" + UUID.randomUUID();
+            network.getVariantManager().cloneVariant(workingVariantId, tmpVariantId);
```
SylvestreSakti, line 109: I am not very sure it is necessary to clone again the variant after acquiring it from the pool. Should it be done ? Or is there a use case (that can be added to tests) where the workingVariant is different than the one in the pool ?

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
geofjamg: it depends on the implementation, in the `VariantCloner` impl, the vraint is really cloned on the fly but in the `PoolVariantAcquirer` variants are pre-cloned and this method only get to an existing pre-cloned variant. This is why I used a neutral naming.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkList.java
geofjamg: In the `VariantCloner`, the variant is really cloned on the fly. There is nothing like pre-cloned variant pool acquisition

SylvestreSakti (approved):

PR #1406, opened 2026-05-12
Fix DC voltage initialisation

bperr:

Bug fix: In the previous version, the AC/DC load flow could fail for DC bipolar networks if the order of DC nodes connected to a converter (i.e. which DC node is declared as DC node 1 and which DC node is declared as DC node 2) was not choosen adequately. 

In the bipolar network below, DC bus 2 (in black) is considered as the second DC bus of converter 1 (see red numbers). Thus, its voltage is initialised at 0 pu. Similarly, DC bus 3 is initialised at 0 pu. This cause non-convergence of the load flow as two DC bus connected to the same converters cannot have the save voltage (this leads to division by zero in the converter equations)
image

DC voltage initialisation is properly handle to ensure DC buses connected to a convverter are initialised with different voltage values. DC buses which are not connected to any converter have no such constraints. Furthermore, the initial voltage value itself has no impact on the load flow result.

- [x] Yes
- [ ] No

The `LfDcBus` interface has been updated:
- Methods `setNeutralPole()` and `isNeutralPole()` have been removed.

The `VoltageInitializer` interface has been updated:
- Method `double getMagnitude(LfDcBus dcBus)` has been renamed to `double getDcVoltage(LfDcBus dcBus)` as *magnitude* refers to a positive value which may not be the case for DC buses

The interfaces `LfAcDcConverter` and `LfVoltageSourceConverter` have been updated
- The method `double getTargetVac()` has been moved from `LfAcDcConverter` to `LfVoltageSourceConverter` as this makes sens only in the case of VSC (not LCC)
- Method `getLossFactors()` now returns a record class `LossFactors` instead of a `List` as we explicitely know there is only three losses factors.

**Other information**:


src/main/java/com/powsybl/openloadflow/network/LfDcBus.java
```diff
@@ -16,9 +16,11 @@ public interface LfDcBus extends LfElement {
...
 
-    boolean isNeutralPole();
+    boolean isInitialVoltageSet();
 
-    void setNeutralPole(boolean isNeutralPole);
+    void setInitialVoltage(double v);
+
+    double getInitialVoltage();
 
     double getNominalV();
```
SGI-Landry, line 29: It would be great to document the interface, and it's probably not so difficult.

src/test/java/com/powsybl/openloadflow/acdc/AcDcBipolarTest.java
```diff
@@ -429,4 +434,52 @@ void testBipolarModelWithoutMetallicReturn() {
...
+            Arguments.of("C13", true, true, false, false),
+            Arguments.of("C14", true, true, false, true),
+            Arguments.of("C15", true, true, true, false),
+            Arguments.of("C16", true, true, true, true)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("bipolarFourConvAllVdcAndSwapCases")
+    void testDifferentVoltageInitialisation(String testCaseId, boolean swapOrder1, boolean swapOrder2, boolean swapOrder3, boolean swapOrder4) {
```
SGI-Landry, line 472: This tests precisely what we want to fix. Thanks!

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
```diff
@@ -36,10 +36,11 @@ public double getAngle(LfBus bus) {
     }
 
     public double getMagnitude(LfDcBus dcBus) {
```
SGI-Landry: The name of this method is inaccurate, because for DC buses voltage (relative) matter rather than magnitude (non-negative). What about taking this opportunity to change it?

src/main/java/com/powsybl/openloadflow/network/LfDcBus.java
```diff
@@ -16,9 +16,11 @@ public interface LfDcBus extends LfElement {
 
     void setV(double v);
 
-    boolean isNeutralPole();
+    boolean isInitialVoltageSet();
 
-    void setNeutralPole(boolean isNeutralPole);
+    void setInitialVoltage(double v);
+
+    double getInitialVoltage();
```
SGI-Landry: It sounds to me that _initial voltage_ is a property of the initialization rather than a property of the bus, i.e. data you will throw away as soon as initialization is done. This logic should likely be put in the initializer UniformValueVoltageInitializer.

It looks like what you really need is to keep track of what DC bus voltage is already set or not. For this you can probably use a HashSet in UniformValueVoltageInitializer::prepare. Then in the initializer you can probably just use LfDcBus::setV to set the initial voltage. If it does not work, you can connect the LfDcBus to its initial voltage value with a local map. This will avoid extending the LfDcBus interface.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfAcDcConverter.java
```diff
@@ -49,10 +49,23 @@ public abstract class AbstractLfAcDcConverter extends AbstractElement implements
...
+            dcBus1.setInitialVoltage(1);
+            dcBus2.setInitialVoltage(0);
+        } else if (dcBus1.isInitialVoltageSet() && !dcBus2.isInitialVoltageSet()) {
+            // Complement ensures the two DC buses get distinct values
+            dcBus2.setInitialVoltage(1 - dcBus1.getInitialVoltage());
+        } else if (!dcBus1.isInitialVoltageSet()) {
+            // Complement ensures the two DC buses get distinct values
+            dcBus1.setInitialVoltage(1 - dcBus2.getInitialVoltage());
+        }
+        // else: both already set by a previous converter, nothing to do
```
SGI-Landry: Logic and algorithms don't belong in constructors.
Doing this defeats the purpose of OOP for `VoltageInitializer`. This logic belongs in `UniformValueVoltageInitializer::prepare`.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfDcBus.java
```diff
@@ -21,7 +23,8 @@ public abstract class AbstractLfDcBus extends AbstractElement implements LfDcBus
 
     protected final double nominalV;
 
-    protected boolean isNeutralPole = false;
+    // Initial voltage used by the UniformValueVoltageInitializer. Only set for DC buses connected to a converter.
+    private Double initialVoltage = null;
```
SGI-Landry: Not needed (see other comments).

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfDcBus.java
```diff
@@ -50,12 +53,18 @@ public double getNominalV() {
...
+        Objects.requireNonNull(initialVoltage);
+        return initialVoltage;
     }
 
     @Override
-    public void setNeutralPole(boolean isNeutralPole) {
-        this.isNeutralPole = isNeutralPole;
+    public boolean isInitialVoltageSet() {
+        return initialVoltage != null;
     }
```
SGI-Landry: Not needed.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfDcBus.java
```diff
@@ -21,7 +23,8 @@ public abstract class AbstractLfDcBus extends AbstractElement implements LfDcBus
```
SGI-Landry: Can you please use this opportunity to document the unit of this protected attribute?

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfDcBus.java
```diff
@@ -12,6 +12,8 @@
 import com.powsybl.openloadflow.network.LfDcBus;
 import com.powsybl.openloadflow.network.LfNetwork;
 
+import java.util.Objects;
+
 /**
  * @author Denis Bonnand {@literal <denis.bonnand at supergrid-institute.com>}
  */
```
SGI-Landry, line 17: Not sure of the added value of this AbstractClass? Do we plan to add more DC bus implementations?

src/test/java/com/powsybl/openloadflow/network/AcDcNetworkFactory.java
```diff
@@ -2218,4 +2218,81 @@ public static Network createBipolarModelWithoutMetallicReturn() {
...
+     *  |    conv3 DC1----dl14----DC4----dl47----DC7 conv7   |
+     *  |    conv1-|                               |-conv5   |
+     *  AC1 -|     DC2----dl25----DC5----dl58----DC8     |- AC2
+     *  |    conv2-|               g               |-conv6   |
+     *  |    conv4 DC3----dl36----DC6----dl69----DC9 conv8   |
+     *  |                                                    |
+     *  |--------------l12-----------------------------------|
+     * </pre>
+     * @param id: Name of the network test case
+     * @param swapOrder1 : Whether converter conv1 DC nodes should be DC1 and DC2 or DC2 and DC1
```
SGI-Landry, line 2234: This is very clear, thanks for this accurate and relevant comment.

SGI-Landry (commented):
Hello @bperr,
Thanks for this activity.
- The test is exactly what we want.
- The initialization logic should be removed from LfDcBus and AbstractLfAcDcConverter and put in the initializer.
- You can use this opportunity to enhance existing comments and do a small refactoring.

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
bperr: Renamed to getDcVoltage

src/main/java/com/powsybl/openloadflow/network/LfDcBus.java
bperr: All logic has been moved to UniformValueVoltageInitializer prepare method, using a local map

bperr:
@SGI-Landry I added docstrings for DC components interface. Could you check this too ?
By the way I realised the method `getTargetVac()` in `LfAcDcConverter` belongs rather to `LfVoltageSourceConverter` as this is not valid for a LCC. I updated the interface and classes accordingly

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
```diff
@@ -11,18 +11,43 @@
...
     public void prepare(LfNetwork network, ReportNode reportNode) {
         Reports.reportVoltageInitializer(reportNode, NAME);
+
+        // We need to ensure that two LfDcBuses connected to the same converter are initialized with different voltage
+        // values. There is no such constraints for other LfDcBuses.
+        for (LfVoltageSourceConverter converter : network.getVoltageSourceConverters()) {
+            LfDcBus dcBus1 = converter.getDcBus1();
+            LfDcBus dcBus2 = converter.getDcBus2();
+
+            if (!dcBusInitialVoltage.containsKey(dcBus1) && !dcBusInitialVoltage.containsKey(dcBus2)) {
```
SGI-Landry, line 42: This is up to 6 lookups in `dcBusInitialVoltage` which is mildly inefficient. I don't think this is an issue here and at least this is very clear, all the more as lookup is O(1), but let's keep it in mind.

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
```diff
@@ -11,18 +11,43 @@
...
+                dcBusInitialVoltage.put(dcBus1, 1.0);
+                dcBusInitialVoltage.put(dcBus2, 0.0);
+            } else if (dcBusInitialVoltage.containsKey(dcBus1) && !dcBusInitialVoltage.containsKey(dcBus2)) {
+                // Complement ensures the two DC buses get distinct values
+                dcBusInitialVoltage.put(dcBus2, 1.0 - dcBusInitialVoltage.get(dcBus1));
+            } else if (!dcBusInitialVoltage.containsKey(dcBus1)) {
+                // Complement ensures the two DC buses get distinct values
+                dcBusInitialVoltage.put(dcBus1, 1.0 - dcBusInitialVoltage.get(dcBus2));
+            }
+            // else: both already set by a previous converter, nothing to do
```
SGI-Landry: There may be an edge case when both buses are set to the same value because of other converters. This happens in the contrived case where we have 3 VSC in a triangular configuration, and sharing their buses. Or if a VSC is shorted by mistake.

Still, I suggest that an exception is raised. This will at least help diagnose the contrived edge case if it happens.

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
```diff
@@ -35,11 +60,9 @@ public double getAngle(LfBus bus) {
...
     }
 
-    public double getMagnitude(LfDcBus dcBus) {
-        if (dcBus.isNeutralPole()) {
-            return 0.0;
-        }
-        return 1.0;
+    @Override
+    public double getDcVoltage(LfDcBus dcBus) {
+        return dcBusInitialVoltage.getOrDefault(dcBus, 1.0);
```
SGI-Landry, line 79: So the algorithm is
- initialize voltage of DC buses of all VSCs.
- all DC buses that are not connected to VSCs are initialized with V = 1 pu.
Can you please explicitly document it in the class Javadoc?

It looks like this behavior should also be tested, since the class is public and the behavior is selectable through parameters. What do you think?

src/main/java/com/powsybl/openloadflow/network/LfAcDcConverter.java
```diff
@@ -17,39 +17,102 @@
...
-
+    /**
+     * Get the three losses factors of the converter.
+     * 1. idle losses: losses independent of the DC current. In MW.
+     * 2. switching losses: losses proportional to the DC current. In MW/A.
+     * 3. resistive losses: losses proportional to the current squared. In Ohm.
+     *
+     * @return The three loss factor of the converter.
+     */
     List<Double> getLossFactors();
```
SGI-Landry: This is a poor man's 3-uple and a code smell because it relies on a brittle rule of size and order between `LfAcDcConverter` and `AbstractConverterDcCurrentEquationTerm`.  Can we please discuss this?

src/main/java/com/powsybl/openloadflow/network/LfAcDcConverter.java
```diff
@@ -17,39 +17,102 @@
...
 
+    /**
+     * @return The converter control mode.
+     */
     AcDcConverter.ControlMode getControlMode();
 
+    /**
+     * @return the active power on the AC side of the converter in per unit.
+     */
     double getPac();
```
SGI-Landry, line 89: It can be a right place where to remind sign convention.

src/main/java/com/powsybl/openloadflow/network/LfAcDcConverter.java
```diff
@@ -17,39 +17,102 @@
  */
 public interface LfAcDcConverter extends LfElement {
```
SGI-Landry, line 16: Thanks, the Javadoc really helps. What about reminding sign conventions?

By the way, I am wondering if we expect alternative loss models at some point.

src/main/java/com/powsybl/openloadflow/network/LfDcLine.java
```diff
@@ -14,21 +14,56 @@
...
 
+    /**
+     * @return The resistance of the DC line in Ohm.
+     */
     double getR();
 
+    /**
+     * @param i1 evaluable computing the current at side one of the DC line in per unit.
+     */
     void setI1(Evaluable i1);
```
SGI-Landry, line 35: Don't we rather need to specify the direction rather than the line end? i.e. entering/leaving the line in bus 1.

src/main/java/com/powsybl/openloadflow/network/LfDcLine.java
```diff
@@ -14,21 +14,56 @@
...
 
+    /**
+     * @param i2 evaluable computing the current at side two of the DC line in per unit.
+     */
     void setI2(Evaluable i2);
 
+    /**
+     * @param p1 evaluable computing the power at side one of the DC line in per unit.
+     */
     void setP1(Evaluable p1);
```
SGI-Landry, line 45: Same about specifying direction convention.

src/main/java/com/powsybl/openloadflow/network/LfVoltageSourceConverter.java
```diff
@@ -12,7 +12,18 @@
  */
 public interface LfVoltageSourceConverter extends LfAcDcConverter {
 
+    /**
+     * @return whether the voltage source converter controls AC voltage magnitude or reactive power.
+     */
     boolean isVoltageRegulatorOn();
```
SGI-Landry, line 18: Can you please specify what `true` and `false` respectively mean?

src/main/java/com/powsybl/openloadflow/network/LfVoltageSourceConverter.java
```diff
@@ -12,7 +12,18 @@
  */
 public interface LfVoltageSourceConverter extends LfAcDcConverter {
```
SGI-Landry, line 13: It would make sense to me to explain the sign convention in the interface Javadoc.

src/test/java/com/powsybl/openloadflow/acdc/AcDcBipolarTest.java
```diff
@@ -429,4 +434,52 @@ void testBipolarModelWithoutMetallicReturn() {
...
+        // The purpose of this test is to ensure the network definition, and especially which DC node is set as DC node1
+        // or DC node 2 of a converter has no impact on the load flow result (as long as the targetVdc are defined
+        // consistently with that order.
+
+        // Create network and run load flow
+        Network network = AcDcNetworkFactory.createFourConvertersBipole(testCaseId, swapOrder1, swapOrder2, swapOrder3, swapOrder4);
```
SGI-Landry: Can you please take the SonarQube warning into account?

SGI-Landry (changes requested):
This is much better, especially with the added documentation. Can you please add sign conventions?

Also there are a few code smells from the previous activity. It would be a good idea to record them and plan to fix them while it's still fresh.

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
bperr: A dedicated unit test has been added

src/main/java/com/powsybl/openloadflow/network/LfAcDcConverter.java
bperr: Replaced List by Triple

src/main/java/com/powsybl/openloadflow/network/LfDcLine.java
```diff
@@ -14,21 +14,56 @@
  */
 public interface LfDcLine extends LfElement {
 
+    /**
+     * @return The LfDcBus one side one of the DC line
```
SGI-Landry: Typo

```suggestion
     * @return The LfDcBus on side one of the DC line
```

src/main/java/com/powsybl/openloadflow/network/LfDcLine.java
```diff
@@ -14,21 +14,56 @@
...
+     */
     LfDcBus getDcBus1();
 
+    /**
+     * @return The LfDcBus one side two of the DC line
```
SGI-Landry: Typo

```suggestion
     * @return The LfDcBus on side two of the DC line
```

src/main/java/com/powsybl/openloadflow/network/LfDcLine.java
```diff
@@ -14,21 +14,56 @@
...
+     */
     void setI2(Evaluable i2);
 
+    /**
+     * @param p1 evaluable computing in per unit the power entering the DC line in DC bus 1.
+     */
     void setP1(Evaluable p1);
 
+    /**
+     * @param p2 evaluable computing in per unit the power entering the DC line in DC bus 1.
```
SGI-Landry: Typo

```suggestion
     * @param p2 evaluable computing in per unit the power entering the DC line in DC bus 2.
```

src/main/java/com/powsybl/openloadflow/network/LfDcLine.java
```diff
@@ -14,21 +14,56 @@
...
+     */
     void setP2(Evaluable p2);
 
+    /**
+     * Update the DC line state after the load flow.
```
SGI-Landry, line 53:
```suggestion
     * Update the DC line state.
```
Do you have a specific reason to assume this is after the load flow computation, or can we un-specify this?

src/main/java/com/powsybl/openloadflow/network/LfDcLine.java
```diff
@@ -14,21 +14,56 @@
...
+     *
+     * @param parameters   Parameters of state update.
+     * @param updateReport report of connected/disconnected branches and open/closed switch count.
+     */
     void updateState(LfNetworkStateUpdateParameters parameters, LfNetworkUpdateReport updateReport);
 
+    /**
+     * Update the DC line terminals current and power at the end of the load flow.
```
SGI-Landry, line 61:
```suggestion
     * Update the DC line terminals current and power.
```
Same as above.

src/main/java/com/powsybl/openloadflow/ac/equations/dcnetwork/AbstractConverterDcCurrentEquationTerm.java
```diff
@@ -39,7 +40,7 @@ public abstract class AbstractConverterDcCurrentEquationTerm extends AbstractEle
 
     protected final double resistiveLoss;
 
-    protected List<Double> lossFactors;
+    protected Triple<Double, Double, Double> lossFactors;
```
SGI-Landry: No need to use a member variable. This can be local to the constructor.

src/main/java/com/powsybl/openloadflow/ac/equations/dcnetwork/AbstractConverterDcCurrentEquationTerm.java
```diff
@@ -63,9 +64,9 @@ protected AbstractConverterDcCurrentEquationTerm(LfVoltageSourceConverter conver
         dcNominalV = nominalV;
         this.dcBus1 = dcBus1;
         this.dcBus2 = dcBus2;
-        this.idleLoss = lossFactors.get(0) / PerUnit.SB;
-        this.switchingLoss = lossFactors.get(1) * 1000d / nominalV;
-        this.resistiveLoss = lossFactors.get(2) / PerUnit.zb(nominalV);
+        this.idleLoss = lossFactors.getLeft() / PerUnit.SB;
```
SGI-Landry: idleLoss, switchingLoss and resistiveLoss are not used in the parent abstract class, so no code is factored. Would it make sense to put them in ConverterDcCurrentEquationTerm?

Additionnally this makes it possible to use other loss models later on.

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
```diff
@@ -7,22 +7,53 @@
...
+            } else if (dcBusInitialVoltage.containsKey(dcBus1) && !dcBusInitialVoltage.containsKey(dcBus2)) {
+                // Complement ensures the two DC buses get distinct values
+                dcBusInitialVoltage.put(dcBus2, 1.0 - dcBusInitialVoltage.get(dcBus1));
+            } else if (!dcBusInitialVoltage.containsKey(dcBus1)) {
+                // Complement ensures the two DC buses get distinct values
+                dcBusInitialVoltage.put(dcBus1, 1.0 - dcBusInitialVoltage.get(dcBus2));
+            } else {
+                // Both have been already set by a previous converter, nothing to do
+                if (Objects.equals(dcBusInitialVoltage.get(dcBus1), dcBusInitialVoltage.get(dcBus2))) {
+                    throw new PowsyblException("Could not initialize DC bus voltage properly");
```
SGI-Landry: Thanks for putting the check.

Can you please make the message a bit more informative?

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfAcDcConverter.java
```diff
@@ -49,12 +46,11 @@ public abstract class AbstractLfAcDcConverter extends AbstractElement implements
...
     protected AbstractLfAcDcConverter(AcDcConverter<?> converter, LfNetwork network, LfDcBus dcBus1, LfDcBus dcBus2, LfBus bus1) {
         super(network);
+
         this.dcBus1 = dcBus1;
         this.dcBus2 = dcBus2;
-        // By convention, the dcBus2 is supposed to be the neutral layer, it is just needed for voltage initialization
-        dcBus2.setNeutralPole(true);
         this.bus1 = bus1;
-        this.lossFactors = List.of(converter.getIdleLoss(), converter.getSwitchingLoss(), converter.getResistiveLoss());
+        this.lossFactors = Triple.of(converter.getIdleLoss(), converter.getSwitchingLoss(), converter.getResistiveLoss());
```
SGI-Landry: Still not fond of Triple (as opposed to a record) but maybe it's simpler.

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
```diff
@@ -7,22 +7,53 @@
...
  */
 public class UniformValueVoltageInitializer implements VoltageInitializer {
 
     public static final String NAME = "Uniform Values";
 
+    private final Map<LfDcBus, Double> dcBusInitialVoltage = new HashMap<>();
+
     @Override
     public void prepare(LfNetwork network, ReportNode reportNode) {
         Reports.reportVoltageInitializer(reportNode, NAME);
```
SGI-Landry, line 33: What about clearing dcBusInitialVoltage first ? This will ensure we always get to the same state after calling prepare().

src/main/java/com/powsybl/openloadflow/network/LfAcDcConverter.java
```diff
@@ -9,47 +9,116 @@
...
     double getQac();
 
+    /**
+     * @param qac the reactive power on the AC side of the converter in per unit.
+     *            Positive value means the power flows from AC network to DC network.
+     */
     void setQac(double qac);
 
+    /**
+     * Update the AC/DC converter state after the load flow.
```
SGI-Landry, line 110: why "after the load flow"?

src/main/java/com/powsybl/openloadflow/network/LfAcDcConverter.java
```diff
@@ -9,47 +9,116 @@
...
+     *
+     * @param parameters   Parameters of state update.
+     * @param updateReport report of connected/disconnected branches and open/closed switch count.
+     */
     void updateState(LfNetworkStateUpdateParameters parameters, LfNetworkUpdateReport updateReport);
 
+    /**
+     * Update the DC line terminals current and power at the end of the load flow.
```
SGI-Landry, line 118: Likewise, why at the end of the load flow computation?

SGI-Landry (changes requested):
Thanks, many enhancements already, especially for the place where initialization takes place and for the documentation of sign conventions and units. Still a few small points to fix.

src/main/java/com/powsybl/openloadflow/network/LfDcLine.java
bperr: The only usage of current implementation is performed after the load flow as the goal of this method is to update the attributes of the original `DcLine` (i.e. from iidm not a `LfDcLine`)

src/main/java/com/powsybl/openloadflow/ac/equations/dcnetwork/AbstractConverterDcCurrentEquationTerm.java
bperr: I don't see any problem moving them to the child class.

src/main/java/com/powsybl/openloadflow/network/impl/AbstractLfAcDcConverter.java
bperr: I defined the record inside the LfAcDcConverter interface

src/main/java/com/powsybl/openloadflow/ac/equations/dcnetwork/AbstractConverterDcCurrentEquationTerm.java
bperr: As the three loss factors are defind in the LfAcDcConverter interface, I do not think we can use other loss models

bperr:
@SylvestreSakti @jeandemanged This PR is mainly a bug fix. 
In some cases, the voltage of DC Bus was poorly initialised which caused non-convergence. This was especially true for bipolar links where converter had to be created in a specific way (the neutral node had to be the `dcBus2` for both converters). With this PR there is no more such constraints.

src/test/java/com/powsybl/openloadflow/ac/UniformValueVoltageInitializerTest.java
jeandemanged, line 1: header and author

jeandemanged (commented):
thank you, one minor comment

src/main/java/com/powsybl/openloadflow/ac/equations/dcnetwork/ConverterDcCurrentEquationTerm.java
```diff
@@ -20,8 +21,17 @@
  */
 public class ConverterDcCurrentEquationTerm extends AbstractConverterDcCurrentEquationTerm {
 
+    private final double idleLoss;
+
+    private final double switchingLoss;
+
+    private final double resistiveLoss;
+
```
SylvestreSakti, line 29: Just to know, what is the reason to move these variables from `AbstractConverterDcCurrentEquationTerm` to `ConverterDcCurrentEquationTerm` ? (is it due to a future development of LCC that would not use the same losses ?)

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
```diff
@@ -7,22 +7,55 @@
...
+                dcBusInitialVoltage.put(dcBus2, 0.0);
+            } else if (dcBusInitialVoltage.containsKey(dcBus1) && !dcBusInitialVoltage.containsKey(dcBus2)) {
+                // Complement ensures the two DC buses get distinct values
+                dcBusInitialVoltage.put(dcBus2, 1.0 - dcBusInitialVoltage.get(dcBus1));
+            } else if (!dcBusInitialVoltage.containsKey(dcBus1)) {
+                // Complement ensures the two DC buses get distinct values
+                dcBusInitialVoltage.put(dcBus1, 1.0 - dcBusInitialVoltage.get(dcBus2));
+            } else {
+                // Both have been already set by a previous converter, nothing to do
+                if (Objects.equals(dcBusInitialVoltage.get(dcBus1), dcBusInitialVoltage.get(dcBus2))) {
```
SylvestreSakti, line 53: Question about the algorith itself: Isn't this exception a real case that could happen ? Since only two values of initialization are possible (0 kV and 1 kV), what happens if a DC Bus is connected to more than 2 converters with already set values (that probably never happens... but asking the question in case)

Another algorithm idea that I suggest since any value of voltage is ok (only if we want to avoid that, but maybe it's not necessary, or maybe it's a bad idea 😄  ):


```        dcBusInitialVoltage.clear();
        for (LfDcBus dcBus : network.getDcBuses()) {
            dcBusInitialVoltage.put(dcBus, 1.0);
        }
        for (LfVoltageSourceConverter converter : network.getVoltageSourceConverters()) {
            LfDcBus dcBus1 = converter.getDcBus1();
            LfDcBus dcBus2 = converter.getDcBus2();
            double v1 = dcBusInitialVoltage.get(dcBus1);
            double v2 = dcBusInitialVoltage.get(dcBus2);
            // Applying operation that ensures the voltages are different
            dcBusInitialVoltage.put(dcBus1, 2 * v1 + 1);
            dcBusInitialVoltage.put(dcBus2, 2 * v2);
        }
```

SylvestreSakti (dismissed):
Thanks for all the javadoc methods that make it more clear ! A small algorithm suggestion, but your proposition may already be satisfying, so I approve this PR.

src/main/java/com/powsybl/openloadflow/ac/equations/dcnetwork/ConverterDcCurrentEquationTerm.java
bperr: No, it is simply that the abstract class did not use it.

src/main/java/com/powsybl/openloadflow/network/util/UniformValueVoltageInitializer.java
bperr: I don't foresee any realistic cases where a converter has DC buses with invalid set values. We can think of some odd-topologies that causes the algorithm to fail but I think we can consider this out of scope.

For reference purpose, I already implemented a test with a DC buses connected to more than 2 converters (`testDifferentVoltageInitialisation` in `AcDcBipolarTest.java`)

SGI-Landry: @SylvestreSakti the values are in pu. Each pole of a VSC should be connected either to positive, negative or neutral line. Conflicts are expected to occur where a line is connected to another one with different polarity. This looks odd enough to declare this exceptional. If this assumption is wrong, we can also use pseudo-random values.

SGI-Landry (dismissed):
Good for me.

SylvestreSakti (approved):

jeandemanged (approved):

PR #1414, opened 2026-05-24
Load Flow performance: remove redundant outer loop executions

jeandemanged:

No

Performance fix

We run more outer loops than strictly needed, for both AC and DC Load Flow.
Impact:
- confusing report
- most outer loop have a very fast check and this is not a real performance issue, however for incremental outer loops the performance hit can become noticeable

For example, with only DistributedSlack and ReactiveLimits outerloops enabled, and only the first DistributedSlack going unstable, todays we run:
1. DistributedSlack (unstable)
2. ReactiveLimits (stable)
3. DistributedSlack (stable)
4. ReactiveLimits (stable)

but because 2. ReactiveLimits outer loop did not change anything, what is strictly necessary is only:
1. DistributedSlack (unstable)
2. ReactiveLimits (stable)
(stop)

We ensure outer loop are executed only when strictly needed, for both AC and DC Load Flow.

- [x] No

**Other information**:


src/main/java/com/powsybl/openloadflow/ac/AcloadFlowEngine.java
```diff
@@ -258,9 +263,14 @@ public AcLoadFlowResult run() {
...
+                    if (outerLoop == runningContext.lastUnstableOuterLoop) {
+                        // We went through all other outer loops and none of them modified anything -> we are done.
+                        break;
+                    }
+                    runOuterLoop(outerLoop, outerLoopAndContext.getRight(), solver, runningContext, checkUnrealisticStates);
 
-                    if (outerLoopAndContext.getLeft() == runningContext.lastUnrealisticStateFixingLoop) {
+                    if (outerLoop == runningContext.lastUnrealisticStateFixingLoop) {
                         checkUnrealisticStates = true;
                     }
```
jeandemanged, line 282: wondering if broken by change....

jeandemanged: is fine

SylvestreSakti (commented):
If my understanding is correct, we are going from the current behavior:

Image


----
To the new behavior:

Image

Could the code be simplified to do a check only on the first outerloop ?

SylvestreSakti (approved):
Thanks !

PR #1415, opened 2026-05-27
Fix sensitivity miscalculation when using negative variable weights

SylvestreSakti:

No

Bug fix (detected by @phiedw ) concerning the calculation of sensitivity with multiple weighted variables (with some of the weights being negative). Negative weights can be used for example for generators that have a negative targetP.

When having multiple variables, the sensistivity is made by adding the sum of the *absolute value* of the weights, for example: If we have three variables of weights 50, 25, and -25, the scaling of the weights is done by summing 50 + 25 + 25.

If we have three variables of weights 50, 25, and -25, the scaling of the weights is done by summing 50 + 25 - 25.

- [ ] Yes
- [x] No


src/main/java/com/powsybl/openloadflow/sensi/AbstractSensitivityAnalysis.java
jeandemanged, line 557: what if weightSum is 0 ?

SylvestreSakti: :astonished: Did not think about this one, indeed it would throw an exception !

jeandemanged: on my side I never thought of a use case for negative weight...

SylvestreSakti: Putting this PR as draft for the moment, something needs to be discussed about the use cases of negative weights

PR #1417, opened 2026-05-28
Incremental Transformer Voltage Control: filter insensitive controllers

jeandemanged:

No

Feature

The incremental transformer voltage control outer loop modifies ratio tap changers positions by calculating the required voltage change at regulated bus and using the voltage sensitivity to ratio change. In some cases the sensitivity is zero, or almost zero due to numerical precision. Current behavior in such situation is that the tap changers goes at limit with no effective change in the regulated bus voltage. This causes many outer loop / NR executions and degrades performance, and also creates bad voltage result. In Security Analysis that epsilon sensitivity may also change sign for each contingency, causing the transformer to constantly change tap position upward/downward, also degrading performance.

Ratio tap changers with low sensitivity are excluded from regulation. A sensitivity threshold is currently hardcoded at 0.05, meaning that tap changers having a ratio change of 0.1 p.u. causing a voltage change of less than 0.005 p.u. are excluded from regulation. This threshold is currently hardcoded but could be made configurable later on - although this might not be necessary. E.g. assume a ratio tap with a ratio range of 0.4 p.u. (which is a really wide ratio range, that is 0.4 x 220 = 88 kV for a 220 kV transformer), this means filtering the transformer if is maximum influence on a 220 kV bus voltage is less that 0.4 x 0.05 x 220 = 4.4 kV.
Sensitivities are normally re-evaluated at each outerloop. Because calculating sensitivities has some performance cost, once a transformer is identified as being insensitive, it is excluded from further sensitivity calculation. This is however done only for consecutive outer loops. Indeed, PQ/PV switching may modify the set of insensitive controllers.

- [x] No

**Other information**:

Test on real network with 20k buses, distributed slack, reactive limits, transformers voltage regulation:
- Load Flow:
  - without this change: solves with 21 outer loops and 55 NR iterations
  - with this change: solves with 15 outer loops and 31 NR iterations
- Security Analysis (100 contingencies):
  - without this change: solves in 117 seconds, 788 outer loops, 2746 NR iterations
  - with this change: solves with 40 seconds, 287 outer loops, 532 NR iterations


src/main/java/com/powsybl/openloadflow/ac/outerloop/IncrementalTransformerVoltageControlOuterLoop.java
```diff
@@ -231,13 +250,19 @@ public OuterLoopResult check(AcOuterLoopContext context, ReportNode reportNode)
         AcLoadFlowContext loadFlowContext = context.getLoadFlowContext();
         var contextData = (IncrementalContextData) context.getData();
 
-        // filter out buses/branches which are outside their deadbands
+        // If another outerloop than us ran, this will reset insensitive statuses.
+        // Insensitive statuses they may have changed, in particular in case of PV/PQ switching.
```
SylvestreSakti:
```suggestion
        // If another outerloop than us ran, this will reset insensitive statuses
        // because insensitive statuses may have changed, in particular in case of PV/PQ switching.
```

SylvestreSakti (commented):
Ok with this PR 👍 Just a minor comment typo

SylvestreSakti (approved):

PR #1423, opened 2026-06-04
Support DC switch resistance in load flow

### LINKED ISSUE #1371: Support Enhanced DC circuit breaker

No resistance in DC switch.

Take into account in Open Load Flow the enhancements requested in https://github.com/powsybl/powsybl-core/issues/3827
The resistance of the DC circuit breaker is taken into account in the load flow computation.

### Describe the motivation

Resistance of breakers are essential to analyze losses as well as security in AC/DC load flow analyses.

### Extra Information

_No response_

SGI-Landry:

Fixes #1371 

- [ ] Yes
- [x] No


SGI-Landry:
@bperr can you please review this activity? CI fails but this is normal, as long as tests run locally.

docs/loadflow/loadflow.md
```diff
@@ -23,10 +23,10 @@ Open Load Flow also supports networks with HVDC lines (High Voltage Direct Curre
 
 ### DC detailed model
 
-Additionally, Open Load Flow supports AC-DC load flow formulation with detailed model of DC elements.  
-However, it is currently restricted to embedded DC components, meaning that all converters of a given DC component must be connected to the same synchronous component. 
+Additionally, Open Load Flow supports AC-DC load flow formulation with detailed model of DC elements.
```
bperr, line 26: Not a big deal, but why removing the line return ?

src/main/java/com/powsybl/openloadflow/network/impl/LfDcSwitchImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import com.powsybl.openloadflow.network.LfNetworkParameters;
+import com.powsybl.openloadflow.network.LfNetworkStateUpdateParameters;
+import com.powsybl.openloadflow.network.LfNetworkUpdateReport;
+
+import java.util.Objects;
+
+/**
+ * @author Landry Huet {@literal <landry.huet at supergrid-institute.com>}
+ */
+public class LfDcSwitchImpl extends AbstractLfDcLine {
```
bperr: Should `AbstractLfDcLine` be renamed to `AbstractLfDcBranch` ? It may now represent something else than a DC line

src/main/java/com/powsybl/openloadflow/network/impl/LfDcSwitchImpl.java
```diff
@@ -0,0 +1,66 @@
...
+        if (isDisabled()) {
+            updateFlows(Double.NaN, Double.NaN, Double.NaN, Double.NaN);
+        } else {
+            updateFlows(i1.eval(), i2.eval(), p1.eval(), p2.eval());
+        }
+    }
+
+    @Override
+    public void updateFlows(double i1, double i2, double p1, double p2) {
+        // TODO: DcSwitch has no DcTerminal yet in the IIDM model - revisit when terminals are added.
```
bperr, line 64: This mean that the result of the load flow cannot be propagated to the iidm `Network` object. Is there no workaround for this ?

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -611,11 +613,25 @@ private static void createBranches(List<LfBus> lfBuses, LfNetwork lfNetwork, LfT
 
     private static void createDcLines(LfNetwork lfNetwork, LoadingContext loadingContext, LfNetworkParameters parameters) {
```
bperr: Similarly, this may be renamed to `createDcBranches`

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -1250,6 +1271,7 @@ private LfNetwork createDc(int numCC, int numDcc, Network network, List<DcBus> d
 
         List<LfDcBus> lfDcBuses = new ArrayList<>();
         createDcBuses(dcBuses, parameters, lfNetwork, lfDcBuses, loadingContext);
+        network.getDcSwitches().forEach(loadingContext.dcSwitchSet::add);
```
bperr, line 1274: I understand that the `DcTopologyVisitor` interface does not allow to fill `dcSwitchSet` as for DC lines and ground so this is a workaround. 
However I think your current implementation is incorrect as it adds all DC switches of the network, rather than the ones of the DC component. If your network has two DC components, the switches will be added twice.

You can either add a filter here using `numDcc`, or group DC Switches by (numCC, numDCC) (as done for AC switches)

src/main/java/com/powsybl/openloadflow/network/impl/LfDcSwitchImpl.java
```diff
@@ -0,0 +1,66 @@
+/**
+ * Copyright (c) 2025, SuperGrid Institute (http://www.supergrid-institute.com)
```
bperr: 2026

src/test/java/com/powsybl/openloadflow/acdc/AcDcDcSwitchTest.java
```diff
@@ -0,0 +1,102 @@
+/**
+ * Copyright (c) 2025, SuperGrid Institute (http://www.supergrid-institute.com)
```
bperr: 2026

src/test/java/com/powsybl/openloadflow/network/AcDcNetworkFactory.java
```diff
@@ -2218,4 +2218,124 @@ public static Network createBipolarModelWithoutMetallicReturn() {
...
+        network.getDcLine("dl34").remove();
+        network.newDcSwitch()
+                .setId("sw34")
+                .setKind(DcSwitchKind.BREAKER)
+                .setDcNode1("dn3")
+                .setDcNode2("dn4")
+                .setOpen(false)
+                .setR(rSwitch)
+                .add();
+        addStandardConverters(network);
```
bperr, line 2245: Since #1406 you have helper functions in `AbstractLoadFlowNetworkFactory.java ` to create elements faster.

bperr (changes requested):

docs/loadflow/loadflow.md
SGI-Landry: It does not show this way locally. I suspect this has something to do with a tool in the chain (editor, merge resolution, github commit hooks) removing trailing whitespace. I don't see the value in the line break in this case. Otherwise I would switch to a double linefeed to enforce line break in markdown.

src/main/java/com/powsybl/openloadflow/network/impl/LfDcSwitchImpl.java
SGI-Landry: It is a good idea. It has consequences in many files, however. In many places we then need to rename 'DcLine' to 'DcBranch'.

src/main/java/com/powsybl/openloadflow/network/impl/LfDcSwitchImpl.java
SGI-Landry: You can still see the voltages at each node where the DcSwitches are connected. You do lose the data on the current. Fixing this requires however another activity in PowSyBl Core and it raises more questions. I will create a GitHub issue for this.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
SGI-Landry: Like many other methods and variables.

src/test/java/com/powsybl/openloadflow/network/AcDcNetworkFactory.java
SGI-Landry: Let's discuss this. I don't see how this is less verbose.

jeandemanged:
Hi@SGI-Landry , cc @SylvestreSakti 
Could you please update your branch ? (for inclusion in next release)
we do have powsybl-core RC now, and final powsybl-core should be ready soon

SGI-Landry:
> Hi@SGI-Landry , cc @SylvestreSakti Could you please update your branch ? (for inclusion in next release) we do have powsybl-core RC now, and final powsybl-core should be ready soon

Hello @jeandemanged I have just updated the branch. I have discovered a bug in the Core part of this activity, with a consequence in the load flow computation (see https://github.com/powsybl/powsybl-core/issues/3953).

**Please let's not merge the activity in the release train, but rather wait until the issue in Core is fixed.**

jeandemanged:
Hello @SGI-Landry ,
DC switches were made in IIDM without Terminals just like AC switches.
Not sure https://github.com/powsybl/powsybl-core/issues/3953 is really a bug, rather a feature.
The needed feature is the ability to store DC flow results on DC switches. Modifying DC Switches to be DC Connectable and having DC Terminals could be one solution - but involves more work.
Some alternatives by decreasing complexity:
- an IIDM Extension of DC Switch to store these results. I think the https://github.com/powsybl/powsybl-core/issues/3953 proposal has an issue: DC Switches without resistance will never have results, and yet the DC Terminals are there. With extension you would put an extension only where results are relevant solving that problem.
- putting results as simple IIDM properties on the DC Switches with non zero resistance. This mechanism is already used e.g. in LfBoundaryLineBus to store solved voltage magnitude and angle as properties.
- not exporting the results for DC Switches in IIDM

My call would be for one of the above options (not writing, or writing as properties). If this cannot be done for this release train, I think it would make sense to do a small PR to have OLF throwing an exception on networks with DC Switch with non-zero resistance.

Let us know

SGI-Landry:
> Hello @SGI-Landry , DC switches were made in IIDM without Terminals just like AC switches. Not sure [powsybl/powsybl-core#3953](https://github.com/powsybl/powsybl-core/issues/3953) is really a bug, rather a feature. The needed feature is the ability to store DC flow results on DC switches. Modifying DC Switches to be DC Connectable and having DC Terminals could be one solution - but involves more work. Some alternatives by decreasing complexity:
> 
>     * an IIDM Extension of DC Switch to store these results. I think the [DcTerminals in DcSwitches powsybl-core#3953](https://github.com/powsybl/powsybl-core/issues/3953) proposal has an issue: DC Switches without resistance will never have results, and yet the DC Terminals are there. With extension you would put an extension only where results are relevant solving that problem.
> 
>     * putting results as simple IIDM properties on the DC Switches with non zero resistance. This mechanism is already used e.g. in LfBoundaryLineBus to store solved voltage magnitude and angle as properties.
> 
>     * not exporting the results for DC Switches in IIDM
> 
> 
> My call would be for one of the above options (not writing, or writing as properties). If this cannot be done for this release train, I think it would make sense to do a small PR to have OLF throwing an exception on networks with DC Switch with non-zero resistance.
> 
> Let us know

Thanks for your answer, and thanks @SylvestreSakti for putting the exception.

The issue is more than just recording the results. Continuing the discussion on https://github.com/powsybl/powsybl-core/issues/3953

PR #1438, opened 2026-06-23
Support BRANCH_RESISTANCE, BRANCH_REACTANCE and BRANCH_ADMITTANCE variable types in AC Sensitivity Analysis

gautierbureau:

It introduces the sensitivity of a function with respect to R, X or Y of a branch.

No such sensitivity exist.

Sensitivity against R, X and Y exist for supported functions

- [ ] Yes
- [x] No


src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
```diff
@@ -100,6 +101,89 @@ private void calculateSensitivityValues(List<LfSensitivityFactor<AcVariableType,
...
+     * depends explicitly (not only through the state) on the variable, i.e. for self-sensitivity where the
+     * variable element and the function element are the same. Today only branch parameters (R / X / Y) contribute;
+     * every other variable type returns 0.
+     */
+    private static double computeParameterDirectPartial(LfSensitivityFactor<AcVariableType, AcEquationType> factor) {
+        if (!(factor instanceof SingleVariableLfSensitivityFactor<AcVariableType, AcEquationType> singleFactor)) {
+            return 0;
+        }
+        SensitivityVariableType varType = factor.getVariableType();
+        if (varType != SensitivityVariableType.BRANCH_RESISTANCE && varType != SensitivityVariableType.BRANCH_REACTANCE
```
SylvestreSakti, line 116: I suggest to put this check first in this method, to stop the computation as early as possible in nominal use cases that do not use this feature, to have less impact on the performance (I suppose it is negligible, but just in case...)

Maybe even outside the method ?

SylvestreSakti (commented):
Small request, but maybe it is not necessary.

Otherwise I am ok with this PR ! (Maybe using assertEquals instead of assertTrue would be cleaner for Sonar)

src/main/java/com/powsybl/openloadflow/sensi/AcSensitivityAnalysis.java
gautierbureau: I think the two tests are pretty much harmless in computation time and I think it is clearer to left it there and not directly here : sensi += computeParameterDirectPartial(factor);

SylvestreSakti (approved):
Some refactoring may be needed to put the direct partials code outside the Sensitivity classes (probably directly in the ClosedBranch terms)

geofjamg (approved):

PR #1457, opened 2026-07-07
Fix handling of pure phase-shift taps in DC sensitivity analysis

geofjamg:

No

Bug fix

When a phase shift action does not change the impedance, it should not produce a woodburry element. Otherwise we might get a nan in the alpha solve.

Pure phase shift without any impedance change are not producing a woodburry element anymore.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/dc/fastdc/ComputedElement.java
+++ b/src/main/java/com/powsybl/openloadflow/dc/fastdc/ComputedElement.java
@@ -24,6 +24,8 @@ import com.powsybl.openloadflow.equations.EquationSystem;
 import com.powsybl.openloadflow.graph.GraphConnectivity;
 import com.powsybl.openloadflow.network.LfBranch;
 import com.powsybl.openloadflow.network.LfBus;
+import com.powsybl.openloadflow.network.PiModel;
+import com.powsybl.openloadflow.network.TapPositionChange;
 import com.powsybl.openloadflow.network.action.AbstractLfBranchAction;
 import com.powsybl.openloadflow.network.action.AbstractLfTapChangerAction;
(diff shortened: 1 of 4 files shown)


src/main/java/com/powsybl/openloadflow/dc/fastdc/ComputedElement.java
```diff
@@ -168,4 +177,18 @@ private static Stream<Map.Entry<LfAction, List<ComputedElement>>> computeActionE
...
         return Stream.of(Map.entry(lfAction, elements));
     }
+
+    /**
+     * Returns true if the tap position change modifies the branch power, i.e. its impedance or ratio. In DC the branch
+     * power only depends on the resistance, the reactance and the first side ratio (see
+     * {@link com.powsybl.openloadflow.dc.equations.AbstractClosedBranchDcFlowEquationTerm#computePower}). A pure phase
+     * shift leaves the power unchanged and so does not require a Woodbury element.
+     */
+    private static boolean changesBranchPower(TapPositionChange change) {
```
SylvestreSakti: This probably depends on the **DcApproximationType** (e.g. if `IGNORE_R`, the difference in R should be ignored). Could we return instead :
`return AbstractClosedBranchDcFlowEquationTerm.computePower(true, dcApproximationType, oldPiModel) !=  AbstractClosedBranchDcFlowEquationTerm.computePower(true, dcApproximationType, newPiModel)`

But some work may have to be done to get access to the **dcApproximationType** (which is only present in the `DcEquationSystemCreationParameters`)

geofjamg: You are right. 
Fixed

SylvestreSakti (approved):

PR #1458, opened 2026-07-07
Incremental connectivity with DTree algorithm

### LINKED ISSUE #912: Incremental connectivity

In case of security analysis with remedial actions, we use a naive graph connectivity analysis. It is very slow!

A analysis of state-of-art incremental algorithms must be done. Depending on the results of the analysis, a new implementation must be developed to speed computations.

### Describe the motivation

_No response_

### Extra Information

_No response_

PoulpoGaz:

Fixes #912 

Feature

Security analysis with operator strategies uses NaiveGraphConnectivity which is slow.

This PR adds a new GraphConnectivity implementation: DTreeGraphConnectivity using the algorithm by [Chen et al, 2022](https://arxiv.org/pdf/2207.06887)  It has the following mathematical properties:
- addVertex: $\mathcal{O}(1)$,
- addEdge: $\mathcal{O}(h)$ where $h$ is the average depth. DTree maintains a set of trees and actively tries to reduce the average depth.
- removeEdge: $\mathcal{O}(h)$,
- getComponentNumber/getConnectedComponent : $\mathcal{O}(h)$.

In terms of performance, the total time required for connectivity is: 
- For a security analysis with ~5000 contingencies:
  - DTree: 0.23 s 
  - EvenShiloach: 0.46 s
  - Naive: 15.7 s
- ~5000 contingencies and, for each, one operator strategy with one action (close switch):
  - DTree: 8.71 s. It is possible to improve this time to 0.64 s, but this will be done in another PR.
  - Naive: 77.5 s

- [ ] Yes
- [x] No

**Other information**:

Algorithm description: https://arxiv.org/pdf/2207.06887
Authors' python implementation: https://github.com/qingchen3/D-tree/
Experimental study of existing algorithms: https://arxiv.org/pdf/2501.02278

DTree is an algorithm trying to minimizes something called $S_d$ or sum of distances. That's the sum of depth of every nodes. Because $S_d$ is linked to the complexity of all operations, the smaller it is, the faster DTree is. But, finding the minimal value takes too much time, so heuristics are used. Some heuristics were changed:
- Instead of searching the best replacement edge, the first one encountered is selected. It increases the sum of distances but improves performance of removeEdge.
- When inserting a non-tree edge whose endpoints have a depth difference $\Delta \geq 2$, the ancestor at distance $\Delta / 2 − 1$ from the deeper endpoint is disconnected from its parent, rather than disconnecting the ancestor at distance $\Delta − 2$ as in the original algorithm. It improves the sum of distances and performances.
- In queries, the centroid heuristic isn't used. It doesn't decrease the sum of distances that much and deteriorate performances.

The last two changes were proposed by [Xu et al, 2024](https://dl.acm.org/doi/abs/10.1145/3698805)

The following graph compares the evolution of sum of distances after each call to a method of GraphConnectivity (operation) between 4 implementations of DTree. This was done during a security analysis on a network with ~7000 buses, ~5500 contingencies and ~5500 operator strategies.
- `DTreeGraphConnectivity` selects the first replacement edge in `removeEdge` and uses $\Delta/2 - 1$ in `insertNonTeeEdge`. This is the proposed implementation in this PR.
- `Delta2DTreeGraphConnectivity` uses $\Delta - 2$ in `insertNonTreeEDge`
- `ReplaceWithBestDTreeGraphConnectivity` selects the best replacement edge
- `Delta2ReplaceWithBestDTreeGraphConnectivity` selects the best replacement edge and uses $\Delta - 2$ as in the algorithm's description.
DTree_vs_Delta2_vs_ReplaceWithBest_vs_Delta2ReplaceWithBest
Our implementation performs very well and is competitive compared to `ReplaceWithBestGraphConnectivity` which uses all heuristics to minimize the sum of distances.  In terms of performances, `ReplaceWithBestGraphConnectivity` is 3.6x-8.9x slower than `DTreeGraphConnectivity`. `Delta2ReplaceWithBest` is the slowest, 61x slower on big network (65000 buses). `Delta2DTreeGraphConnectivity` is the 2nd fastest, 2x slower than `DTreeGraphConnectivity`. These results justify the first two changes.

The following graph compare two DTree implementations: `DTreeGraphConnectivity` reroot tree during queries and `DTreeNoOptRerootGraphConnectivity` never reroot trees. They both have similar results. 
DtreeWithFindOptReroot_versus_Without
`DTreeGraphConnectivity` is also a bit faster. These results justify the last change.


src/main/java/com/powsybl/openloadflow/graph/dtree/AbstractSetView.java
```diff
@@ -0,0 +1,56 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.openloadflow.graph;
+
+import java.util.AbstractSet;
+import java.util.Collection;
+import java.util.Iterator;
+
+/**
+ * @author Valentin Carrez {@literal <valentin.carrez at rte-france.com>}
```
flo-dup, line 22: Could you describe here what you call an AbstractSetView and why you need it

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+ */
+public class DTreeGraphConnectivity<V, E> extends AbstractGraphConnectivity<V, E, DTreeGraphConnectivity.DTGraph<V, E>> {
+
+    public DTreeGraphConnectivity() {
+        super(new DTGraph<>());
+    }
+
+    @Override
+    protected void updateConnectivity(EdgeRemove<V, E> edgeRemove) {
+        componentSets = null;
```
flo-dup: Could you add a tiny comment here (and repeat it in the below methods) to explain why the update isn't done here

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+    protected void updateComponents() {
+        if (componentSets != null) {
+            return;
+        }
+
+        DTGraph<V, E> graph = getGraph();
+        List<DTGraph<V, E>.DTNode> roots = graph.roots;
+
+        // sorting roots will sort components as components is a wrapper around roots
+        roots.sort((s1, s2) -> s2.size - s1.size);
```
flo-dup: Use `Comparator.comparingInt(root -> root.size).reversed()` for clarity

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+        for (int i = 0; i < graph.roots.size(); i++) {
+            roots.get(i).rootIndex = i;
```
flo-dup: Is the gain worth the maintainability cost? graph.roots is most of the time a tiny list, so indexOf might be enough and avoids the risks of non synchronized information.

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+
+    @Override
+    protected Set<V> getVerticesNotInMainComponent(V mainComponentVertex) {
+        // first determine the excluded tree: either the tree containing
+        // the mainComponentVertex, either the biggest tree
+        DTGraph<V, E>.DTNode excludedTree = getMainComponentRoot(mainComponentVertex);
+        return new VerticesNotInMainComponent(excludedTree);
+    }
+
+    private final class VerticesNotInMainComponent extends AbstractSetView<V> {
```
flo-dup: There are several inner final static classes (VerticesNotInMainComponent, DTGraph, DTNode, ...), which are big enough to be in separate files, this would better separate concerns and improve readability of DTree class. Could you put those DTree utility classes in a specific package?

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+                return getGraph().rootOf((V) o) != excludedTree.findRoot();
+            }
+
+            return false;
+        }
+
+        @Override
+        public int size() {
+            if (size < 0) {
+                size = 0;
```
flo-dup: using streams might slightly improve readability?
```java
        size = getGraph().roots.stream()
                .filter(root -> root != excludedTreeRoot)
                .mapToInt(root -> root.size)
                .sum();
```

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+                return true;
+            }
+
+            DTGraph<V, E> graph = getGraph();
+            while (index < graph.roots.size()) {
+                DTGraph<V, E>.DTNode next = graph.roots.get(index);
+                index++;
+
+                if (next != excludedTree) {
+                    curIt = getGraph().iterator(next);
```
flo-dup: the root node itself has enough information to create the iterator, so it's a bit strange to call the graph to create it. Maybe directly create the iterator here?

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+            return biggestRoot;
+        }
+    }
+
+    @Override
+    public boolean supportTemporaryChangesNesting() {
+        return true;
+    }
+
+    public long computeSd() {
```
flo-dup: This method is only used in tests, I think you can remove it

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+        return getGraph().sumOfDistances();
+    }
+
+    public static final class DTGraph<V, E> implements GraphModel<V, E> {
+
+        // map a vertex to a node in a spanning tree
```
flo-dup: Use standard javadoc comment 
```suggestion
        /** map a vertex to a node in a spanning tree */
```

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+         * @param vertex the vertex whose tree root is to be returned.
+         * @return the root of the tree in which {@code vertex} is.
+         */
+        DTNode rootOf(V vertex) {
+            return vertexToTreeNode.get(vertex).findRootOptReroot();
+        }
+
+        // =============
+        // * INSERTION *
+        // =============
```
flo-dup: Could you remove these kind of comments? We don't use such comments in powsybl, and they don't bring much (although I must admit it helps me going through the review!). It's good that you sorted the methods in a logical way, but there's no need to add those

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+            if (node == null) {
+                throw new IllegalArgumentException("no such vertex in graph: " + v);
+            }
+
+            return node;
+        }
+
+        /**
+         * Insert a non tree edge between {@code nodeU} (whose depth is {@code depthU})
+         * and {@code nodeV} (whose depth is {@code depthV}). The two node must be in the
```
flo-dup:
```suggestion
         * and {@code nodeV} (whose depth is {@code depthV}). The two nodes must be in the
```

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+         *
+         * @param root the root of the tree in which an edge is to be added.
+         * @param nodeU one endpoint of the edge to add.
+         * @param depthU the depth of {@code nodeU}.
+         * @param nodeV the other endpoint of the edge to add.
+         * @param depthV the depth of {@code nodeU}
+         * @param edge edge linking {@code nodeU} and {@code nodeV}
+         * @return {@code true} if the edge inserted is a tree edge. This is true if |depthU - depthV| >= 2.
+         */
+        private boolean insertNonTreeEdge(DTNode root, DTNode nodeU, int depthU, DTNode nodeV, int depthV, Edge edge) {
```
flo-dup: You should rename this method or split it into two parts, as it's sometimes inserting a tree edge...

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+                DTNode node = this;
+
+                while (node.parent != null) {
+                    node = node.parent;
+                }
+
+                return node;
+            }
+
+            private Pair<DTNode, Integer> findRootWithDepth() {
```
flo-dup: Rather return a inner Record than a Pair, it would be much more understandable than getValue / getKey

flo-dup (changes requested):

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
```diff
@@ -0,0 +1,1063 @@
...
+
+                return next.vertex;
+            }
+        }
+
+        private final class Edge {
+            private final DTNode nodeU;
+            private final DTNode nodeV;
+            private final E edgeData;
+            private boolean treeEdge;
```
flo-dup: isn't this information already available easily in the nte sets of both corresponding nodes? removing it should have little impact on performances, it's the contains cost which is called each time an edge is removed (not many calls then) instead of maintaining a boolean

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
PoulpoGaz: renamed to insertEdgeInComponent

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
PoulpoGaz: I realized there is one more way to determine if an edge is a tree edge without maintaining a boolean: using the parent/child relationship. An Edge `edge` is a tree edge if one of the endpoint is the parent of the other one and `edge` is used to link them. The second part is to take into account multi-edges. This can be implemented like this: 
```java
nodeU.getParent() == nodeV && edge == nodeU.getParentEdge() 
|| nodeV.getParent() == nodeU && edge == nodeV.getParentEdge()
```
where `nodeU` and `nodeV` are the two endpoints.

I benchmarked the three versions by first running a security analysis and writing every call to GraphConnectivity in a file. Then, I loaded the file and used JMH to measure the average time needed to run all calls. This excludes the impact of OLF, but measuring precisely the time dedicated to connectivity and doing a security analysis is difficult. For all three versions, I have similar results across different inputs.

**Details**:
I have three versions of DTree:
- With boolean: version maintaining a treeEdge boolean
- With nte: version using the set of nte to distinguish between nte and tree edge as you suggested
- Check parent: version that uses the parent/child relationship

I have three inputs:
- 5541-1: security analysis on a network with ~7000 nodes on which 5541 contingencies each affecting one Line were simulated
- 5541-1/1: same as before but for each contingency there is an OperatorStrategy with one action (close switch)
- 10000-10/10: SA on a bigger network ~65000 nodes, with 10 000 contingencies each affecting 10 Lines and for each one OperatorStrategy with 10 actions (TerminalsConnectionAction)

I have the following results, all in "s/op".
| Parameter \ DTree | With boolean | With nte | Check parent |
|--------|--------|------|--------|
| 5541-1           | 0.009 ± 0,001 | 0.009 ± 0,001 | 0.009 ± 0,001 |
| 5541-1/1       | 1,725 ±  0,041 | 1,730 ±  0,063 | 1,720 ±  0,069 |
| 10000-10/10 | 0,266 ±  0,030 | 0,270 ±  0,006 | 0,263 ±  0,012 |

Overall, all DTree have similar results across all inputs. Using the nte set doesn't introduce significant overhead and is considerably simpler than the two other implementations. I think it is the right solution.

src/main/java/com/powsybl/openloadflow/graph/DTreeGraphConnectivity.java
PoulpoGaz: Instead of using an `ArrayList` to store roots, I can use a `LinkedHashSet`. This way, removing and adding a root can be fast but is still a bit slower than maintaining `rootIndex`. With this change, `rootIndex` serves only one purpose: getting the index of a component. So it only needs to be updated in `getComponentNumber`, which is done in `updateComponents`. So with this approach, there is no risk of unsynchronized information I think.

Nevertheless, I need to do one more thing in `updateComponents`, mapping DTNode to Component and storing them in `componentSets`. This can be costly when there are a lot of different connected components. However, sorting components and revalidating `rootIndex` are also costly.

Also, I don't use a `HashSet` because iterating over roots is slower. `HashSet` uses an array to store elements, but it can only grow. After all vertices are added, the length of the array is at least n (number of vertices) and will never be shrunk. Each time I want to iterate over roots, it will internally iterate over n elements instead of iterating only over the elements it will return, leading to a huge performance loss.

When benchmarking, i have the following results (op/s):

| Contingencies-Lines/OperatorStrategy-Actions | DTree ArrayList | DTree LinkedHashSet |
|--------|--------|--------|
| 5541-1 | 0,011 ± 0,001 | 0,011 ± 0,001 |
| 5541-1/1 | 1,985 ± 0,077 | 2,360 ± 0,038 |
| 10000-10/10-10 | 0,367 ± 0,008 | 0,407 ± 0,187 |

For 5541-1/1, there are 2011 roots! When using a `Set`, I need to copy them into `componentSets`, copy avoided when using an `ArrayList`. Hence the worse performances when using a Set.

PoulpoGaz: Accepted as the loss of performance is acceptable. However, I didn't remove methods manipulating roots (removeRoot, addRoot, replaceRoot) to simplify potential changes to the linked set of roots.

PR #1471, opened 2026-07-24
Support disconnection of DC elements in the load flow

bperr:

No

Feature

The AC-DC load flow did not consider wheher DC terminals of elements were connected or disconnected.

Disconnnected elements are properly taken into account during load flow

- [ ] Yes
- [x] No

**Other information**:


docs/loadflow/loadflow.md
```diff
@@ -377,6 +378,8 @@ The voltage source converter is the link between AC and DC networks, it is linke
 DC buses at the other side.<br>
 Please note that converters with a second optional AC terminal are not supported by Open Load Flow.
 
+If a terminal of the converter is disconnected, the converter is not included in the load flow, and none of the equations below is added.
```
SGI-Landry, line 381: This will probably require a bit more details, e.g. whether an equation imposing I=0 is added, or what happens with losses and on the AC side.

src/main/java/com/powsybl/openloadflow/ac/equations/dcnetwork/OpenDcLineEquationTerm.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public double der(Variable<AcVariableType> variable) {
+        Objects.requireNonNull(variable);
+        throw new IllegalStateException("Unknown variable: " + variable);
+    }
+
+    @Override
+    public String getName() {
+        return "dc_open";
```
SGI-Landry:
```suggestion
        return "dc_open_line";
```
?

src/test/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImplTest.java
```diff
@@ -331,4 +333,56 @@ void testDiscardGeneratorsWithTargetPOutsideActiveLimitsFromVoltageControl() {
         assertEquals(LfGenerator.GeneratorControlType.OFF, generators.get(1).getGeneratorControlType());
         assertEquals(LfGenerator.GeneratorControlType.VOLTAGE, generators.get(2).getGeneratorControlType());
     }
+
+    @Test
+    void testDcBusWithDisconnectedDcGroundIsNotGrounded() {
```
SGI-Landry, line 338: I am a bit puzzled by these tests. What behavior are we testing here? Are you testing the interface or the implementation details? This is not clear to me, can we please discuss it?

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+
+    @BeforeEach
+    void setUp() {
+        loadFlowRunner = new LoadFlow.Runner(new OpenLoadFlowProvider(commonTestConfig.matrixFactory()));
+        parameters = new LoadFlowParameters();
+        OpenLoadFlowParameters.create(parameters).setAcDcNetwork(true);
+    }
+
+    /**
+     * Verify the DC node voltage is coherent with the nominal voltage
```
SGI-Landry:
```suggestion
     * Verify the DC node voltage is consistent with the nominal voltage
```
But is this really needed? I would expect that you should know what values to expect precisely from your test cases, rather than test that the value is finite and "small enough".

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.List;
+import java.util.stream.Stream;
+
+import static java.lang.Math.abs;
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * @author Baptiste Perreyon {@literal <bapstiste.perreyon at supergrid-institute.com>}
```
SGI-Landry:
```suggestion
 * @author Baptiste Perreyon {@literal <baptiste.perreyon at supergrid-institute.com>}
```

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+     * @param network                       The network with load flow result
+     * @param disconnectedDcLineId          id of the disconnected DC line. Its current should be zero (partially disconnected)
+     *                                      or NaN (if fully disconnected)
+     * @param expectedCurrent               Expected current in the disconnected DC line. Should be zero or NaN (see above).
+     * @param otherLineIdsWithZeroCurrent   Ids of other DC lines which current should be zero.
+     * @param convertersIdWithZeroDcCurrent Ids of converter which DC current should be zero. Their absorbed AC power
+     *                                      should correspond to their idle loss.
+     */
+    void checkResult(Network network, String disconnectedDcLineId, double expectedCurrent, List<String> otherLineIdsWithZeroCurrent, List<String> convertersIdWithZeroDcCurrent) {
+        // We must use a 1e-3 tolerance because Java considers 0 and -0 different values otherwise
```
SGI-Landry: After checking, it looks like in IEEE 754, -0. == 0., but it does not work for boxed values in Double (with Double.equals). Here I only see unboxed doubles so I don't understand the issue. If it's really only a matter of comparing -0 and +0, then you can use a far stricter tolerance, e.g. Double.MIN_VALUE or Double.MIN_NORMAL.

But I would rather check for other issues in the code first.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+        // Verify converters have no DC current and AC power simply matches idle losses
+        for (String converterId : convertersIdWithZeroDcCurrent) {
+            VoltageSourceConverter converter = network.getVoltageSourceConverter(converterId);
+            assertEquals(0., converter.getDcTerminal1().getI(), 1e-3);
+            assertEquals(0., converter.getDcTerminal2().getI(), 1e-3);
+            assertEquals(converter.getIdleLoss(), converter.getTerminal1().getP(), 1e-3);
+        }
+
+        // Verify DC nodes voltage is realistic
+        network.getDcNodes().forEach(this::assertVoltageRealistic);
```
SGI-Landry: This routine can be a convenient shorthand to check zero values, but I don't understand the point of `assertVoltageRealistic`. If you want to compare results, it makes more sense to check against analytical results, or at least reference values from a run.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+    @ParameterizedTest
+    @MethodSource("sidesToDisconnectAndExpectedCurrent")
+    void testDcLineDisconnectionAsymmetricalMonopole(List<TwoSides> sidesToDisconnect, double expectedCurrent) {
+        /// Removing the DC line creates two separated DC components. It is expected that :
+        /// - The DC current in the line is zero
+        /// - The only power that goes into VSC is to compensate their idle losses
+        Network network = AcDcNetworkFactory.createAcDcNetwork1();
+        network.getDcLine("dl34").getDcTerminals().stream()
+            .filter(dcTerminal -> sidesToDisconnect.contains(dcTerminal.getSide()))
+            .forEach(DcTerminal::disconnect);
```
SGI-Landry:
```suggestion
        network.getDcLine("dl34").getDcTerminal1().disconnect();
```
You lose most of the benefit of your parametrized test with this line. If you must, a switch will be simpler. But is the parametrized test necessary in the first place? Let me suggest
1. Check on a very simple case that the disconnection has the same effect whether it is done on one or two terminals of the line.
2. Stick to a single case for your other test cases. You hardly lose anything this way.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+        network.getVoltageSourceConverter("conv23p").setTargetVdc(200).setControlMode(AcDcConverter.ControlMode.V_DC);
+
+        // Run load flow
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        assertTrue(result.isFullyConverged());
+
+        checkResult(network, "dl34p", expectedCurrent, List.of(), List.of("conv23p", "conv45p"));
+
+        // Check current flows in the connected DC line
+        assertNotEquals(0., network.getDcLine("dl34n").getDcTerminal1().getI(), 1e-3);
```
SGI-Landry: _Non-zero_ is too loose. Can you please rather check the expected current flow?

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+    void testDoubleDcLineDisconnectionRigidBipole() {
+        /// Removing the two DC lines creates two DC components. Similarly to testDcLineDisconnectionAsymmetricalMonopole(), it is expected that:
+        /// - The DC current in the DC lines is zero
+        /// - The only power that goes into VSC is to compensate their idle losses
+        Network network = AcDcNetworkFactory.createBipolarModelWithoutMetallicReturn();
+        network.getDcLine("dl34p").getDcTerminal1().disconnect();
+        network.getDcLine("dl34n").getDcTerminal1().disconnect();
+
+        // Add idle losses to converters and change their control mode to ensure DC buses voltage is imposed
+        network.getVoltageSourceConverters().forEach(vsc -> vsc.setIdleLoss(0.5).setTargetVdc(400).setControlMode(AcDcConverter.ControlMode.V_DC));
```
SGI-Landry: You can even add non-idle losses, which will help make sure the converters run at zero current.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+
+        // Run load flow
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        assertTrue(result.isFullyConverged());
+
+        // Check the current in the disconnected DC lines
+        assertEquals(0., network.getDcLine("dl34p").getDcTerminal1().getI(), 1e-3);
```
SGI-Landry: Should it be strictly zero or converge to zero? This will give you the required tolerance.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,392 @@
...
+import java.util.stream.Stream;
+
+import static java.lang.Math.abs;
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * @author Baptiste Perreyon {@literal <bapstiste.perreyon at supergrid-institute.com>}
+ */
+@ExtendWith(ServiceParameterResolver.class)
+class AcDcLoadFlowWithDisconnectionTest {
```
SGI-Landry, line 35: This can be simplified and made more robust. Can we please write down what behaviors to expect in a _if condition then expectation_ way, and create just these tests, over observable behaviors? This should rely on analytical cases when this is numerical, with comparisons to expected values.

src/main/java/com/powsybl/openloadflow/network/impl/DcNetworkValidationHelpers.java
```diff
@@ -0,0 +1,176 @@
...
+ * @author Baptiste Perreyon {@literal <baptiste.perreyon at supergrid-institute.com>}
+ */
+final class DcNetworkValidationHelpers {
+
+    private DcNetworkValidationHelpers() {
+    }
+
+    /**
+     * Check that at least one of the DC terminal of each AC-DC converter is indirectly connected (i.e. through DC
+     * lines) to a DC ground.
```
SGI-Landry:
```suggestion
     * Check that at least one of the DC terminal of each AC-DC converter is indirectly connected (i.e. through DC
     * branches) to a DC ground.
```

src/main/java/com/powsybl/openloadflow/network/impl/DcNetworkValidationHelpers.java
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * @param acDcConverters A list of AC-DC converters to check.
+     * @throws PowsyblException If at least one AC-DC converter is not indirectly connected to a DC ground
+     */
+    public static void checkAllConvertersAreIndirectlyConnectedToADcGround(List<AcDcConverter<?>> acDcConverters) {
```
SGI-Landry: Would it make sense to leave it package-private? I think these static methods are only used in `LfNetworkLoaderImpl` anyway.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -653,6 +659,9 @@ private static void createAcDcConverter(AcDcConverter<?> acDcConverter, LfNetwor
                 addVoltageSourceConverter(lfNetwork, voltageSourceConverterImpl);
             }
             // lcc converter not implemented yet
+        } else {
+            // Remove previous load flow results (we don't want the user to believe this are results from the new load flow)
+            acDcConverter.unsetSolvedValues();
```
SGI-Landry: Very good point. However, a converter whose DC terminals are both disconnected belongs to no DcBus so it is never added to the acDcConverterSet in the first place.

Same issue with DcLines disconnected at both ends.

src/main/java/com/powsybl/openloadflow/network/impl/DcNetworkValidationHelpers.java
```diff
@@ -0,0 +1,176 @@
...
+
+import java.util.*;
+import java.util.function.Predicate;
+
+/**
+ * Helper methods to validate a DcComponent configuration
+ *
+ * @author Baptiste Perreyon {@literal <baptiste.perreyon at supergrid-institute.com>}
+ */
+final class DcNetworkValidationHelpers {
```
SGI-Landry: Good call to factor out these static functions. I am wondering if it is possible to simplify this code a bit. Maybe extract the topology visitor, and replace the predicate on converters by a boolean stating whether the test is needed or not?

SGI-Landry (changes requested):
Thanks for this activity @bperr !
There are a few minor comments here and there. I think however that we should discuss tests in `AcDcLoadFlowWithDisconnectionTest`.

src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
bperr: Done after the load flow in Networks.resetState as all other equipments

docs/loadflow/loadflow.md
bperr: If the converter is disconnected, it is removed from the load flow and no equation is added to the equation system : no DC voltage control nor AC voltage/active power/reactive power control, no inclusion in the active/reactive power balance of AC buses... 
As stated in the doc :  **none of the equations below is added**

src/main/java/com/powsybl/openloadflow/network/impl/DcNetworkValidationHelpers.java
bperr: Yes we can do it

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,227 @@
...
+        Network network = AcDcNetworkFactory.createAcDcNetworkTwoParallelDcLines();
+
+        // Run load flow on the complete network
+        LoadFlowResult result = loadFlowRunner.run(network, parameters);
+        assertTrue(result.isFullyConverged());
+
+        double dcLineR = network.getDcLine("dl34").getR(); // 0.1 Ohm
+        double expectedDcCurrent = 125; // 50 MW/400kV = 125A
+        // Check DC current
+        assertEquals(-expectedDcCurrent, network.getVoltageSourceConverter("conv23").getDcTerminal1().getI(), 1e-2);
```
SGI-Landry: How is the tolerance chosen?
1e-2 absolute ~ 1e-5 relative. Not bad, but depending on convergence parameters we can probably do better.

What about
* checking the convergence parameters to decide on the relative tolerance.
* factor it in a named constant, and document the choice.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,227 @@
...
+        assertEquals(expectedDcCurrent, network.getVoltageSourceConverter("conv45").getDcTerminal1().getI(), 1e-2);
+        assertEquals(expectedDcCurrent / 2, network.getDcLine("dl34").getDcTerminal1().getI(), 1e-2);
+        assertEquals(expectedDcCurrent / 2, network.getDcLine("dl34_bis").getDcTerminal1().getI(), 1e-2);
+        // Check DC voltage, taking into account the voltage rise due to the equivalent resistance of the two DC lines
+        assertEquals(400 + expectedDcCurrent / 1000 * dcLineR / 2, network.getDcNode("dn3").getV(), 1e-3);
+        assertEquals(400, network.getDcNode("dn4").getV());
+
+        // Disconnect a DC line and run load flow. The current in the disconnected DC line should be zero, and doubled in the other DC line
```
SGI-Landry, line 96: Thanks for the comments, it really helps understanding.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,227 @@
...
+        assertEquals(expectedDcCurrent, network.getDcLine("dl34").getDcTerminal1().getI(), 1e-2);
+        assertEquals(0., network.getDcLine("dl34_bis").getDcTerminal1().getI(), 0);
+        // Check DC voltage, taking into account the voltage rise due to resistance of ONE DC line
+        assertEquals(400 + expectedDcCurrent / 1000 * dcLineR, network.getDcNode("dn3").getV(), 1e-3);
+        assertEquals(400, network.getDcNode("dn4").getV());
+    }
+
+    @Test
+    void dcLineTotalDisconnection() {
+        /// Same case study as dcLinePartialDisconnection. However, both terminals of the DC line are disconnected.
```
SGI-Landry: There seems to be a copy-paste error here. Code is inconsistent with the comment.

May I suggest to do this test at the end of the previous function? We would have
* Results with both lines connected.
* Results with one line disconnected.
* Results with both lines disconnected.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,227 @@
...
+        assertEquals(expectedDcCurrent, network.getVoltageSourceConverter("conv45").getDcTerminal1().getI(), 1e-2);
+        assertEquals(expectedDcCurrent, network.getDcLine("dl34").getDcTerminal1().getI(), 1e-2);
+        assertEquals(Double.NaN, network.getDcLine("dl34_bis").getDcTerminal1().getI());
+        // Check DC voltage, taking into account the voltage rise due to resistance of ONE DC line
+        assertEquals(400 + +expectedDcCurrent / 1000 * dcLineR, network.getDcNode("dn3").getV(), 1e-3);
+        assertEquals(400, network.getDcNode("dn4").getV());
+    }
+
+    @Test
+    void dcLineDisconnectionLeadsToZeroCurrentInConverter() {
```
SGI-Landry, line 125: I don't understand the difference with the previous test cases. Does it make sense to regroup them? I miss the general intent. What is your activity supposed to implement and what simple behaviors do you want to check?

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -0,0 +1,227 @@
...
+        assertEquals(expectedDcCurrent / 1000 * dcLineR + 400, network.getDcNode("dn3p").getV(), 1e-3);
+    }
+
+    /*
+     * AC-DC converter disconnection
+     */
+    @Test
+    void testAcDcConverterDisconnectionLeadsToZeroCurrentInNearbyIsolatedDcLines() {
+        /// When disconnecting the converter, the current in the DC line at the disconnected pole (positive) should be
+        /// null, but the current should still pass through the negative and neutral pole, due to the second converter.
```
SGI-Landry: I would state the expected behavior as something like

> Disconnecting a terminal of a VSC removes the equations from its branch but not of the others.

Is this what you want to test?

SGI-Landry (changes requested):
Thanks for largely simplifying tests! Can you please make sure that the behaviors you want to test are perfectly clear? I suggest writing them down in comments as _if condition then expectation_. Then you will be sure you are testing exactly what you want.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
bperr: The network is a bipolar connection with DMR. When removing a converter with positive and neutral layer, the positive layer is isolated and no current can pass through it. However, the neutral layer is still connected to the negative layer, so the current can pass here. This is what this test verifies.

I don't understand what you mean by 'branch' here so I am not confortable with your proposal.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
bperr: In the previous test case, the disconnection of the DC line did not prevent the current to pass in the DC network (because there was a second DC line). 
Here the disconnection of the DC line isolates the positive layer and triggers the automatic update of a converter's control mode. Additionnaly there is no DC current in this converter so its AC consumption corresponds to its idle loss

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
bperr: The difference between the two test is whether the DC line is disconnected on one side or on both side; not whether there is one or two lines disconnected.

However its true that running the load flow on the fully connected network in the two tests brings no added value. I'll merge the two tests

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
bperr: 1e^-2 is the default tolerance in OLF for Newton-Raphson.  
I imported it in the test from `NewtonRaphsonStoppingCriteria.java` in which it is defined in p.u.

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -215,13 +200,14 @@ void testAcDcConverterDisconnectionLeadsToZeroCurrentInNearbyIsolatedDcLines() {
         LoadFlowResult result2 = loadFlowRunner.run(network, parameters);
         assertTrue(result2.isFullyConverged());
 
-        // Check current is 0 in the positive pole
+        // Check current is NaN in the disconnected converter and 0 in the positive pole
+        assertEquals(Double.NaN, network.getVoltageSourceConverter("conv23p").getDcTerminal1().getI());
```
SGI-Landry, line 205: Usually you cannot compare with NaN with == because `Double.NaN == Double.NaN` is false. Now, I am not sure how assertEquals works. Can you confirm how this behaves?

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
SGI-Landry: You are making a good point. It would be very helpful to put this in your comments. Does this mean that you are testing "isolate the pole of a VSC => trigger automatic update of control mode" and maybe "=> zero DC current"? Then is it possible to check it more directly?

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
```diff
@@ -34,6 +36,8 @@ class AcDcLoadFlowWithDisconnectionTest {
 
     private final CommonTestConfig commonTestConfig;
 
+    private final double tol = NewtonRaphsonStoppingCriteria.DEFAULT_CONV_EPS_PER_EQ * PerUnit.SB;  // 10^-2
```
SGI-Landry, line 39: Great, thanks, it makes sense!

SGI-Landry (changes requested):

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
bperr: JUnit docs states that "NaNs are considered equal: assertEquals(Double.NaN, Double.NaN, *) passes "

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
bperr: I reformulated the comment hoping it is clearer now, but this was already mentioned.

In the test, I already check that the DC current in the isolated converter is 0. 
As for the automatic update of the control mode, the IIDM is left untouched, so we cannot check it directly. However, what is already tested is that the voltage difference between the two DC buses of the converter is **exactly** 400 kV (the nominal voltage, set as `targetVdc` when the control mode of the converter is updated)

src/test/java/com/powsybl/openloadflow/acdc/AcDcLoadFlowWithDisconnectionTest.java
SGI-Landry: Ok, good. Again, a short comment will avoid any reader wondering if/how it works. Thanks.

src/main/java/com/powsybl/openloadflow/network/impl/DcComponentValidator.java
```diff
@@ -0,0 +1,196 @@
...
+     */
+    private static boolean isConnectedToGround(DcTerminal startTerminal) {
+        return DcIsland.around(startTerminal).hasConnectedDcGround();
+    }
+
+    /**
+     * The set of DC buses reachable from a DC bus by following connected DC lines, together with
+     * the connected DC grounds and the AC-DC converters attached to them. Built by breadth-first search.
+     */
+    private static final class DcIsland implements DcTopologyVisitor {
```
jeandemanged, line 125: I am wondering why this is needed, all topology processing: AC islands, DC islands, Connected AC/DC islands are already computed within IIDM.

LfNetwork-s are built from IIDM connected component-s
LfSynchronousNetwork are built from IIDM synchronous component-s

Instead of recomputing islands because of information lost at LfNetwork loading time, I think you should create for example LfDcNetwork from IIDM DC component-s at LfNetwork loading time.

jeandemanged (commented):
Thank you
This is a quite large PR with more features that the PR title and description suggests, please break it down because it makes reviewing difficult. for example:
- import also DC components in LfNetwork (my other comment, is there any reason to not just use the islands computed by IIDM?)
- the "Sanity checks : detecting invalid DC configuration and automatically resolving reference-less islands"
- a fix to set unsolved values (Networks.java)
- support of disconnected elements

also please check for sonar (coverage and issues).

src/main/java/com/powsybl/openloadflow/network/impl/DcComponentValidator.java
bperr: I realise the name of this class conflicts with the DC Island concept (I only had DC component wording in mind, sorry), and the docstring may be a bit unclear.

Here the only elements used to navigate between DC buses are DC lines. Therefore, one can only reach DC buses of the same layer (positive/neutral/negative) which is different from the DC island computed in IIDM. I'll refactor this to make it clear.

_Side note:_ This process is done using the original `Network` object not the `LfNetwork`, so no information has been lost at this stage.

bperr:
Hi @jeandemanged , thanks for your review !
The scope of this PR has indeed quite grown up. If needed, I'm fine with splitting it. Else, I'll update the description as you suggested

src/main/java/com/powsybl/openloadflow/network/impl/DcComponentValidator.java
bperr: I would add that when a DC line is disconnected, a "layer" may be split in two : the DC buses are not connected via the DC line anymore so their voltage are independent (so we need an element imposing it for each of them). However they are still part of the same DC component (because they are indireclty connected by converters and lines of another layer)

bperr:
As suggested, I have split this PR:
- #1486 covers the fix on DC variables reset
- #1487 covers detection and auto-correction of invalid DC component
- #1471 (this one) implements load flow with DC elements disconnection and relies on the two others

bperr:
@jeandemanged @SylvestreSakti This PR is now ready for review

src/main/java/com/powsybl/openloadflow/ac/equations/dcnetwork/OpenDcLineEquationTerm.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public class OpenDcLineEquationTerm extends AbstractElementEquationTerm<LfDcLine, AcVariableType, AcEquationType> {
+
+    public OpenDcLineEquationTerm(LfDcLine dcLine) {
+        super(dcLine);
+    }
+
+    @Override
+    public double eval() {
+        return 0.0;
```
SylvestreSakti: Hello @bperr, I do not get the use of creating this whole EquationTerm which seems to be what `EvaluableConstants.ZERO` does. Am I missing the purpose ?

src/main/java/com/powsybl/openloadflow/ac/equations/AcEquationSystemCreator.java
```diff
@@ -819,13 +824,27 @@ protected void createDcLineEquations(LfDcLine dcLine, LfDcBus dcBus1, LfDcBus dc
...
+            i1 = openI1;
+            p2 = EvaluableConstants.ZERO;
+            i2 = EvaluableConstants.ZERO;
+        } else if (dcBus2 != null) {
+            openP2 = new OpenDcLineEquationTerm(dcLine);
+            openI2 = new OpenDcLineEquationTerm(dcLine);
+            p1 = EvaluableConstants.ZERO;
+            i1 = EvaluableConstants.ZERO;
+            p2 = openP2;
+            i2 = openI2;
```
SylvestreSakti: This seems to work, that's why I am not getting why creating the specific EquationTerm
```suggestion
        } else if (dcBus1 != null) {
            p1 = EvaluableConstants.ZERO;
            i1 = EvaluableConstants.ZERO;
            p2 = EvaluableConstants.ZERO;
            i2 = EvaluableConstants.ZERO;
        } else if (dcBus2 != null) {
            p1 = EvaluableConstants.ZERO;
            i1 = EvaluableConstants.ZERO;
            p2 = EvaluableConstants.ZERO;
            i2 = EvaluableConstants.ZERO;
```

bperr: It is a matter of typing. An `Evaluable` cannot be added to an equation; it must be an `EquationTerm` (see `createDcLineEquations()` line 988). 
That is why I chose to create an `EquationTerm`, even though it returns always zero. Other possible solutions would be 
- Creating an `EquationTerm` from an `Evaluable`
- Not adding the open equation term to the equation (since they are zero). However, that would mean the equation might end up with no terms at all, which felt like a risky path

SylvestreSakti: To my mind, there is no reason to add a zero EquationTerm to an equation. If the equations ends up with no terms, then it means the associated bus should have been disabled or does not exist in the LfNetwork (another risk if we keep such an equation would be to have a zero line in the Jacobian matrix).

SylvestreSakti: > Not adding the open equation term to the equation (since they are zero). However, that would mean the equation might end up with no terms at all, which felt like a risky path

I think it's cleaner to do so

bperr: I just tested it and it works fine indeed. So no need for an OpenDcLineEquationTerm

SylvestreSakti (approved):

PR #1487, opened 2026-09-09
Detect and auto-repair invalid DC configuration

bperr:

No

This PR add a validator class that verifies if a DC component can be simulated by Open Load Flow and fix it if needed.

Before builing the LfNetwork, we make sure at least one AC/DC converter controls DC voltage, and throw an error otherwise.

We introduce here the concept of DC subcomponent: a set of DC buses connected by DC lines only (i.e. not converters).
A DC component is valid if:
- All AC/DC converters are indirectly connected to a DC ground with at least one of their DC terminal. An error is raised otherwise.
- In each DC subcomponent, there is at least one element imposing voltage (a DC ground or a converter not in P_PCC mode). This ensures that the voltage of all DC buses in the DC component is controlled. If no element control the DC voltage in a DC subcomponent, all converters belonging to this subcomponent are set in V_DC mode during the load flow with the nominal voltage as targetVdc.

_Example of invalid DC component configuration:_ a bipolar P2P connection with 3 converters in P_PCC mode and one converter in V_DC mode (connecting the positive and the neutral pole). The voltage in the negative pole would not be imposed : the load flow cannot converge.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
+++ b/src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
@@ -630,7 +630,14 @@ public class LfNetworkLoaderImpl implements LfNetworkLoader<Network> {
         }
     }
 
-    private static void createAcDcConverter(AcDcConverter<?> acDcConverter, LfNetwork lfNetwork, LfNetworkParameters parameters) {
+    private static void createAcDcConverters(LfNetwork lfNetwork, LoadingContext loadingContext, LfNetworkParameters parameters,
+                                             List<AcDcConverter<?>> convertersToSetInVdcMode, double dcNominalV) {
+        for (AcDcConverter<?> acDcConverter : loadingContext.acDcConverterSet) {
+            createAcDcConverter(acDcConverter, lfNetwork, parameters, convertersToSetInVdcMode.contains(acDcConverter) ? Optional.of(dcNominalV) : Optional.empty());
+        }
+    }
+
+    private static void createAcDcConverter(AcDcConverter<?> acDcConverter, LfNetwork lfNetwork, LfNetworkParameters parameters, Optional<Double> vdcOverride) {
 
         if (acDcConverter.getTerminal2().isPresent()) {
             throw new PowsyblException("Open Load Flow does not support AC/DC converters with two AC terminals");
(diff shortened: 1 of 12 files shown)


src/main/java/com/powsybl/openloadflow/network/impl/LfNetworkLoaderImpl.java
```diff
@@ -1365,17 +1372,19 @@ private void createDc(LfNetwork lfNetwork, int numDcc, Network network, List<DcB
...
-                break;
-            }
-        }
-        if (!isVdcControlled) {
-            throw new PowsyblException("At least one AC/DC converter control mode must be V_DC in each DC component, but DC component " + numDcc + " does not have any");
-        }
+        // -- Sanity checks : detecting invalid DC configuration and automatically resolving reference-less islands
+        double dcNominalV = dcVoltages.iterator().next();
+        List<AcDcConverter<?>> convertersToSetInVdcMode =
+                DcComponentValidator.resolveDcComponent(dcBuses, loadingContext.acDcConverterSet, numDcc);
```
SylvestreSakti, line 1378: Is it something that we want to do systematically ? Or should it be parameterized ?
What happens in real, if an island is separated with no voltage constraint ? Is it a data definition problem or a real behaviour ?

Depending on the answer, maybe some user may want to have an exception instead of an invisible data correction

jeandemanged: This is the equivalent of e.g. no voltage control for AC LF, in this case PowSyBl flags the LfNetwork as invalid (does not throw). Probably the same should be done, with auto-reparing as an opt-in parameter.

bperr: In real, if an island is separated with no voltage control, the voltage rises until the converter limits are exceeded, damaging the equipment. Therefore, converters usally activate a droop.

In the code, I have chosen to rather go for V_DC mode because there is no guarantee the converter have a droop curve. All converters are updated with the same targetVdc so the load flow does not fail but the invalid subcomponent can easily be identified.
If the converter control mode is not changed, the DC bus voltage goes infinite to compensate a near-zero current to ensure the converter's targetP equation is respected. --> Basically unrealistic results.

Having a parameter to choose whether to throw/fix is not a problem (throwing was the behaviour before this PR). However, flagging the LfNetwork as invalid when AC voltage control is missing occurs at a later stage, with an algorithm based on the LfNetwork while the process of invalid DC subcomponent detection uses the iidm so this would require consequent changes

jeandemanged: If this reflects/models what converters would do in real life then fine for me.
There is a report, so this is a transparent processing, adding parameter is not necessary. Flagging island as invalid not necessary.

Maybe some things to consider for the future (I am not requesting it here):
- OLF's graph package could have been used to compute the DC sub-components instead of rolling your own computation
- maybe more important, using LfNetwork instead of IIDM in the checks would allow more naturally to perform these checks within security analysis. The feature is today confined to LfNetworkLoading, but will probably need in the future to be an LfNetwork feature. E.g. a meshed DC grid where the only converter in DC voltage control is lost by the contingency, and the rest of the DC island survives by adjusting control modes.

bperr: Using the LfNetwork for these checks within security analysis is a good idea. I will keep that in mind !

bperr:
Hi @SylvestreSakti @jeandemanged ,
Apart from the point we discussed two weeks ago (on which I understand we concluded that the report was enough for now), are there any changes you would like to make yothis PR ?

jeandemanged:
> Hi @SylvestreSakti @jeandemanged , Apart from the point we discussed two weeks ago (on which I understand we concluded that the report was enough for now), are there any changes you would like to make yothis PR ?

Hello @bperr ,
From my perspective it is fine and I am approving, thank you for the explanations provided!

jeandemanged (approved):

SylvestreSakti:
Hello @bperr for me also this is good. If you are ok with it we can merge it.

In the future indeed, this feature may have to be handled directly in the LfNetwork if we want to support it in the security analysis.

SylvestreSakti (approved):

